PR #108, opened 2020-12-17
add namespace revisions data

samthor:
This adds release data for _namespaces_ only.

There's a couple of cases where the versions show up earlier than the old site. This seems to largely be because _types_ were added earlier than functions for some APIs (it's not clear what the old site used as the version source—paging @karan173 ), and I suspect an API was only considered ready when any functions existed.

This is generated via changes to the "chrome-types" repo which lets us check out old versions of the code and run our TSDoc generation on that.


samthor:
I tried filtering to namespaces with properties or methods and... the changes were:

- a bunch of namespaces disappeared (since e.g. "chome.types" literally has no methods/properties, just `ChromeSetting`)—some others are manifest keys only, but they don't show up in our site anyway
- "chrome.declarativeNetRequest" went from 62 to 67, because that was the first time it had methods 🤦 , but the old site data claims it was brought in at 84.

karan173:
An API function/permission etc. might just be available to a particular channel. This availability is specified using [feature files](https://source.chromium.org/chromium/chromium/src/+/master:chrome/common/extensions/api/_features.md?q=file:md%20file:feature%20file:extensions). Hence accurate availability information can't be obtained from just parsing the API IDL/JSON schema files. The feature files need to be ingested to get a complete view of the API availability. 

For example, the declarativeNetRequest API is considered to be available since Chrome 84 since that was the first version where the declarativeNetRequest API feature was marked as available to Stable. See [link](https://chromium-review.googlesource.com/c/chromium/src/+/2194802).

I have never really dived into the [availability code](https://source.chromium.org/chromium/chromium/src/+/master:chrome/common/extensions/docs/server2/availability_finder.py?q=AvailabilityFinder) and don't really grasp how it all works (I think it iterates over git history and what not). I guess what developers are really interested is in:
- When was the API node first available on Stable. This will require parsing the corresponding feature files. 
- If an API node is not currently available on Stable, we should mention on what channel the API node is available on. The information here would still be available through feature files.
- Generally we should only list the availability of an API node separately if it is different from the availability of the namespace.

The old Docserver even identified whether the first available milestone for an API node was the current Beta, trunk etc. E.g. see https://3-72-0-dot-chrome-apps-doc.appspot.com/extensions/declarativeNetRequest#method-setExtensionActionOptions (available on current beta) and https://3-72-0-dot-chrome-apps-doc.appspot.com/extensions/declarativeNetRequest#method-getAvailableStaticRuleCount (available on current trunk). 

Also, @rdcronin in case he has more context.

PR #145, opened 2020-12-29
remove all plural domain use

samthor:
Remove references to the plural domain, as we're canonically served from developer.chrome.com.

The major impact of this PR is that often users would bounce to the plural domain and then _back_ to the singular domain when following links. We remove that bounce.


MichaelSolati (approved):

dotproto (approved):
LGTM

PR #152, opened 2020-12-31
Fix duplicate chrome.cookies API reference appearing in indexes

### LINKED ISSUE #45: Api Reference @[https://developer.chrome.com/docs/extensions/reference/] Contains 2 Entries to Cookies
In Api Reference Table located [Here](https://developer.chrome.com/docs/extensions/reference/) Contains Two Entries of Cookies & They Also Refer Same Content.

### LINKED ISSUE #115: Docs Error
**Describe the bug**
2 cookies API entries on table at https://developer.chrome.com/docs/extensions/reference/

mlcui-google:
This was caused by a copy of the chrome.cookies API reference appearing inside the chrome.commands API reference folder. Fixes #45, fixes #115.


dotproto (approved):
Tested locally and it appears to be working as expected. Thanks for the PR!

PR #165, opened 2021-01-04
Formatting

Yash-Singh1:
Formatted storage onchanged listener


Yash-Singh1:
@googlebot I signed it!

samthor:
This doesn't improve the repo. Our code has pretty inconsistent formatting right now; we're not specifically going to accept PRs that change it one way or another until we have a general policy (and at that point we'll probably just migrate everything automagically).

PR #172, opened 2021-01-05
resolves md linter issues, enable remark

samthor:
Currently, `remark` runs but it does not fail on errors.

This PR enables it and fixes remaining linter issues, except that it disables the `lint-no-literal-urls` plugin: we have lots of URLs sprinkled throughout the code.

Frustratingly this PR probably needs some manual review as there's a few pages where I had to make changes. Some notable ones:
- I removed a couple of hand-written TOCs, we don't need these and they were causing issues
- some minor manual fixes to the HTML
- code tags added on ``` blocks.


MichaelSolati (approved):

site/en/docs/extensions/api_other/index.md
```diff
@@ -12,62 +12,57 @@ API libraries into your extension.
 
 Here's a sampling of the APIs that extensions can use:
 
-TODO clean up definition list
+Standard JavaScript APIs
```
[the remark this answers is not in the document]
samthor: The definition list Markdown code is a bit hard to read but it basically works like:

```text
Unindented line

: Indented line starting with :

  More lines can also be indented
```

GitHub doesn't support it, so I can't preview it. But basically every line _before_ a line starting with ` will get bolted (as it's put inside a `<dt>`).

samthor:
Also FWIW I need to submit #171 before this otherwise remark will fail anyway.

.remarkrc.js
```diff
@@ -8,7 +8,6 @@ exports.plugins = [
   require('remark-preset-lint-markdown-style-guide'),
   ['lint-blockquote-indentation', false],
   ['lint-emphasis-marker', false],
-  // ['lint-fenced-code-flag', false],
```
robdodson, line 11: Do you want to remove this? I think it grumps when people forget to include a language with their code fence.

site/en/404/index.md
```diff
@@ -3,7 +3,7 @@ title: 404
 layout: 'layouts/error.njk'
 ---
 
-# Aw, Snap!
+## Aw, Snap!
```
robdodson, line 6: I think this is supposed to stay an h1 but was it making the linter angry?

site/en/docs/extensions/api_other/index.md
```diff
@@ -12,62 +12,57 @@ API libraries into your extension.
...
 
-```
-<style>
-  div:hover {
-    transform: rotate(360deg);
-    transition: all 1s ease-out;
-  }
-</style>
-```
+  ```html
```
robdodson, line 47: does this have extra indentation or is that github displaying it oddly?

robdodson: or is it appearing inside of a definition list?

site/en/docs/extensions/mv2/faq/index.md
```diff
@@ -12,45 +12,6 @@ If you don't find an answer to your question here, try the [Chrome Web Store FAQ
 [\[google-chrome-extension\] tag on Stack Overflow][2], the [chromium-extensions group][3], or the
 [store help][4].
 
-#### General {: #general }
```
robdodson, line 15: Only concern here is that we're removing these headings which seem to have specific id targets (possibly for deep linking?) Might want to check with @dotproto and @awfuchs to make sure this change is ok

site/en/docs/extensions/mv3/faq/index.md
```diff
@@ -10,45 +10,6 @@ If you don't find an answer to your question here, try the [Chrome Web Store FAQ
 [\[google-chrome-extension\] tag on Stack Overflow][2], the [chromium-extensions group][3], or the
 [store help][4].
 
-#### General {: #general }
```
robdodson, line 13: Same comment as above re: removing deep links

site/en/docs/extensions/mv3/manifest/version/index.md
```diff
@@ -6,7 +6,7 @@ updated: 2018-04-26
 description: Reference documentation for the version property of manifest.json.
 ---
 
-# Manifest - Version
+## Manifest - Version
```
robdodson, line 9: This seems like it could be deleted

site/en/docs/extensions/mv2/publish_app/index.md
```diff
@@ -10,7 +10,7 @@ description: TODO
 
 <div class="aside aside--caution"><b>Important:</b> Chrome will be removing support for Chrome Apps on all platforms. Chrome browser and the Chrome Web Store will continue to support extensions. <a href="https://blog.chromium.org/2020/01/moving-forward-from-chrome-apps.html"><strong>Read the announcement</strong></a> and learn more about <a href="https://developer.chrome.com/apps/migration"><strong>migrating your app</strong></a>.</div>
 
-# Publish Your App
+## Publish Your App
```
robdodson, line 13: this seems like it could maybe be deleted (it repeats the title from above)

site/en/docs/webstore/faq/index.md
```diff
@@ -10,124 +10,7 @@ find them. This FAQ answers many common questions about the Chrome Web Store. If
...
-## Extensions
-
-- [Can I sell extensions in the store?][88]
-- [Can I use the Licensing API with extensions?][89]
-
-## Themes
-
-- [Can I sell themes in the store?][90]
-- [Can I use the Licensing API with themes?][91]
+Please expand the _Table of Contents_ to quickly navigate this page.
```
robdodson, line 13: idk, i'm not sure I agree with this change. If the linter is grumping we can add an html comment to make it chill.

robdodson (commented):
I think this seems like a good change, but I'm hesitant to make such big changes to the various FAQ pages. I'm wondering if it might be safer to leave those pages as is and use an html comment to make remark ignore them?

.remarkrc.js
samthor: Yes, I'd like it to grump. The default is true, so leaving it out is good.

site/en/404/index.md
samthor: I've added a disable comment on this page.

site/en/docs/extensions/api_other/index.md
samthor: Yes, this is probably unrelated, and just reflects that the code should be inside the list.

site/en/docs/extensions/mv2/faq/index.md
samthor: Yes and no; there was actually duplication (`general` vs `general2`) for headings here because the TOC used some of them up-front, if that makes sense.

I'm less worried partially only because we _already_ have a bunch of bad #-refs. I think we can fix those at some point by looking at missing links broadly.

site/en/docs/extensions/mv3/faq/index.md
samthor: And same reply :)

site/en/docs/webstore/faq/index.md
samthor: Why? We have a TOC which literally replaces this: now we'd just be duplicating it. And on mobile it's even more obvious, because expanding the TOC means you see the TOC then the inline TOC right below it.

robdodson: On desktop I quite like having all of the links already there for me to quickly scan. But I think it's personal preference so maybe @dotproto and @awfuchs should weigh in. If they like the change then I'm all for it.

samthor: I sort of agree because I think having a visible TOC makes sense for a page where you're not reading top-to-bottom, it's very much a reference page.

I would almost suggest an `expand_toc: true` flag if you want this for FAQ pages like this—that would make it always open.

samthor:
@MichaelSolati can you take a look at this?

site/en/docs/webstore/faq/index.md
samthor: I'm making a call on this in that having TOCs _written out_ in the Markdown is redundant and I'm going to submit this PR. If we want to auto-expand a TOC later, for a FAQ page, that might be useful.

cc @jimoong who might have opinions on this.

PR #175, opened 2021-01-06
Fast cheerio

samthor:
This extends on @slightlyoff's change in the internal repo. It:

- sets`_useHtmlParser2` for cheerio, which makes Eleventy about 1/3 faster (thanks Alex)
- puts cheerio into worker threads, which on my [iMac Pro](https://en.wikichip.org/wiki/intel/xeon_w/w-2140b) uses 12 cores (of 16, `async-transforms` is set to 75% by default), and makes it 1/4 faster again
  - This works because our transforms don't depend on each other and are entirely CPU-bound.

For me, Eleventy goes from taking 16s to about 6s.

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -24,7 +24,7 @@ const {youtube} = require('./site/_shortcodes/youtube');
 const {columns} = require('./site/_shortcodes/columns');
 
 // Transforms
-const {domTransformer} = require('./site/_transforms/dom-transformer');
+const {domTransformer} = require('./site/_transforms/dom-transformer-pool');
 const {purifyCss} = require('./site/_transforms/purify-css');
 const {minifyHtml} = require('./site/_transforms/minify-html');
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -3252,9 +3252,9 @@
       }
     },
     "async-transforms": {
-      "version": "1.0.4",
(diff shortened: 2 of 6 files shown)


robdodson:
> This works because our transforms don't depend on each other and are entirely CPU-bound.

Just want to double check. Is it ok that purifyCss and minifyHtml have to run after domTransformer? IOW, those transforms do depend on dom-transformer completing first. But we only run them for production builds so I wanted to double check that you tested them.

samthor:
The `purifyCss` and `minifyHtml` run in sequence as _part_ of `domTransformer`, but now they just run on a different thread for every file.

robdodson:
Hm, maybe I'm misunderstanding but purfyCss and minifyHtml aren't part of domTransformer, they're separate transforms that are added further down in the config file. I just did a production build and everything seems ok so I guess it's a non-issue.

robdodson (approved):
let's give it a shot! it cut the dev build time in half on my macbook pro 👍

samthor:
Oh—I got confused. So those two could _also_ be put into a worker pool, I got that wrong.

The thing this parellizes is just the asides, prettyUrls, and tables code, which rely on the Cheerio parsed DOM—the part that is slow.

cleytontiagosorinho (approved):
Só os aprovados

cleytontiagosorinho (approved):
Sim

PR #180, opened 2021-01-07
Adds basic ref docs for scripting and action APIs

awfuchs:
Adds the minimum entries to pull in the generated reference documentation for the scripting and action APIs.

### DIFF
--- a/site/en/docs/extensions/reference/action/index.md
+++ b/site/en/docs/extensions/reference/action/index.md
@@ -3,5 +3,8 @@ api: action
 ---
 
 !!!.aside.warning
-This documentation is in active development. We're planning to add an expanded introduction section soon to provide better context and how-to focus. This work is tracked in issue [#193](https://github.com/GoogleChrome/developer.chrome.com/issues/193).
+This documentation is in active development.
+We're planning to expand the introduction section to provide more context and
+guidance on how to use this API.  This work is tracked in issue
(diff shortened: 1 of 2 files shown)


dotproto (changes requested):
As we discussed in our sync, please add an intro description or callout box to inform visitors that this page is not final. Also add a link to the relevant issue to flesh out this section on GH either in this message or in a comment.

awfuchs:
> As we discussed in our sync, please add an intro description or callout box to inform visitors that this page is not final. Also add a link to the relevant issue to flesh out this section on GH either in this message or in a comment.

Done. PTAL.

site/en/docs/extensions/reference/action/index.md
```diff
@@ -0,0 +1,7 @@
+---
+api: action
+---
+
+!!!.aside.warning
+This documentation is in active development. We're planning to add an expanded introduction section soon to provide better context and how-to focus. This work is tracked in issue [#193](https://github.com/GoogleChrome/developer.chrome.com/issues/193).
```
dotproto, line 6: I don't really follow this phrasing: "provide better context and how-to focus". It seems like something is missing to set up or contextualize the note about "how-to fous". 

What do you think about replacing the current sentence with this: "We're planning to expand the introduction section to provide more context and guidance on how to use this API."

awfuchs: Done. PTAL.

dotproto (approved):

PR #202, opened 2021-01-11
Adding shared array buffer article

jakearchibald:
Port of https://github.com/GoogleChrome/web.dev/pull/4322.

This site doesn't have a specific template for handling origin trials, so I just copied the content from the web.dev template.

### DIFF
--- a/site/en/blog/enabling-shared-array-buffer/index.md
+++ b/site/en/blog/enabling-shared-array-buffer/index.md
@@ -12,25 +12,40 @@ hero: image/CZmpGM8Eo1dFe0KNhEO9SGO8Ok23/VwqAOKnWHDx3nJegh94L.jpg
 alt: Iframes, images, and windows being accessed by someone 'evil'
 ---
 
-It's fair to say `SharedArrayBuffer` has had a bit of a rough landing on the web, but things are settling down. Here's what you need to know:
+It's fair to say `SharedArrayBuffer` has had a bit of a rough landing on the
+web, but things are settling down. Here's what you need to know:
 
 ## In brief
 
-- `SharedArrayBuffer` is supported in Firefox 79+, and will arrive in Android Chrome 88. However, it's only available to page that are 'cross-origin isolated' (details below).
-- `SharedArrayBuffer` is currently available in desktop Chrome, but from Chrome 91 it will be limited to cross-origin isolated pages.
+- `SharedArrayBuffer` is supported in Firefox 79+, and will arrive in Android
+  Chrome 88. However, it's only available to page that are 'cross-origin
+  isolated' (details below).
+- `SharedArrayBuffer` is currently available in desktop Chrome, but from Chrome
+  91 it will be limited to cross-origin isolated pages.
 
-You can make a page _cross-origin isolated_ by serving the page with these headers:
+You can make a page _cross-origin isolated_ by serving the page with these
+headers:
 
 ```http
 Cross-Origin-Embedder-Policy: require-corp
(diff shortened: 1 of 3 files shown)


site/en/blog/enabling-shared-array-buffer/index.md
```diff
@@ -0,0 +1,92 @@
...
+authors:
+  - jakearchibald
+description: >
+  Browsers are imposing new requirements on SharedArrayBuffer usage. Learn how to enable it cross-browser and cross-platform.
+origin_trial:
+  url: /origintrials/#/view_trial/303992974847508481
+date: 2021-01-11
+hero: image/CZmpGM8Eo1dFe0KNhEO9SGO8Ok23/VwqAOKnWHDx3nJegh94L.jpg
+alt: Iframes, images, and windows being accessed by someone 'evil'
+---
```
robdodson, line 13: We may want to add tags for this article. I put together a guide here which explains how to add tags: https://github.com/GoogleChrome/developer.chrome.com/pull/219

Because every tag generates its own page we usually try to be pretty coarse with our tags (e.g. preferring "security" as opposed to "shared-array-buffer"). This is mostly so we don't end up with a lot of pages that only have a single post.

robdodson (approved):
Thank you @jakearchibald for writing this up! Let me know if you hit any issues along the way. We're still working on putting together good onboarding guides.

Deferring to @kaycebasques for the actual editorial review.

site/en/blog/enabling-shared-array-buffer/index.md
jakearchibald: With web.dev, I usually cargo-culted the frontmatter from another article, but obviously couldn't do that here 😄 .

Is there anything else I might be missing?

robdodson: > Is there anything else I might be missing?

I just pushed a commit to this PR because I noticed the image on the blog card looked a bit off.

I gave the image on the cards an explicit height but noticed that it looked a bit blurry on my non-retina desktop monitor. I think it's because [the image is really wide](https://developer-chrome-com.imgix.net/image/CZmpGM8Eo1dFe0KNhEO9SGO8Ok23/VwqAOKnWHDx3nJegh94L.jpg?auto=format).

What do you think about cropping the image and uploading another version?

Unfortunately the homepage card is a tad larger than the blog card so the image aspect ratios are a little different but they're in the same ballpark:

414 x 208 (homepage)
328 x 184 (blog landing page)

kaycebasques:
@jakearchibald I made various minor edits: https://github.com/GoogleChrome/developer.chrome.com/pull/202/commits/cf53776035ed7f68cd37c68f3f001d1c63de7f5b

kaycebasques (approved):

site/en/blog/enabling-shared-array-buffer/index.md
jakearchibald: I took that image from another web.dev article, so I don't have the original. I can try and find a new image. What is the recommended upload size?

robdodson: at its largest the image will be displayed at 960 x 480. So I guess 1920 x 960 to support 2x displays. As an image expert I'm curious what your thoughts are? We could go even higher if you think we should be supporting higher resolutions.

jakearchibald: I think supporting 2x is enough for an image that's usually photographic

PR #219, opened 2021-01-11
Add a guide on adding a new tag.

robdodson:
Adds a guide to adding a new tag to the site.

### DIFF
--- a/site/en/docs/handbook/how-to/add-a-tag/index.md
+++ b/site/en/docs/handbook/how-to/add-a-tag/index.md
@@ -1,0 +1,39 @@
+---
+layout: 'layouts/doc-post.njk'
+title: Add a tag
+description: 'Add a new tag.'
+date: 2021-01-11
+---
+
+## Add the tag to the list of supported tags
+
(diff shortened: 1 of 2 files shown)


MichaelSolati (approved):

site/en/docs/handbook/how-to/add-a-tag/index.md
```diff
@@ -0,0 +1,39 @@
...
+
+- Add your tag to `site/_data/i18n/tags.yml`. The `en` and `es` fields correspond
+to different languages. For now, just add your tag using the `en` field.
+
+```
+news:
+  en: News
+```
+
+- Add your tag to `site/_data/supportedTags.json`. For the `title` field, use
```
petele, line 18: Why do I have to create it twice in separate files? Shouldn't it be able to auto-generate on of the files automatically?

robdodson: good question, I think it's a side effect of us implementing the supportedTags feature similar to how we did it on web.dev, and then later realizing we needed things to be translated so using the same methodology that we use elsewhere for localized strings. It could probably be improved if folks find it onerous.

PR #222, opened 2021-01-12
types update, provide const values

### LINKED ISSUE #61: [nodoc] in IDL/JSON API files is not taken into account.
When a node is marked as "nodoc" in the JSON or IDL schema file for an API, it shouldn't be visible on the site documentation. However the nodoc annotation doesn't seem to be respected currently. 

Example: See https://source.chromium.org/chromium/chromium/src/+/master:extensions/common/api/declarative_net_request.idl. Nodes marked as [nodoc] are visible in the served documentation at https://developer.chrome.com/docs/extensions/reference/declarativeNetRequest/

### LINKED ISSUE #110: Value for API Properties is not specified.
Check out the https://developer.chrome.com/docs/extensions/reference/declarativeNetRequest/#property-DYNAMIC_RULESET_ID API constant in the new documentation. Compare it with the old documentation at https://3-72-0-dot-chrome-apps-doc.appspot.com/extensions/declarativeNetRequest#property-DYNAMIC_RULESET_ID. 

The value for various API constants is not specified in the new documentation.

### LINKED ISSUE #147: chrome.browserAction.openPopup should have (un)availability clearly marked
https://developer.chrome.com/docs/extensions/reference/browserAction/#method-openPopup

the description for this makes it sound like `chrome.browserAction.openPopup` is generally available when it is def not.  it's only available to a small number of explicitly listed extensions, it's been like this for over 5 years, and it's unclear whether it'll ever change.  can a warning block be added to its description so devs don't waste time trying to figure out why it's not working?

the source:
https://chromium.googlesource.com/chromium/src/+/167131bf0e8024af163881f62a988589838d7dec/chrome/common/extensions/api/_api_features.json#160
```
// This API is whitelisted on stable and should not be enabled for a wider
// audience without resolving security issues raised in API proposal and
// review (https://codereview.chromium.org/25305002).
```

the (still) open bug:
https://crbug.com/436489

samthor:
Fixes #147 and fixes #61: updates chrome-types to respect `nodoc`.

Fixes #110: chrome-types includes constant values now, both as 'properties of references' (not really a subclass, but close) and top-level properties.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5083,9 +5083,9 @@
       }
     },
     "chrome-types": {
-      "version": "0.0.7",
-      "resolved": "https://registry.npmjs.org/chrome-types/-/chrome-types-0.0.7.tgz",
-      "integrity": "sha512-tY2zjlTkhD+9h2GXKVbuDgvwbgeHAmmRRW8JOqsDVY9aL/nUOFVXZTsCfzhqPRyESevqCoxnYL2qcFkvZle1KQ=="
+      "version": "0.0.8",
+      "resolved": "https://registry.npmjs.org/chrome-types/-/chrome-types-0.0.8.tgz",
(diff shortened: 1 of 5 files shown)


dotproto (approved):
LGTM

PR #224, opened 2021-01-12
grab @deprecated and include in output

### LINKED ISSUE #113: Deprecated API functions/events are not marked as such.
Check out https://3-72-0-dot-chrome-apps-doc.appspot.com/extensions/extension#method-sendRequest. The function is marked deprecated and the corresponding Chrome milestone is also mentioned. On the [new documentation](https://developer.chrome.com/docs/extensions/reference/extension/) however, the sendRequest function is not marked deprecated. The deprecation status can be inferred from the deprecated property in API schema files. [Example](https://source.chromium.org/chromium/chromium/src/+/master:chrome/common/extensions/api/extension.json;l=40;bpv=1;bpt=0;drc=ec6e924f84b07de2270ebc8cddfea12297dd0cd0).

samthor:
Fixes #113 (mostly—we don't do version data yet, but that will come soon).

### DIFF
--- a/site/_scss/blocks/_code-sections.scss
+++ b/site/_scss/blocks/_code-sections.scss
@@ -103,6 +103,10 @@
   color: var(--color-code-number);
 }
 
+.code-sections__deprecated {
+  color: var(--color-red-medium);
+}
+
 .code-sections__callback {
   background: get-color('grey-50');
   border: 1px solid var(--color-hairline);
--- a/site/_includes/macros/render-type.njk
+++ b/site/_includes/macros/render-type.njk
@@ -17,7 +17,7 @@
 {% endif %}
 
 <div class="type stack code-sections__description">
-  {{ spec.comment | safe }}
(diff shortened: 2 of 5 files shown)


site/_scss/blocks/_code-sections.scss
```diff
@@ -99,6 +99,10 @@
   color: var(--color-pink-medium);
 }
 
+.code-sections__deprecated {
+  color: var(--color-red-medium);
```
dotproto, line 103: Out of scope for this PR, but should we have safe fallbacks in the event that CSS custom properties aren't supported?

dotproto (approved):
LGTM

site/_scss/blocks/_code-sections.scss
samthor: We use a _lot_ of CSS Custom Properties. This was Rob's design decision.

Issue #149, opened 2020-12-31
Various extension/app APIs are missing from the API reference


mlcui-google:
A diff of the old site via archive.org and the new one shows some APIs missing from the new site:

- [x] idle
- [ ] networking.config
- [x] tabGroups
- [x] browser (app only)
- [x] mediaGalleries (app only) - see #148
- [x] serial (app only)
- [x] socket (app only)
- [x] syncFileSystem (app only)

dotproto:
Whoops, misread this issue & thought it was fixed by PR #151. Reopening.

samthor:
This also include tabGroups.

robdodson:
@samthor is this still an issue now that #225 has been merged?

samthor:
This will be fixed shortly with an update to `chrome-types`. However, `networking.config` was just a blank page so we won't bring it across.

PR #225, opened 2021-01-12
re-add missing api pages

samthor:
Part of #149. Adds three missing pages, although `tabGroups` was never documented on the old site, it's part of the regular extensions API.


dotproto (approved):
LGTM

PR #227, opened 2021-01-12
Adds a Compare widget

### LINKED ISSUE #213: Need compare component from web.dev
https://web.dev/handbook/web-dev-components/#compare

petele:
Fixes #213

- Adds a `Compare` widget, similar to the one used on web.dev
- Adds a components page to /docs/handbook/components, similar to https://web.dev/handbook/web-dev-components/


site/_scss/blocks/_compare.scss
```diff
@@ -0,0 +1,52 @@
...
+.compare {
+  line-height: 1.5em;
+  margin-top: 32px;
+  padding: 0 0 0 16px;
+  position: relative;
+}
+
+.compare--label {
+  font-size: 1.125em;
+  // font-weight: $FONT_WEIGHT_MEDIUM;
```
petele, line 18: I copied the settings over from web.dev, but I didn't see a system level property for font weight. Seems OK to leave as not set. Thoughts?

site/en/blog/compare-test/index.md
```diff
@@ -0,0 +1,81 @@
+---
+title: Compare Widget
```
petele, line 2: Where should this go? It looks like we don't have a widget document page yet.

petele:
One odd thing that happens, is if you put a fenced code block in, you must put a caption in, otherwise the widget never closes. We could probably fix that easily enough switching to a `{%%}` style widget, but no more markdown in elements. Open to suggestions here!

robdodson:
> One odd thing that happens, is if you put a fenced code block in, you must put a caption in, otherwise the widget never closes. We could probably fix that easily enough switching to a {%%} style widget, but no more markdown in elements. Open to suggestions here!

Yeah I think there's a bug in the `!!!` code. But I think I may remove it and go back to using shortcodes for Aside because it's annoying to ask web.dev folks to learn a new system. And I figured out how to do markdown reference links with shortcodes so there shouldn't be any downside.

robdodson:
@petele I pushed a bunch of changes and redid the component so it's a `{% Compare %}` shortcode, similar to web.dev

I think I'm going to need to do a bigger refactor across the site to make all of our shortcodes consistent. I can tackle `{% Aside %}` next.

petele:
> I think I'm going to need to do a bigger refactor across the site to make all of our shortcodes consistent and use uppercase like web.dev, otherwise it's just going to be too frustrating for authors to go between the two sites.

That would be fantastic, and my preference. The new syntax feels a little awkward, and I think the shortcodes are easier to use, so I'm very happy with that!

And thanks for re-writing it!

site/_data/docs/handbook/toc.yml
```diff
@@ -1,3 +1,4 @@
+- url: /docs/handbook/components
```
petele, line 1: 1. What's the difference between `/docs/handbook/...` and `/style-guide/`? 
2. My instinct is that `handbook` should be at root level, since its content applies to the whole site, not just stuff under `/docs/`

robdodson (approved):

PR #253, opened 2021-01-15
fix(rss): polish up rss feeds and trim post count of feeds

MichaelSolati:


site/_filters/trim-html.js
```diff
@@ -0,0 +1,29 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const clipper = require('text-clipper').default;
+
+/**
+ * Trims HTML string to 280 characters.
```
robdodson, line 20: Would this cause the article to abruptly end in the RSS reader?

site/_collections/feeds.js
```diff
@@ -23,21 +27,34 @@
 module.exports = collection => {
   /** @type FeedsCollection */
   const feeds = {};
-  const tags = ['devtools'];
+  const tags = Object.keys(supportedTags);
```
robdodson, line 30: I'm not sure if we need an RSS feed for every supported tag... Practically speaking I doubt most folks would subscribe to them.

I think I prefer the idea of having the following feeds:

- blog (this would be the default feed for the site)
- all (this would be blog + all other docs)
- devtools (this would be specific to devtools)

In the future we may add others that are exclusive to tags. So perhaps it would be worth having a yml that defines the tags that we turn into RSS feeds.

robdodson (changes requested):
Looking at daverupert.com and css-tricks.com, it looks like both of their feeds only include the 11 most recent posts.

https://daverupert.com/atom.xml
https://css-tricks.com/feed/

I'm not really an RSS expert, but I think it would be a better experience to only include the 11 most recent items (without trimming them).

site/_filters/trim-html.js
MichaelSolati: It seems to properly close out the html tags. here's an example:

```xml
<entry>
  <title>Style guide</title> 
  <link>https://developer.chrome.com/en/style-guide/</link> 
  <updated>2021-01-15T10:29:47-08:00</updated> 
  <id>https://developer.chrome.com/en/style-guide/</id> 
  <content type="html"><!DOCTYPE html> <html lang="en"> <head> <meta charset="UTF-8"> <meta name="viewport" content="width=device-width, initial-scale=1.0"> <title>Style guide - Chrome Developers</title> <link rel="canonical" href="https://developer.chrome.com/style-guide/"> <meta property="og:site_name" content="Chrome Developers"> <meta property="og:title" content="Style guide - Chrome Developers"> <meta property="og:type" content="website"> <meta property="og:url" content="https://developer.chrome.com/style-guide/"> <meta name="description" content="Build the next generation of web experiences."> <meta name="twitter:description" content="Build the next generation of web experiences."> <meta property="og:description" content="Build the next generation of web experiences."> <link rel="apple-touch-icon" sizes="180x180" href="https://developer.chrome.com/images/meta/apple-touch-icon.png"> <link rel="icon" type="image/png" sizes="32x32" href="https://developer.chrome.com/images/meta/favicon-32x32.png"> <link rel="icon" type="image/png" sizes="16x16" href="https://developer.chrome.com/images/meta/favicon-16x16.png"> <meta name="theme-color" content="#ffffff"> <link rel="preload" as="font" crossorigin="" href="https://developer.chrome.com/fonts/google-sans-v2003/regular/latin.woff2"> <link rel="preload" as="font" crossorigin="" href="https://developer.chrome.com/fonts/google-sans-v2003/medium/latin.woff2"> <link rel="stylesheet" href="https://developer.chrome.com/css/main.css"> <script> let ctaUrl = 'https://policies.google.com/technologies/cookies'; let savedCtaUrl = localStorage.getItem('user-cookies'); if …</script></head></html></content> 
  </entry>
```

robdodson: Sorry, I meant that it seems like it would cut the article off. Whereas css-tricks and dave seem to include the entire content of their article. Otherwise someone reading in an rss reader would only get part of the article and then it would stop.

site/_collections/feeds.js
```diff
@@ -14,30 +14,75 @@
...
+    }
 
-  for (const tag of tags) {
-    const items = collection.getFilteredByTag(tag).reverse();
-    if (items.length) {
-      feeds[tag] = {
-        permalink: `/feeds/${tag}.xml`,
-        items,
-      };
+    // Check if post belongs to suuported tags
```
robdodson, line 43:
```suggestion
    // Check if post belongs to supported tags
```

site/feed.njk
```diff
@@ -9,23 +9,21 @@ pagination:
...
-  <author>
-    <name>{{ site.title }}</name>
-  </author>
+  <id>{{ absoluteFeedUrl }}</id>
   {% for item in feed.items %}
   {% set absolutePostUrl %}{{ site.url }}{{ item.url | url }}{% endset %}
   <entry>
     <title>{{ item.data.title }}</title>
-    <link href="{{ absolutePostUrl }}" />
+    <link>{{ absolutePostUrl }}</link>
```
robdodson, line 23: taking a look at an rss validator, it seemed to indicate that link was supposed to have an href

site/feed.njk
```diff
@@ -9,23 +9,21 @@ pagination:
...
 <?xml version="1.0" encoding="utf-8"?>
 <feed xmlns="http://www.w3.org/2005/Atom">
+  {% set absoluteFeedUrl %}{{ site.url }}{% if feed.url %}{{ feed.url | url }}{% endif  %}{% endset %}
   <title>{{ feed.title or site.title }}</title>
-  <subtitle>{{ feed.description or site.summary }}</subtitle>
-  <link href="{{ site.url }}{{ feed.permalink }}" rel="self" />
-  <link href="{{ site.url }}/" />
+  <subtitle>{{ feed.description or site.description }}</subtitle>
+  <link href="{{ absoluteFeedUrl }}" rel="self"/>
+  <link>{{ site.url }}{{ feed.permalink | url }}</link>
```
robdodson, line 16: taking a look at an rss validator, it seemed to indicate that link was supposed to have an href

robdodson: this is the one i was using https://validator.w3.org/feed/#validate_by_input

robdodson (changes requested):
I think we want to add

```
<link rel="alternate" type="application/rss+xml" title="developer.chrome.com" href="https://developer.chrome.com/feeds/blog.xml">
```

to `base.njk` as well.

site/_collections/feeds.js
```diff
@@ -14,30 +14,75 @@
...
+    .getAllSorted()
+    .reverse()
+    .filter(i => i.data.locale === defaultLocale);
+  const tags = ['devtools'];
   /** @type FeedsCollection */
   const feeds = {};
-  const tags = ['devtools'];
+  const blog = [];
+
+  for (const post of posts) {
```
robdodson, line 37: Would it also work to do something like:

```
for (const tag of tags) {
  const posts = collection.getFilteredByTag(tag).slice(MAX_POSTS);
  feeds[tag] = {
    items: posts,
    permalink: ...
    ...
  };
}
```

I realize it's technically not as efficient, but it does seem like it might be easier to read?

robdodson (approved):
LGTM with one comment. I think the code might be more legible with my suggestion but I defer to your judgement.

site/_collections/feeds.js
MichaelSolati: @robdodson Something like this would technically work, though it'll have us looping through all the posts for each tag. Would you like me to possibly break it up and use some well named functions to try to make the code more readable?

robdodson: nah it's not a big deal

PR #257, opened 2021-01-15
Migration of New in Chrome posts

petele:
- Migrates all New in Chrome posts from dgc to dcc


robdodson:
Rebased to fix the netlify preview. I also noticed that the video shortcode requires a named src= argument so I added that to the posts, otherwise the build was breaking.

I have it as a personal TODO to work on sorting out the differences between the shortcodes on web.dev and d.c.c.

robdodson (approved):

robdodson:
Oh annoyingly we have a bug that's causing Netlify to only show the first six blog posts. I've filed it here: https://github.com/GoogleChrome/developer.chrome.com/issues/284

Until that's fixed, the best way to preview this PR is to run it locally. I've done so and I think it all looks good.

PR #272, opened 2021-01-18
Wrong year in timer throttling article

jakearchibald:
Yes I am idiot.

### DIFF
--- a/site/en/blog/timer-throttling-in-chrome-88/index.md
+++ b/site/en/blog/timer-throttling-in-chrome-88/index.md
@@ -13,7 +13,7 @@ hero: image/CZmpGM8Eo1dFe0KNhEO9SGO8Ok23/5diIn5KBi44iwy3o6GRi.jpg
 alt: A collection of broken clocks
 ---
 
-Chrome 88 (January 2020) will heavily throttle chained JavaScript timers for
+Chrome 88 (January 2021) will heavily throttle chained JavaScript timers for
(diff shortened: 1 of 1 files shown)


robdodson (approved):

PELock:
You are an idiot bringing breaking changes to the standard APIs. You can throw away those APIs (setTimer & setInteval) to the bin anyway because people will avoid it and use some dumb hacky solutions.

This is a crazy idea to make the standard functions behave differently than it supposes to work.

I know you feel entitled to tell everyone how you feel about it, but I don't care about your feelings. Imagine this kind of throttling in Windows or Linux standard APIs like for example SetTimer() it would have tragic consequences. Linus Torvalds would shred you to pieces if you ever tried to introduce something like this in Linux.

Nobody would trust it to build reliable software, yet you single-handled decided to break the API and you are too arrogant to admit it's bad for developers.

With no regards.

PS. To hell with your battery in your MacBook and your soy latte.

jakearchibald:
For the benefit of anyone else reading this: That guy slid into my Twitter DMs, threw insults at me, then demanded by help. I politely declined due to his behaviour, but offered him free support if he'd change his tact. Instead he had a multi-day tantrum across multiple platforms, part of which is above.

robdodson:
I've locked and removed some of the responses to this thread as they seemed disrespectful. Please be mindful of our [code of conduct](https://opensource.google/docs/releasing/template/CODE_OF_CONDUCT/) when communicating with one another.

PR #276, opened 2021-01-19
Remove "signature" and "spellcheck" manifest keys.

karan173:
Removed in the chromium repo in https://chromium-review.googlesource.com/c/chromium/src/+/2621972 and https://chromium-review.googlesource.com/c/chromium/src/+/2622572.

### DIFF
--- a/site/en/docs/apps/manifest/index.md
+++ b/site/en/docs/apps/manifest/index.md
@@ -81,7 +81,6 @@ discusses each field.
   "requirements": {...},
   "sandbox": [...],
   "short_name": "Short Name",
-  "signature": ...,
   "sockets": {
     "tcp": {
       "connect": "*"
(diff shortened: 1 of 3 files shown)


dotproto:
@karan173, should "signature" also be removed from the apps documentation? ([page](https://developer.chrome.com/docs/apps/manifest/), [source](https://github.com/GoogleChrome/developer.chrome.com/blob/main/site/en/docs/apps/manifest/index.md))

karan173:
> @karan173, should "signature" also be removed from the apps documentation? ([page](https://developer.chrome.com/docs/apps/manifest/), [source](https://github.com/GoogleChrome/developer.chrome.com/blob/main/site/en/docs/apps/manifest/index.md))

Good catch, done!

samthor (approved):

PR #286, opened 2021-01-21
improved redirect at runtime code

samthor:
(This does not yet delete the old redirect generator OR have tests, both things which are possible in this PR.)

- This reads the "dist/" directory at server start time (we awkwardly load the handler when it finishes `async` loading, because the library we use is async, even though it doesn't need to be)
- This finds the deepest path which the pathname of all files are unique, _except_ for subpaths we are trying to avoid
  - These are then hoisted to the top _only_ if nothing else collides with them
- Matching starts at the deepest path of the string and walks up until we find a URL match

This is a bit different but I'm actually happier because we can write proper tests for this.


robdodson:
If @MichaelSolati or @devnook has time to review this I would much appreciate that.

server/unique-redirect.js
```diff
@@ -14,46 +14,295 @@
...
+ * @return {Generator<string, void, void>}
  */
-function buildUniqueRedirectHandler(staticPaths, code = 301) {
-  /** @type {{[right: string]: url}} */
-  let raw;
+const iterateDirs = function* (p) {
+  if (!p.startsWith('/')) {
+    throw new Error(`expected absolute URL, was: ${p}`);
+  }
+  for (;;) {
```
MichaelSolati, line 60: I think a `while (true) {` would be more readable than this.

MichaelSolati (approved):

samthor:
Some minor spacing has changed and is effecting Percy.

PR #297, opened 2021-01-22
Updated .../scripting/index.md to lowercase labels

samdutton:
As discussed with Rob.

BTW: it might be nice to standardise URL format to use lowercase only with hyphens between words, as on web.dev. 

At the moment we have  `/docs/extensions/mv3/content_scripts`, `/docs/extensions/reference/webNavigation`, etc.


robdodson:
> At the moment we have /docs/extensions/mv3/content_scripts, /docs/extensions/reference/webNavigation, etc.

yeah it's sort of an unfortunate leftover from the old site. We've had a heck of a time sorting through all the redirects and 404s so I think the extensions docs may jsut have quirky URLs. But future docs should use dashes.

PR #302, opened 2021-01-22
feat(img): add default sizes attribute, fixes #75

### LINKED ISSUE #75: img shortcode implicit sizes
@robdodson commented on [Sat Dec 05 2020](https://github.com/GoogleChrome/dcc-private/issues/431)

The img shortcode should set an implicit sizes="" assuming the image is inside of a post or doc. That will be the overwhelmingly common use case.

MichaelSolati:
Fixes #75

If no sizes defined we pull the sizes from the generated srcsset.


samthor (approved):

robdodson (changes requested):
Per the discussion in chat I think we want to grab a default sizes from the page if it's defined. This is possible using `this.page.someYamlField` but we have to change the shortcode from using arrow functions to using regular functions.

robdodson (approved):

PR #308, opened 2021-01-24
checking whether there's a hero image or not

### LINKED ISSUE #303: Blog posts without a hero image look odd
I recently realized that I slightly messed up our implementation of the blog landing page.

Here's the figma design. Notice posts with a hero image don't have any preview copy, whereas hosts without a hero image do. This makes it so they're roughly similar in size.
image

In our implementation I forgot to do this, and instead every post has preview copy:
image

Here's what it would look like if it matched the design. I removed a bit of copy from Pete's new in Chrome post to show a sort of worst case scenario:
image

I think it looks "OK" although I do miss having the preview copy on the posts with images.

iifawzi:
Fixes #303 

- checking whether there's a hero image or not. 

Name used to sign the CLA: Fawzi E. Abdulfattah

### DIFF
--- a/site/_includes/macros/cards/blog-card.njk
+++ b/site/_includes/macros/cards/blog-card.njk
@@ -31,7 +31,9 @@
   </h2>
 
   <p class="type--caption gap-top-200">
-    {{ pd.description }}
+    {% if not pd.hero %}
+      {{ pd.description }}
+    {% endif %}
   </p>
 
   {% if pd.tags %}
--- a/site/_includes/macros/cards/blog-card.njk
+++ b/site/_includes/macros/cards/blog-card.njk
@@ -30,11 +30,11 @@
     </a>
   </h2>
 
+{% if not pd.hero %}
   <p class="type--caption gap-top-200">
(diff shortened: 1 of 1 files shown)


site/_includes/macros/cards/blog-card.njk
```diff
@@ -31,7 +31,9 @@
   </h2>
 
   <p class="type--caption gap-top-200">
```
robdodson, line 33: Would you mind updating the PR to look like this:

```njk
{% if not pd.hero %}
  <p class="type--caption gap-top-200">
    {{ pd.description }}
  </p>
{% endif %}
```

Otherwise it will create an empty paragraph with extra margin on the top

robdodson (changes requested):

site/_includes/macros/cards/blog-card.njk
iifawzi: my bad, will commit the changes.

robdodson (approved):

robdodson:
great thank you!

PR #314, opened 2021-01-25
Only trim blog if running Percy

### LINKED ISSUE #284: Netlify preview only shows the first 6 blog posts.
We need to change this line https://github.com/GoogleChrome/developer.chrome.com/blob/main/.eleventy.js#L83
If we're doing a Percy build then we need to run against a predefined list of posts.
If we're doing a Netlify build then we should build the entire blog.

Related to #78

robdodson:
Fixes #284

- Only trim the blog if we're running as part of a Percy build. This checks one of the env vars that Percy sets automatically.


robdodson:
i'm going to go ahead and merge this cuuz folks keep running into it

Issue #270, opened 2021-01-18
Image upload behaviour unclear


jakearchibald:
It's unclear to me how I should be using the image uploader after reading https://developer.chrome.com/docs/handbook/how-to/add-media/.

Does the service re-encode the image? If so, does this mean I should upload a lossless copy, or extremely high quality version to avoid generational loss? If I upload a photo as a PNG (to avoid loss), will the service convert it to jpeg/webp for me, or assume that I want the image to remain lossless?

Or should I be uploading a well compressed version of the image, as it may be used as-is on the site?

robdodson:
I can update the handbook.

But some quick answers here:

> Does the service re-encode the image? If so, does this mean I should upload a lossless copy, or extremely high quality version to avoid generational loss?

yes, upload a high quality version

> If I upload a photo as a PNG (to avoid loss), will the service convert it to jpeg/webp for me, or assume that I want the image to remain lossless?

I believe it will convert to webp automatically. But if the browser doesn't support webp and you uploading and request a png, then I think you'll get a png.

jakearchibald:
Feels like there should be a way to control that, so a png can be uploaded, but it's lossy compressed, especially if it doesn't have an alpha channel.

jakearchibald:
Eg output format:
🔘 Auto
⚪ Lossy
⚪ Lossless

robdodson:
I need to read up on it a tad more but here's what we're setting https://docs.imgix.com/apis/rendering/auto/auto#format

https://docs.imgix.com/tutorials/improved-compression-auto-content-negotiation

It looks like we could maybe expose the fm property (https://docs.imgix.com/tutorials/improved-compression-auto-content-negotiation#graceful-fallback-with-fmfm) to authors to let them set a fallback.

samthor:
I think @MichaelSolati has updated this handbook page enough now.

PR #329, opened 2021-01-27
Improve image upload instructions

robdodson:
Helps with #270 but doesn't entirely fix it yet. I'd need to research the auto image conversion stuff a bit more. But this at least adds the instruction to tell folks to upload a high quality image. It also documents the `params` option on `img`. It also removes the width/height adjustment section because previously you had to add those manually but now the uploader does it automatically in the returned shortcode.

### DIFF
--- a/site/en/docs/handbook/how-to/add-media/index.md
+++ b/site/en/docs/handbook/how-to/add-media/index.md
@@ -13,23 +13,27 @@ Googlers access, so signing in with a personal account will fail.
 
 ## Choose a file
 
-- Click **Choose file** and select either an image or a video to upload.
+Upload a high quality image (jpg or png if you need alpha transparency). Our
+image CDN will handle converting the image to webp if the browser supports it
+and it will resize the image so you don't have to.
+
+- Drag one or more files to the **Drop files here!** area
 - Click **Upload**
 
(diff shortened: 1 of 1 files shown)


site/en/docs/handbook/how-to/add-media/index.md
```diff
@@ -40,26 +44,45 @@ These are custom components for `developer.chrome.com` that ensure our media is
...
+{% column %}
+{% img src="image/foR0vJZKULb5AGJExlazy1xYDgI2/iuwBXAyKJMz4b7oRyIdI.jpg", alt="ALT_TEXT_HERE", width="380", height="240" %}
+Original
+{% endcolumn %}
+{% column %}
+{% img src="image/foR0vJZKULb5AGJExlazy1xYDgI2/iuwBXAyKJMz4b7oRyIdI.jpg", alt="ALT_TEXT_HERE", width="380", height="240", params={flip: 'h'} %}
+Flipped
+{% endcolumn %}
+{% endcolumns %}
+
```
samthor, line 85: I'd add a section here basically saying...

```md
Please call out in a review if you're calling a specific Imgix API, so we can be aware of custom use-cases and potentially support them through our own shortcode directly.
```

samthor (approved):

PR #333, opened 2021-01-27
fix a variety of bad links

samthor:
Fixes a variety of bad links.

Also removes https://developer.chrome.com (or http://, or developers.chrome.com) where found, as it's implied.


samthor:
This is impractical to review.

PR #334, opened 2021-01-27
fix broken tests

### LINKED ISSUE #66: A number of tests failing
@robdodson commented on [Thu Dec 10 2020](https://github.com/GoogleChrome/dcc-private/issues/509)

I think this may have to do with us updating components and not updating their tests. But if you run `npm test` you'll get like 15 failures. We need to fix these.

samthor:
Fixes #66.

I actually don't know the best way to make sure GitHub runs these tests for us on PR so we don't break again. Happy for input. cc @robdodson

### DIFF
--- a/site/_shortcodes/video.js
+++ b/site/_shortcodes/video.js
@@ -10,8 +10,8 @@ const {bucket, gcs} = require('../_data/site.json');
  * @return {string}
  */
 const generateSource = src => {
-  let type = src.split('.').pop();
-  type = mime('.' + type);
+  const extname = path.extname(src);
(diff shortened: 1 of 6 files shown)


robdodson:
> I actually don't know the best way to make sure GitHub runs these tests for us on PR so we don't break again. Happy for input. cc @robdodson

We could have a GitHub action that runs `npm test`? would that work ok?

samthor:
I added this already. Thanks for the update :)

robdodson:
yeah just saw that. i'm (slowly) working my way through my notifications backlog

PR #342, opened 2021-01-29
Add VersionHistory docs

kaycebasques:


samthor:
This shows up [on the docs page](https://deploy-preview-342--developer-chrome-com.netlify.app/docs/). We should probably make it so we can hide it before it's ready.

kaycebasques:
@samthor need a Percy review

PR #343, opened 2021-01-29
fix pagination styles

### LINKED ISSUE #289: Pagination doesn't seem to be styled
image

it's supposed to look like this:
image

samthor:
Fixes #289

See /blog/ for a visual demo of the fix.

cc @MichaelSolati

### DIFF
--- a/site/_data/i18n/common.yml
+++ b/site/_data/i18n/common.yml
@@ -101,8 +101,8 @@ skip_to_content:
   en: 'Skip to content'
   es: 'Saltar al contenido'
 
-previous:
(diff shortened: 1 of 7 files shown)


MichaelSolati (approved):

robdodson:
@samthor @MichaelSolati this seems like it should maybe use our button styles? https://developer.chrome.com/docs/handbook/components/#buttons

It's not a huge deal, but wanted to flag it from a sort of site consistency perspective. We're using these buttons on the homepage and they have a slightly different hover behavior from the ones in this PR.

PR #349, opened 2021-01-30
Improve purge css

robdodson:
I was having [a conversation on twitter about this](https://twitter.com/tailwindcss/status/1355570032479244292) and got really excited to give it a shot. It turns out you can point purgeCss at your JS files as well and it'll include classes that it sees. This is *much* nicer than having to manually tell it to include all of the classes used by our custom elements. We can ditch all of the `purgecss start ignore` comments as well as the weird `whitelist` property in the config file.

It would be good for folks to double check any files that this touches. I did a quick pass and it looked good but Sam may want to do a production build and confirm that the code sections look correct.


samthor:
Is this just basically doing a grep for CSS-like things over our JS?

I think it's worth calling out the behavior in a comment in the builder.

PR #361, opened 2021-02-02
Fix broken masonry layout on Windows.

### LINKED ISSUE #356: Windows users seeing a broken masonry layout on the home page.
**Describe the bug**

It sounds like this may be specific to Windows.

Folks are reporting seeing a broken masonry layout when they scroll down on the homepage.

image

**To Reproduce**
1. Go to https://developer.chrome.com/
1. Scroll down to the "Chrome dev summit" post
1. You should see a visual bug with the error reporting screen

 - OS: Possibly Windows only?
 - Browser: Chrome and Firefox
 - Version Chrome 87 and 88

**Additional context**

robdodson:
Fixes #356 (sorta).

This doesn't fix the underlying issue that our card measurement seems to be a bit off on Windows. You can "fix" the issue by resizing the page so it's smaller than 960px and then the card will snap into place and remain in place. So likely something is off with how the card is being measured during its initial layout.

- Remove the CDS card as CDS has been over for a while now
- Remove an extra `1r` that should not have been in the CSS and was causing column widths to be slightly off on desktop.

### DIFF
--- a/site/_scss/blocks/_masonry-grid.scss
+++ b/site/_scss/blocks/_masonry-grid.scss
@@ -16,7 +16,6 @@
     display: grid;
     gap: get-size(500);
     grid-template-columns: 1fr 1fr;
-    grid-template-columns: 1fr;
     justify-items: center;
   }
 
--- a/site/_includes/layouts/home.njk
+++ b/site/_includes/layouts/home.njk
@@ -10,14 +10,13 @@
 
     {{ heroCard() }}
     <div class="masonry-grid gap-top-500">
-      {% include 'partials/featured-cds-card.njk' %}
       {% set blog = 'blog-' + locale %}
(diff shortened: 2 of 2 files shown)


site/_scss/blocks/_masonry-grid.scss
```diff
@@ -16,7 +16,6 @@
     display: grid;
     gap: get-size(500);
     grid-template-columns: 1fr 1fr;
-    grid-template-columns: 1fr;
```
samthor, line 19: This was the issue on Windows? Wild.

samthor (approved):

site/_scss/blocks/_masonry-grid.scss
robdodson: No sorry I misspoke. I'm writing a longer comment

PR #365, opened 2021-02-03
feat: Copyable heading links

tomayac:

- Make heading links copyable. To try this: go to https://deploy-preview-365--developer-chrome-com.netlify.app/docs/extensions/mv3/getstarted/ and click any of the subheadings to get a deeplink to this subheading on your clipboard.

### DIFF
--- a/site/_js/main.js
+++ b/site/_js/main.js
@@ -25,4 +25,4 @@ import './web-components/top-nav';
 import './third-party/announcement-banner/announcement-banner';
 
 // Copyable heading links
-import './behavior/copyable-heading-links';
+import './behaviors/copyable-heading-links';
--- a/site/_js/main.js
+++ b/site/_js/main.js
@@ -23,3 +23,6 @@ import './web-components/search-box';
 import './web-components/side-nav';
 import './web-components/top-nav';
 import './third-party/announcement-banner/announcement-banner';
+
+// Copyable heading links
+import './behavior/copyable-heading-links';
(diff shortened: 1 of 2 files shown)


site/_js/main.js
```diff
@@ -23,3 +23,6 @@ import './web-components/search-box';
 import './web-components/side-nav';
 import './web-components/top-nav';
 import './third-party/announcement-banner/announcement-banner';
+
+// Copyable heading links
+import './behavior/copyable-heading-links';
```
tomayac, line 28:
```suggestion
import './behaviors/copyable-heading-links';
```

samthor:
It's pretty easy for this to happen with this code:

image

tomayac:
> It's pretty easy for this to happen with this code:

@samthor Good catch. I have added logic to prevent this now (https://github.com/GoogleChrome/developer.chrome.com/pull/365/commits/3f230c4ea5a816669ead8bee562a909cc903e5f6).

robdodson:
I left some thoughts on the web.dev PR for this: https://github.com/GoogleChrome/web.dev/pull/4570

Let's make sure we have the same implementation for both sites.

samthor:
Marking this as draft as it needs to be rebased and updated for the new #-links which are now to the left of the heading.

I think I'm happy with this behavior but I would like to see the "copied" message either be (a) a popup below the hash, or (b) a global toast.

tomayac:
I'll close this and start over once https://github.com/GoogleChrome/web.dev/pull/6198 has landed and adapt the mechanics accordingly.

PR #368, opened 2021-02-04
[DNR] Fix part of the documentation that was not updated for global rules

Celsius273:
This replaces a list item that erroneously referred to MAX_NUMBER_OF_RULES with one that references global rules.

Text from here: crrev.com/c/2563238 declarativeNetRequest.html, line 211


dotproto:
@Celsius273, can you take a look at this?

dotproto (approved):
LGTM

Issue #264, opened 2021-01-16
Shortcodes should be uppercased like on web.dev


robdodson:
The shortcodes on d.c.c. are lowercased (e.g. `{% columns %}`) but on web.dev they're uppercased `{% Columns %}`.
Originally I did this because I had aspirations of bringing the lowercased shortcodes over to web.dev, but upon reflection that seems super annoying and it would probably be much easier to make d.c.c. match web.dev's style of uppercasing.

So we need to go through all of our shortcodes and update them to be uppercased.

PR #345, opened 2021-01-29
Improving Progressive Web App offline support detection

petele:
Changes proposed in this pull request:
- PWA offline detection changes post
- Adds tag for `progressive-web-apps`

**TODO Before Merging:**

* [ ] Update shortlink https://goo.gle/improved-pwa-offline-detection 

cc: @b1tr0t

PR #377, opened 2021-02-05
Uppercase Img and Video shortcodes

robdodson:
Part of #264

- Uppercases Img and Video shortcodes


MichaelSolati (approved):

petele:
I just pushed #345, and #386 which uses images and may not included in your update from lowercase to upper case.

robdodson:
> I just pushed #345, and #386 which uses images and may not included in your update from lowercase to upper case of image tags in this PR, can you double check please?

@petele yeah we'll clean up after we merge. The site won't build if we're using the old shortcode so it'll be quick to debug.

PR #380, opened 2021-02-05
Aside shortcode

### LINKED ISSUE #264: Shortcodes should be uppercased like on web.dev

robdodson:
Fixes #264

- Adds an `Aside` shortcode that's consistent with web.dev
- Updates all of the files that were using `!!!.aside`
- Adds new tests for Aside
- Removes the old asides.js transform as its no longer needed
- Removes the code from markdown.js that made `!!!.aside` work.


MichaelSolati (approved):

site/_shortcodes/Aside.js
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Load SVG icons to be injected into the page.
+ * @param {string} name The filename (without extension) of the svg icon to
+ * load.
+ * @return {string} The SVG file contents.
+ */
+const loadIcon = name =>
+  fs.readFileSync(`site/_includes/icons/${name}.svg`, 'utf-8');
+
```
devnook, line 12: Maybe this could be a helper function since other components will likely need icons too.

robdodson: Good idea. Because we're a little tight on time today I think i'm going to go ahead and land this and we can look at extracing the util out in a separate PR

PR #381, opened 2021-02-06
bump chrome-types to fix beta/stable versions of APIs, availability

### LINKED ISSUE #373: Proxy docs indicate the API is only available for CrOS, but it's cross-platform
https://developer.chrome.com/docs/extensions/reference/proxy/ says

Important: This API works only on Chrome OS

But it's actually available on all platforms (AFAIK).  My _suspicion_ is that this was a bug in the features files parsing code, because "proxy" is a complex feature with multiple entries - one of which is CrOS-only, but the other is all platforms:

https://source.chromium.org/chromium/chromium/src/+/master:chrome/common/extensions/api/_permission_features.json;l=685-692;drc=c26e11e0944aa67f6c4633731fb5c03a6a50ca5f

I haven't double-checked if this was correct on the old server or not.

samthor:
Fixes #373.

We weren't parsing the feature files correctly; now we are. This correctly fixes where APIs are available.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,7 @@
         "async-transforms": "^1.0.6",
         "browser-media-mime-type": "^1.0.0",
         "chokidar-cli": "^2.1.0",
-        "chrome-types": "^0.0.10",
+        "chrome-types": "^0.0.14",
         "common-tags": "^1.8.0",
         "compression": "^1.7.4",
         "concurrently": "^5.2.0",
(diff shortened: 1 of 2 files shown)


robdodson (approved):

robdodson:
@samthor will everyone need to run `npm run types` to keep the site building?

samthor:
No, this isn't a breaking change—your types will just be old/wrong until you do. The deploy will clear and reset them.

(As a side note, I think in future I'll get `npm run types` to check what version of the package we have, and run it if it's out of date, so we can make it just a normal part of the build pipeline.)

PR #390, opened 2021-02-09
Add details component

### LINKED ISSUE #262: Add summary/details from web.dev
https://web.dev/handbook/web-dev-components/#details

devnook:
Fixes #262


robdodson (commented):
Hey Ewa, it's kind of a long story but I think this component needs to be rewritten as the original implementation on web.dev has a lot of issues. Would you be ok with me taking over this PR and after I've made updates I can explain what I did? It's sort of hard to fit into a single comment but I could probably explain it over VC if you want to catch up this week?

robdodson:
@devnook I wanted to mention that I was considering getting rid of the 'Read More' item that's coded into the shortcode because it didn't seem like most folks used it and it felt like it relied on the weird text splitting in order to add it

devnook:
PTAL

samthor (approved):
I might ask Jiwoong to look at the UX of this, but mechanically it's fine.

(The UX issues I have are: is blue confusable with a link? Should the details be in a box?)

PR #396, opened 2021-02-10
remove Flash from FAQ

### LINKED ISSUE #394: Chrome Web Store FAQ lists Flash as supported
**Describe the bug**

These two answers in the Chrome Web Store FAQ are incorrect since Flash has support has ended in Chrome:

> I have a free Flash game, can I host it in the store? 
> Yes. If you want to host it on your server, it is exactly the same as hosting a Flash file on your site. If you create a packaged app, you must include the complete app (with Flash file) in your ZIP file.
> 
> I have a Flash game that I would love to be able to take payments for, how can I do it? 
> Flash games are hosted inside a web page, this means that the payment infrastructure can be handled on the server that hosts the container page. Using the Licensing API makes it easy to determine if a user has paid or not.

**To Reproduce**
1. Go to https://developer.chrome.com/docs/webstore/faq/

samthor:
Fixes #394

@dotproto and @awfuchs In general this page is hilariously outdated and refers to Apps and Chrome versions <10. I would be happy to come up with a whole new list of FAQs with y'all soon.

### DIFF
--- a/site/en/docs/webstore/faq/index.md
+++ b/site/en/docs/webstore/faq/index.md
@@ -473,18 +473,6 @@ can use one of the Google Analytics libraries for standard websites (e.g., [anal
 
 The `.crx` file is currently limited to 2GB.
 
-### I have a free Flash game, can I host it in the store? {: #faq-app-08 }
-
-Yes. If you want to host it on your server, it is exactly the same as hosting a Flash file on your
-site. If you create a packaged app, you must include the complete app (with Flash file) in your ZIP
-file.
-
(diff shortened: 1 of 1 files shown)


dotproto (approved):
Oh, for sure, definitely get rid of those flash references.

If you have thoughts on how to improve this FAQ I think it would probably be best to start by creating an issue to discuss and/or a branch to propose specific changes. Either way, I'm all for it!

dotproto:
I'm going to take the liberty of merging this.

PR #413, opened 2021-02-10
Add data practices field description to CWS privacy page

awfuchs:
This adds to https://developer.chrome.com/docs/webstore/cws-dashboard-privacy/ the recently-introduced privacy disclosure fields.

### DIFF
--- a/site/en/docs/webstore/cws-dashboard-privacy/index.md
+++ b/site/en/docs/webstore/cws-dashboard-privacy/index.md
@@ -68,7 +68,8 @@ policies that you provide to the Chrome Web Store.
 group of checkboxes to disclose which types of data your extension collects. Use the second group of
 checkboxes to certify that you comply with each of the disclosure statements.
 
-{% Img src="image/SHhb2PDKzXTggPGAYpv8JgR81pX2/PwmNuP356TsVi80bxWAj.png", alt="ALT_TEXT_HERE", width="800", height="495" %}
+{% Img src="image/SHhb2PDKzXTggPGAYpv8JgR81pX2/PwmNuP356TsVi80bxWAj.png", alt="Screenshot of data
+use certification fields", width="800", height="495" %}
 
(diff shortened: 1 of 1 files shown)


xanderblondin (approved):

site/en/docs/webstore/cws-dashboard-privacy/index.md
```diff
@@ -53,5 +53,22 @@ does not and select "No, I am not using remote code."
...
+* The nature of the data that the extensions collects from users
+* Your certification that the extension complies with the policy on limited use
+
+Your disclosures are displayed to Chrome users, and should be consistent with the existing privacy
+policies that you provide to the Chrome Web Store.
+
+
+**Data usage** Use this field to disclose and certify your data collection practices. Use the first
+group of checkboxes to disclose which types of data your extension collects. Use the second group of
+checkboxes to certify that you comply with each of the disclosure statements.
```
dotproto, line 69: This formatting strikes me as a bit odd. Specifically, we have inline bold text that appears to be used as a header immediately followed by an independent sentence. I feel like this should at least be separated by a colon.

dotproto (changes requested):

site/en/docs/webstore/cws-dashboard-privacy/index.md
awfuchs: It's effectively a run-in heading as in the existing material above (like https://screenshot.googleplex.com/at3yQiPwjnARSNg for example). It's just not very well formatted because of the minimalist markdown usage.

I can see doing this with em dashes like we did below the vision pillars https://developer.chrome.com/docs/extensions/mv3/intro/platform-vision/#where-were-headed but maybe it would look better to stick an em space in there in this case.

awfuchs: Updated with em spaces. It looks better now.

dotproto (approved):
LGTM :shipit:

PR #415, opened 2021-02-11
Build and cache types using 11ty's cache helper

samthor:
We currently need to run `npm run types` at several points throughout our build system. This removes that instead places the step inside a cached bit of data.

To be clear, this depends explicitly on unpkg.com and will fetch us the latest version of types. This is not significantly different/riskier from what we do in other ways:
- we hit omahaproxy for release data
- we hit Twitter for the latest tweet

I'm happy to be convinced otherwise though. The fallback is that we just dynamically generate a cached version of a local checked-out version of chrome-types.

This PR also updates the version of `typedoc` to 0.20.x, which fixes a bunch of issues but needs a bit of rewriting of the parsing code.


robdodson (approved):

codedfiles (commented):
Duplicate of #

PR #416, opened 2021-02-11
Remove references to d.c.c in getting started guide

### LINKED ISSUE #410: Getting started mention the extension only works on developer.chrome.com
**Describe the bug**

https://developer.chrome.com/docs/extensions/mv3/getstarted/ mentions on two occasions:

_-> Near the intro_ 

This tutorial will build an extension that allows the user to change the background color of **any page on developer.chrome.com.** It will use many core components to give an introductory demonstration of their relationships.

_-> In the paragraph 'Layer logic'_

The extension **now knows the popup should be available to users on developer.chrome.com** and displays a colored button but needs logic for further user interaction.

The text in ( https://github.com/GoogleChrome/developer.chrome.com/blob/main/site/en/docs/extensions/mv3/getstarted/index.md  )  should either be rephrased the this extension works on any webpage, or text should be added that it only works on developer.chome.com.

**To Reproduce**
1. Create the extension as proposed in the tutorial 
2. Go to any other website (e.g. google.com)
3. Trigger the extension
4. It is expected that this will only work on developer.chrome.com however it works on any website.

dotproto:
Fixes #410


samthor (approved):

site/en/docs/extensions/mv3/getstarted/index.md
```diff
@@ -296,10 +297,10 @@ function setPageBackgroundColor() {
 }
 ```
 
-The updated code adds an `onclick` event on the button, which triggers a [programmatically injected
-content script][24]. This turns the background color of the page the same color as the button. Using
-programmatic injection allows for user-invoked content scripts, instead of auto inserting unwanted
-code into web pages.
+The updated code adds a `click` event listener to the button, which triggers a [programmatically
+injected content script][24]. This turns the background color of the page the same color as the
+button. Using programmatic injection allows for user-invoked content scripts, instead of auto
```
samthor, line 302: I it might be worth elaborating on why "auto inserted unwanted code" is bad.

Is the point here that extensions used to historically allow running always on every single page? Is it worth mentioning the permission that let them do that, and how it's discouraged? Could be good place to put an `{% Aside %}`.

site/en/docs/extensions/mv3/getstarted/index.md
```diff
@@ -272,9 +273,9 @@ Reload the extension to view the green button.
 
 ## Layer logic {: #logic }
 
-The extension now knows the popup should be available to users on [developer.chrome.com][23] and
-displays a colored button but needs logic for further user interaction. Update `popup.js` by adding
-the following to the end of the file.
+The extension now has a custom icon, a popup, and colors the popup button based on a value saved to
```
awfuchs, line 276: It's strange grammar to say it has a thing, a thing, and verb phrase. Consider:
...has a custom icon and a popup. and it colors the...

awfuchs (approved):

PR #422, opened 2021-02-11
Add warning about reusing email from deleted CWS account

awfuchs:
After you delete your CWS publisher account, you can't reuse the associated email to create a new account (for a while). Added a note about this. Ref: b/173543004


xanderblondin:
Can we add a mention that they should check their email regularly to ensure that they are up to date with:

1. Policy updates.
2. Rejection/take-down notices

xanderblondin (approved):

PR #424, opened 2021-02-11
Add payments deprecation banner

awfuchs:
Orphaned page http://localhost:8080/docs/webstore/money/ needed an updated deprecation banner.

Replaced content of page warning banner. The gist was: Chrome apps are going away. Now: CWS payments is deprecated.


dotproto (changes requested):
I like the new prompt, but I think we still need to keep language about Chrome Apps being deprecated as the page still discusses apps throughout the body copy.

dotproto (approved):

PR #427, opened 2021-02-12
simplify types parser due to chrome-types update

samthor:
I'm updating the generating of chrome-types a bit.

* Old https://unpkg.com/chrome-types@0.0.15/index.d.ts
* New: https://unpkg.com/chrome-types@0.0.16-beta.0/index.d.ts

I'm removing &lt;p>&lt;/p> where possible and fixing links before we publish the .d.ts files (otherwise there's lots of relative links which don't make sense in VSCode and friends).

Additionally, TypeDoc sources are technically Markdown, so parse them as markdown.

⚠️ This is marked "DO NOT MERGE" only because it needs to happen in lockstep with making the next version of the types public. I don't think we'll have any more major "format" changes like this (and the site _would_ still work if these go in out of sync, but it might look weird).


mohamedkofta:
> نشر المعاينة _لمطور chrome-com_ جاهز!
> 
> بنيت مع الالتزام [f22ca19](https://github.com/GoogleChrome/developer.chrome.com/commit/f22ca19fd5979f319199e8d56f829f093d044b20)
> 
> https://deploy-preview-427--developer-chrome-com.netlify.app

tools/types/build.js
```diff
@@ -14,7 +14,9 @@
  * limitations under the License.
  */
 
-const chromeTypesRepoURL = 'https://unpkg.com/chrome-types@latest/';
+// FIXME: remove before submit
```
MichaelSolati, line 17: Should this be removed?

MichaelSolati (approved):

tools/types/build.js
samthor: Yup, I will. This was for testing because the contents of this file have changed but I've not yet marked the new version at "latest".

This should be the last time we have to do this.

PR #432, opened 2021-02-14
fix: docs mv2 options event listener example

theskillwithin:
fix code example in docs

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener


theskillwithin:
@googlebot I signed it!

samthor (approved):

PR #440, opened 2021-02-16
link redirect to same page

iamarpitpatidar:
The removed link redirect to the same page (https://developer.chrome.com/docs/apps/contentSecurityPolicy/) and there is no change in any info provided on both pages.

- A link redirecting to the exact same page as users are reading the docs


iamarpitpatidar:
@googlebot I signed it!

dotproto (changes requested):
Thanks for the PR. This looks like a pretty straight forward improvement. Could you also update the PR to remove link [4] from the footer of the page?

PR #445, opened 2021-02-17
awkwardly support extra permissions info

### LINKED ISSUE #111: API permissions are not accurately specified.
Checkout https://developer.chrome.com/docs/extensions/reference/declarativeNetRequest/. The Permissions section lists the permission as "declarativeNetRequest". 

Compare this with the old documentation at https://3-72-0-dot-chrome-apps-doc.appspot.com/extensions/declarativeNetRequest which lists the permissions as 

> declarativeNetRequest
> declarativeNetRequestFeedback
> host permissions
> 
> Note that declarativeNetRequestFeedback and host permissions should only be specified when necessary.

The permission section was pulled in from the template at [link](https://source.chromium.org/chromium/chromium/src/+/master:chrome/common/extensions/docs/templates/private/intro_tables/declarative_net_request_permissions.html?q=%22declarativeNetRequestFeedback%22%20file:html) in this case. This is likely also a bug for the other html template files in that directory.

samthor:
Fixes #111. There's only three useful cases of this in the old codebase.

(There's a comment on [`socket`](https://3-72-0-dot-chrome-apps-doc.appspot.com/apps/socket), but that API is deprecated in two different ways, and the comment isn't repeated on its peers `sockets.udp` and etc).

This is so rare that it's not worth doing more here. As an aside, I'm very sad I can't set the content of blocks from within Markdown (which would have been much nicer than doing it the front-matter).

### DIFF
--- a/site/en/docs/extensions/reference/declarativeNetRequest/extra_permissions.md
+++ b/site/en/docs/extensions/reference/declarativeNetRequest/extra_permissions.md
@@ -1,0 +1,3 @@
+Hello, I'm an extra *permission*
+
+Am I [linked](https://google.com) to stuff?
--- a/site/en/docs/extensions/reference/declarativeNetRequest/extra_permissions.md
+++ b/site/en/docs/extensions/reference/declarativeNetRequest/extra_permissions.md
@@ -1,3 +1,0 @@
-Hello, I'm an extra *permission*
-
-Am I [linked](https://google.com) to stuff?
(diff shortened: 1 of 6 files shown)


site/en/docs/extensions/reference/declarativeNetRequest/extra_permissions.md
```diff
@@ -0,0 +1,3 @@
+Hello, I'm an extra *permission*
+
+Am I [linked](https://google.com) to stuff?
```
devnook, line 3: Why is this here?

devnook (approved):

mamun52610 (approved):

site/en/docs/extensions/reference/declarativeNetRequest/extra_permissions.md
samthor: This was a first attempt at including Markdown directly in another page. I forgot to remove it!

PR #449, opened 2021-02-18
default to inline images, don't display missing user images

### LINKED ISSUE #367: Need an easy way to handle inline images
Expected (from the old site):

image

Actual (on the new site):

image

My preferred approach would be to use the `{% Img %}` shortcode in both scenarios, passing some sort of flag when we need the image inline.

samthor:
Fixes #367.

### DIFF
--- a/site/_includes/macros/post-author.njk
+++ b/site/_includes/macros/post-author.njk
@@ -1,23 +1,31 @@
 {#
-Renders an author info snippet.
+  Renders an author info snippet.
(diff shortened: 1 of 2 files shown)


MichaelSolati (approved):

MichaelSolati:
It looks like this causes an ever so slight shift to the layout, but honestly, i don't think it matters.

samthor:
That is interesting. I might see if only setting `p img` to `inline-block` works.

PR #458, opened 2021-02-23
build Eleventy as part of build/test/lint

samthor:
This has a commit which introduces a bug we'd like to catch. Just testing that out.


MichaelSolati (approved):

.github/workflows/lint-and-test.yml
```diff
@@ -1,11 +1,11 @@
-name: Lint and test
+name: Lint, test and build
```
MichaelSolati, line 1: Should the markdown file name change?

samthor: Maybe; I think building Eleventy is kind of a test.

samthor: Renamed to check.

PR #463, opened 2021-02-24
Add pullapprove

robdodson:
- Removes CODEOWNERS
- Adds pullapprove config. Sets extensions team as owners for their docs with Simeon and Andy being default reviewers. All other docs fallback to the content team. Eng team owns all programmatic files.


robdodson:
Before merging this we'll need to tweak some repo settings to turn off the required CODEOWNERS review and to set pullapprove as a required check.

.pullapprove.yml
```diff
@@ -0,0 +1,83 @@
...
+      # but skip them when sending review requests.
+      - ~dcc-extensions
+      - ~dcc-content
+      - ~dcc-eng
+      - ~dcc-approvers
+    reviews:
+      # Assign all possible reviewers
+      request: 99
+    conditions:
+      - >
```
samthor, line 24: It seems like these rules only exist to basically say _NOT_ the old nacl or mobile client docs. Is that right?

samthor (approved):

.pullapprove.yml
robdodson: Essentially. Though when we add devtools, workbox, etc, we'll probably need a similar block for them.

btw here's Angular's pullapprove file. They're like expert mode: https://github.com/angular/angular/blob/master/.pullapprove.yml

Probably some good ideas we can steal from them.

PR #364, opened 2021-02-03
updates types infra, add version data

samthor:
Fixes #362
Fixes #360
Fixes #39

PR #465, opened 2021-02-25
Include version data for types

### LINKED ISSUE #39: New site is missing API version information
I found there is no "Since Chrome _version_" in new api reference. In the past, there are many statement say like "Since Chrome 48". This information is useful for compatibility. Developers can detect features or set "minimum_chrome_version" in manifest.

### LINKED ISSUE #360: Many new apis listing as Stable APIs. Is it a bug?
Sorry, I don't use issue template. I just want to confirm the situation first.

I find there are `app.runtime`, `app.window`,  `socket`, `sockets.tcp`, `sockets.tcpServer`, `sockets.udp` (maybe more) listing on the [API Reference](https://developer.chrome.com/docs/extensions/reference/) page. Is these a bug? I remember they are chrome app api, not extension api.

In addition, I find `processes` api listing as Stable APIs, but I remember it is Dev APIs in the past. Is this a bug?

### LINKED ISSUE #362: Chrome Extension API chrome.tabGroups is Stable but not accessible in Stable Channel
When attempting to load an extension into the Stable build of Chrome with the manifest permissions including "tabGroups", an error stating "'tabGroups' requires dev channel or newer, but this is the stable channel." occurs. However, the Chrome extension API lists "tabGroups" as stable [here](https://developer.chrome.com/docs/extensions/reference/). Shouldn't the tabGroups API be accessible in the stable build?

**To Reproduce**
1. Build a chrome extension as shown [here ](https://developer.chrome.com/docs/extensions/mv3/getstarted/)
2. In the manifest.json of a chrome extension, under "permissions" include ["tabGroups"].
3. At [chrome://extensions/](chrome://extensions/), click Load Unpacked and select the application just created. 
4. Click "Errors" under the newly loaded extension.
5. One of the reported errors should be a warning stating "'tabGroups' requires dev channel or newer, but this is the stable channel."

**Expected behavior**
Any calls to the chrome.tabGroups API should be valid according to https://developer.chrome.com/docs/extensions/reference/tabGroups/ because the API should be stable. Currently, any calls to the chrome.tabGroups API are undefined.

There should not be a warning that tabGroups requires the dev channel because according to the documentation, the tabGroups API should be included in the stable channel already.

**Screenshots**
image

 - OS: Mac
 - Browser: Chrome 
 - Version 88.0.4324.146 (Official Build) (x86_64)

samthor:
Fixes #362
Fixes #360
Fixes #39

This supersedes #364 and only focuses on types, not comment rewriting code. There's really a few big parts to this PR:

- a helper which enumerates through all types so we can update them with version data
- code which only includes the difference from the parent (if the parent was released at 75, don't say it again)
- rendering code to display the types.

It also adds "tools/types/index.js" but just for convenience to test running the types.

This does _not_ add tests, but I'm thinking about that separately. The concern is that we are depending on live data, and I actually _want_ to test that.

### DIFF
--- a/site/_includes/layouts/namespace-reference.njk
+++ b/site/_includes/layouts/namespace-reference.njk
@@ -78,11 +78,11 @@
               </div>
             </li>
           {% endif %}
-          {% if namespace.channel !== 'stable' or namespace.version.low %}
+          {% if namespace.version.channel !== 'stable' or namespace.version.low %}
             <li>
               <div class="code-sections__label">Availability</div>
               <div>
-                {% if namespace.channel === 'dev' %}
+                {% if namespace.version.channel === 'dev' %}
                   Dev channel only.
                 {% elif namespace.version.channel === 'beta' %}
                   Beta and Dev channels only.
--- a/site/_includes/layouts/namespace-reference.njk
+++ b/site/_includes/layouts/namespace-reference.njk
@@ -78,14 +78,17 @@
               </div>
             </li>
           {% endif %}
-          {% if namespace.channel !== 'stable' %}
+          {% if namespace.channel !== 'stable' or namespace.version.low %}
             <li>
               <div class="code-sections__label">Availability</div>
               <div>
                 {% if namespace.channel === 'dev' %}
                   Dev channel only.
-                {% else %}
+                {% elif namespace.version.channel === 'beta' %}
                   Beta and Dev channels only.
+                {% else if namespace.version.low %}
+                  Since Chrome {{ namespace.version.low }}.
+                  {# No namespace is currently deprecated. #}
                 {% endif %}
               </div>
             </li>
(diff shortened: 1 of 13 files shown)


site/_includes/layouts/namespace-reference.njk
```diff
@@ -78,14 +78,17 @@
...
+          {% if namespace.channel !== 'stable' or namespace.version.low %}
             <li>
               <div class="code-sections__label">Availability</div>
               <div>
                 {% if namespace.channel === 'dev' %}
                   Dev channel only.
-                {% else %}
+                {% elif namespace.version.channel === 'beta' %}
                   Beta and Dev channels only.
+                {% else if namespace.version.low %}
```
MichaelSolati, line 89: elif?

MichaelSolati (approved):

site/_includes/layouts/namespace-reference.njk
samthor: Fixed

PR #469, opened 2021-02-26
feat: promote use of csp directive frame-ancestors

kyranjamie:
Extension authors should be encouraged to use the [`frame-ancestors`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors) CSP directive.

### DIFF
--- a/site/en/docs/extensions/mv3/security/index.md
+++ b/site/en/docs/extensions/mv3/security/index.md
@@ -120,7 +120,7 @@ scripting attacks. If the extension only loads resources from itself register th
   "name": "Very Secure Extension",
   "version": "1.0",
   "description": "Example of a Secure Extension",
-  "content_security_policy": "default-src 'self'"
+  "content_security_policy": "default-src 'self'; frame-ancestors 'none';",
   "manifest_version": 2
 }
 ```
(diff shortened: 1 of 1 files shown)


kyranjamie:
@googlebot I signed it!

samthor:
I'm confused, when would extension content be iframed?

cc @dotproto for more comments

kyranjamie:
> when would extension content be iframed?

It wouldn't. That's why it should be prohibited.

samthor:
You're right, but this seems to already be disallowed. I can't iframe
content from local extensions; I get a failure to load and a sadface icon.
(I'm not an expert. I am still going to defer to Simeon.)

dotproto:
It is possible for pages to load extension resources in an iframe if the page is exposed via [web_accessible_resources](https://developer.chrome.com/docs/extensions/mv3/manifest/web_accessible_resources/), which is currently undergoing some updates for MV3.

Like Sam, I'm not clear on what additional protections declaring frame-ancestors would have beyond what the extension platform already provides by default.

kyranjamie:
Sorry for delay getting back but I missed these replies cc/ @samthor @dotproto 

> It seems like it's more work for no benefit

I should have explained better. I opened this PR as a security auditing firm shared a clickjacking attack on an extension I've worked on. Without the `frame-ancestors` directive, an attacker can open up an extension in an `iframe`, dupe a user into enabling clipboard sharing permissions, and then, say, pull a users cryptocurrency seed phrase they've just copied, in an app like Metamask.

Not alerting developers to this would be negligent.

samthor (approved):

PR #493, opened 2021-03-04
details fixes

### LINKED ISSUE #448: Clean up dummy copy for Details component
https://developer.chrome.com/docs/handbook/components/#details

This section is showing placeholder copy. It doesn't match the rest of the components page. It should have a description of the shortcode with an example of how to use it.

samthor:
Fixes #448.

This also makes two style changes:

- we put a hairline at top/bottom of details so you can see the boundaries
- we make details stacks so correct spacing appears in their content.

### DIFF
--- a/site/_includes/partials/toc-inner.njk
+++ b/site/_includes/partials/toc-inner.njk
@@ -4,9 +4,12 @@
 
   <div class="toc-container display-block xl:display-none">
     <details class="toc" role="navigation" aria-label="{{ 'i18n.common.toc' | i18n(locale) }}">
-      <summary class="surface display-inline-flex align-center color-secondary-text user-select-none">
-        <div class="type--h6">{{ 'i18n.common.toc' | i18n(locale) }}</div>
-        <span class="toc__icon">{{ icon('arrow-right') }}</span>
+      <summary class="surface color-secondary-text user-select-none">
+        {# This is nested as Safari fails to render `display: inline-flex` on <summary>. #}
+        <div class="display-inline-flex align-center">
+          <div class="type--h6">{{ 'i18n.common.toc' | i18n(locale) }}</div>
+          <span class="toc__icon">{{ icon('arrow-right') }}</span>
+        </div>
       </summary>
       <div class="toc__wrapper">{{ tocContents | safe }}</div>
     </details>
--- a/site/_scss/blocks/_details.scss
+++ b/site/_scss/blocks/_details.scss
@@ -1,11 +1,9 @@
 .details {
-  position: relative;
-
   border: 1px solid var(--color-hairline);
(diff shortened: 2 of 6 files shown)


dotproto:
It looks like details elements are currently rendering with an extra (default) arrow in Chrome 89. 

Screen Shot 2021-03-04 at 3 35 03 PM

dotproto:
It looks like the rendering issue I mentioned is now present in Chrome Stable

* https://developer.chrome.com/docs/handbook/components/
* https://deploy-preview-493--developer-chrome-com.netlify.app/docs/handbook/components

samthor:
@dotproto PTAL, I fixed some styling issues. I think Chrome rev'ed and stopped using the webkit-prefixed selector.

dotproto:
Tested [this page](https://deploy-preview-493--developer-chrome-com.netlify.app/docs/handbook/components) again on Stable 89 and Canary 91. Both versions look good.

site/_includes/partials/toc-inner.njk
```diff
@@ -4,9 +4,12 @@
 
   <div class="toc-container display-block xl:display-none">
     <details class="toc" role="navigation" aria-label="{{ 'i18n.common.toc' | i18n(locale) }}">
-      <summary class="surface display-inline-flex align-center color-secondary-text user-select-none">
-        <div class="type--h6">{{ 'i18n.common.toc' | i18n(locale) }}</div>
-        <span class="toc__icon">{{ icon('arrow-right') }}</span>
+      <summary class="surface color-secondary-text user-select-none">
+        {# This is nested as Safari fails to render `display: inline-flex` on <summary>. #}
```
dotproto, line 8: Yay, comments! Maybe it would be helpful to call out version of Safari this was observed on?

site/_scss/blocks/_details.scss
```diff
@@ -19,13 +26,11 @@
     background-color: var(--color-blue-lighter);
     border: 1px solid  var(--color-blue-medium);
   }
-
 }
 
-.details__header {
+.details > summary > .details__header {
```
dotproto, line 31: Is this necessary? The naming scheme appears to be BEM, but AFAIK BEM tries to avoid this kind of hierarchical selector pattern. This comment also applies to other similar selector patterns in the PR.

samthor: I agree that it's breaking BEM, but think it's worth overriding here as the `details` and `summary` elements have a very specific layout which I want to match.

site/_includes/partials/toc-inner.njk
samthor: bleh. I can fix this later, I submitted accidentally.

PR #500, opened 2021-03-04
Don't title case "Table of contents"

robdodson:
I noticed the ToC title cases "Table of Contents" which seems a bit odd.


samthor (approved):

jpmedley:
@robdodson This looks good. I should make you aware of a best practice that applies here. In good writing you try to omit needless words. "Contents" would be sufficient for this heading. I don't feel strongly about it in this instance though.

robdodson:
thank you @jpmedley 👍

PR #501, opened 2021-03-05
Manifest version info

### LINKED ISSUE #495: add MV2,MV3 or ALL flags to API Reference
**Problems faced by developers**

Some extension APIs are MV2 only or MV3 only, and some APIs are available on both MV2 and MV3. For example:

- chrome.scripting is MV3 only vs chrome.tabs.executeScript is MV2 only.
- chrome.tabGroups is MV3 only
- chrome.browserAction is available on both MV2 and MV3

There is no flags to identify these characteristic.

Like https://packagecontrol.io/ , there are flags to show which Sublime Text version can use a plugin, e.g. `ST2`, `ST3`, `ALL`. There are more similar examples like package control.

The document don't need to add flags to all APIs, but it should add flags to some new and some old APIs which are only available on specific Manifest version.

**Additional context**
Here are two discussions on the forum:
https://groups.google.com/a/chromium.org/g/chromium-extensions/c/2fyCMDhVgQY/
https://groups.google.com/a/chromium.org/g/chromium-extensions/c/q8E56RrWpo4/

samthor:
Fixes #495.

Also simplifies stable/beta/dev: this is only ever for the whole API, so don't include it on the individual methods (we didn't display this anyway).


hanguokai:
Thanks for updates! I see there are MV version flags on chrome._namespace_ . But some individual apis in a chrome._namespace_ are missing this information, like `chrome.tabs.executeScript` is not available on MV3 since it replaced by `chrome.scripting.executeScript`.

samthor:
Thanks for pointing this out, and I see the data here:
https://chromium.googlesource.com/chromium/src/+/master/chrome/common/extensions/api/_api_features.json#810

This probably won't get fixed in the next few days, my apologies, but I'll
try to get the data visible soon.

MichaelSolati (approved):

samthor:
Merging this anyway; it's better than nothing right now.

PR #502, opened 2021-03-05
Fix an outdated comment

robdodson:
Just updates an old comment. Instead of using a plugin, we factored optimize-html into two transforms (purify-css and minify-html)

### DIFF
--- a/site/_includes/layouts/base.njk
+++ b/site/_includes/layouts/base.njk
@@ -14,7 +14,7 @@
     <link rel="preload" as="font" crossorigin href="/fonts/google-sans-v2003/medium/latin.woff2">
 
     {% if process.env.NODE_ENV === 'production' %}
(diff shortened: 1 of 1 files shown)


site/_includes/layouts/base.njk
```diff
@@ -14,7 +14,7 @@
     <link rel="preload" as="font" crossorigin href="/fonts/google-sans-v2003/medium/latin.woff2">
 
     {% if process.env.NODE_ENV === 'production' %}
-      {# CSS will be injected by optimize-html plugin. #}
+      {# CSS will be injected by purify-js plugin. #}
```
robdodson, line 17:
```suggestion
      {# CSS will be injected by purify-css plugin. #}
```

MichaelSolati (approved):

PR #505, opened 2021-03-05
Snapshot pages that change less frequently.

robdodson:
Helps a bit with our issue where Percy is snapshotting live data.

I removed the /blog/ snapshot because it changes too often and is confusing for folks.
I switched to snapshotting the nacl landing page instead of the extensions landing page, and two posts which we don't expect to change often.

### DIFF
--- a/.percy.js
+++ b/.percy.js
@@ -16,11 +16,11 @@ module.exports = {
     'snapshot-files': [
       'index.html',
       'releases/index.html',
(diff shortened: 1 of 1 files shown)


robdodson:
I've gone ahead and approved the percy snapshots for the new pages. I'll wait for Sam to get back from vacation to review.

samthor (approved):

PR #511, opened 2021-03-08
Update index.md

ghzhou:
Fixes #SOME_ISSUE_NUMBER

-
-
-


samthor:
Is there a specific issue?

dotproto:
@samthor, the page currently says

> You may want to modularize your event handling code by dividing it across multiple JavaScript files. In this case, you must import these scripts from within the main service worker file:
> 
> {
>     ...
>     ...
>   }

@ghzhou is correct that we should have included importScripts here, but the suggested change doesn't go far enough IMO, but since they haven't agreed to the CLA we can't merge anyway. I think I'm going to create a separate fix & PR.

PR #532, opened 2021-03-17
Fixed example snippet

### LINKED ISSUE #531: Sample codes from "Extending Devtools - injecting a content script" don't work
**Describe the bug**
The [sample code snippets](https://developer.chrome.com/docs/extensions/mv2/devtools/#injecting) from the "Extending Devtools - injecting a content script" fail to inject a script.

**To Reproduce**
Follow the guide and copy/paste those nippets.

**Expected behavior**
The content script should be injected

 - OS: OSX
 - Browser chrome
 - Version 89

**Additional context**
When trying this sample code it wasn't working as is. Sending the message through the connection instead of `chrome.runtime.sendMessage()` solved the issue. I'm still new to chrome extensions but I believe this was the intention, otherwise please disregard this issue.

pmenichelli:
Fixes #531

When trying this sample code it wasn't working as is. Sending the message through the connection solved the issue. I'm still new to chrome extensions but I believe this was the intention, otherwise please disregard this suggestion.

### DIFF
--- a/site/en/docs/extensions/mv3/devtools/index.md
+++ b/site/en/docs/extensions/mv3/devtools/index.md
@@ -164,7 +164,7 @@ backgroundPageConnection.onMessage.addListener(function (message) {
 });
 
 // Relay the tab ID to the background page
-chrome.runtime.sendMessage({
+backgroundPageConnection.postMessage({
     tabId: chrome.devtools.inspectedWindow.tabId,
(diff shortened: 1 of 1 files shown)


pmenichelli:
@googlebot I signed it!

dotproto (approved):
Thanks for catching this!

PR #533, opened 2021-03-17
Initial baseline draft for promises page

awfuchs:
Adds promises page

### DIFF
--- a/site/en/docs/extensions/mv3/promises/index.md
+++ b/site/en/docs/extensions/mv3/promises/index.md
@@ -176,7 +176,7 @@ promise. We don't set the `lastError` in the success case, because the onFulfill
 argument in the `.then())` is never invoked in the failure case.
 
 {% Aside %}
-Extensions APIs don't set `chrome.runtime.lastError()` with promises.
+Extensions APIs don't set `chrome.runtime.lastError` with promises.
 {% endAside %}
 
 This lets you write async function calls like sync function calls, verifying the failure cases (the
--- a/site/en/docs/extensions/mv3/promises/index.md
+++ b/site/en/docs/extensions/mv3/promises/index.md
@@ -166,7 +166,7 @@ chrome.tabs.create({...})
   .then((result) => {
     // success case
   })
-  .catch((e) => {
+  .catch((error) => {
     // failure case
   });
 ```
--- a/site/en/docs/extensions/mv3/promises/index.md
+++ b/site/en/docs/extensions/mv3/promises/index.md
@@ -170,12 +170,13 @@ Promises are designed to deliver asynchronous results, both success and failure.
 promise (a promise rejection) is handled differently.  It might look like this:
 
 ```
-chrome.tabs.create({...}).then((result) => {
-
-  // success case
-}).catch((e) => {
-  // failure case
-});
+chrome.tabs.create({...})
+  .then((result) => {
+    // success case
+  })
+  .catch((e) => {
+    // failure case
+  });
 ```
 
 Instead of setting `chrome.runtime.lastError`, we deliver the error directly to the "catch" of the
--- a/site/en/docs/extensions/mv3/promises/index.md
+++ b/site/en/docs/extensions/mv3/promises/index.md
@@ -53,10 +53,14 @@ to the use case.
 
 Not all methods in extensions APIs support promises. Sometimes that's because we haven't added
 promise support on the method yet; in many cases it's because using a promise isn't feasible for the
-method. You can check if an API method support promises by checking its API reference page:
+method.
+
+{% if false %}
+You can check if an API method support promises by checking its API reference page:
 
 {% Img src="image/SHhb2PDKzXTggPGAYpv8JgR81pX2/lDjGweNENKyqGSA2nZrp.png", alt="Screenshot of an API
 reference page with promises support noted", width="800", height="467" %}
+{% endif %}
 
 ## How to use promises
 
--- a/site/en/docs/extensions/mv3/promises/index.md
+++ b/site/en/docs/extensions/mv3/promises/index.md
@@ -159,7 +159,6 @@ execution).  The extension would look at the last error like this:
 
 ```js
 chrome.tabs.create({...}, (result) => {
-
   if (chrome.runtime.lastError) {
     // Handle last error
   }
--- a/site/en/docs/extensions/mv3/promises/index.md
+++ b/site/en/docs/extensions/mv3/promises/index.md
@@ -44,7 +44,7 @@ Promises are not available for extensions using Manifest V2, and are not availab
 
 Promises can and should be used in many circumstances. However, there are times (for example, event
 listeners) when a promise won't work and a callback is more appropriate. Methods that support
-promises also support callbacks when invoked asynchronously.
+promises also support callbacks for backwards compatibility reasons.
 
 ## Can I use promises?
 
--- a/site/en/docs/extensions/mv3/promises/index.md
+++ b/site/en/docs/extensions/mv3/promises/index.md
@@ -157,7 +157,7 @@ chrome.tabs.create({...}, (result) => {
 Promises are designed to deliver asynchronous results, both success and failure.  A failure in a
 promise (a promise rejection) is handled differently.  It might look like this:
 
-```
+```js
 chrome.tabs.create({...})
   .then((result) => {
     // success case
--- a/site/en/docs/extensions/mv3/promises/index.md
+++ b/site/en/docs/extensions/mv3/promises/index.md
@@ -123,6 +123,7 @@ Extensions APIs don't set `chrome.runtime.lastError()` with promises.
 This lets you write async function calls like sync function calls, verifying the failure cases (the
 errors) in the same way.
 
+{% if false %}
 ## Dealing with edge cases
 
 There are a few cases where you might need to take some extra care when converting an existing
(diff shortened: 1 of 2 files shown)


site/en/docs/extensions/mv3/promises/index.md
```diff
@@ -0,0 +1,149 @@
...
+Instead of setting `chrome.runtime.lastError`, we deliver the error directly to the "catch" of the
+promise. We don't set the `lastError` in the success case, because the onFulfilled handler (the
+argument in the `.then())` is never invoked in the failure case.
+
+{% Aside %}
+Extensions APIs don't set `chrome.runtime.lastError()` with promises.
+{% endAside %}
+
+This lets you write async function calls like sync function calls, verifying the failure cases (the
+errors) in the same way.
```
dotproto, line 124: This looks like it's referring async/await, but it kind of comes out of nowhere. Also, here's an async/await version of the snippet you've been working with so far.

```js
try {
  let result = await chrome.tabs.create({...});
  // success case
} catch (error) {
  // failure case
}
```

Be aware that `await` is only valid in async functions and the top level bodies of modules. (This phrasing is taken directly from the SyntaxError Chrome reports when you try to use await somewhere you can't.)

dotproto (changes requested):
Added a code sample and a handful of notes. I don't think the current sample is a great fit for the article. We should probably sync to discuss approaches in the near future.

site/en/docs/extensions/mv3/promises/index.md
```diff
@@ -0,0 +1,147 @@
...
+
+You can and should use promises in your extension code, where a promise is available and appropriate
+to the use case.
+
+Not all methods in extensions APIs support promises. Sometimes that's because we haven't added
+promise support on the method yet; in many cases it's because using a promise isn't feasible for the
+method. You can check if an API method support promises by checking its API reference page:
+
+{% Img src="image/SHhb2PDKzXTggPGAYpv8JgR81pX2/lDjGweNENKyqGSA2nZrp.png", alt="Screenshot of an API
+reference page with promises support noted", width="800", height="467" %}
```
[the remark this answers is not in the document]
awfuchs: I don't consider this a production screenshot anyway, with the scribble text on it. I think we need to have an actual working thing before we can be done here.

solomonkinard (approved):
LGTM

solomonkinard (approved):
lgtm

site/en/docs/extensions/mv3/promises/index.md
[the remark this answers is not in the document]
awfuchs: @samthor Is there a rendering coming for promises in these API pages? I'd like to replace the placeholder mock under https://deploy-preview-533--developer-chrome-com.netlify.app/docs/extensions/mv3/promises/#can-i-use-promises with an actual screenshot...

site/en/docs/extensions/mv3/promises/index.md
```diff
@@ -0,0 +1,147 @@
...
+
+---
+
+Many extension API methods support promises.  This document explains how to use promises when
+calling these methods.
+
+## Introduction
+
+Promises were introduced into Chrome not long after they were included in the ES6 specification.
+They are an important feature of the modern JavaScript idiom, providing several benefits, such as:
```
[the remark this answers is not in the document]
awfuchs: Simplified to avoid the term.

site/en/docs/extensions/mv3/promises/index.md
```diff
@@ -0,0 +1,147 @@
...
+
+# A list of authors. These usernames correspond to the keys in the
+# _data/authorsData.json file.
+
+---
+
+Many extension API methods support promises.  This document explains how to use promises when
+calling these methods.
+
+## Introduction
```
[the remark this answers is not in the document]
awfuchs: Done.

site/en/docs/extensions/mv3/promises/index.md
```diff
@@ -0,0 +1,147 @@
...
+## How to use promises
+
+There are many places where using promises will result in cleaner, easier-to-maintain code. You
+should consider using promises in situations such as the following:
+
+* Any time that you want to clean up your code by using a more "synchronous" invocation style.
+* Where error handling would be too difficult using callbacks.
+* When you want a simpler way to invoke a number of concurrent methods and gather the results into a single thread of code.
+
+* When you can’t
```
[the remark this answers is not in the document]
awfuchs: Sure you can... Done.

site/en/docs/extensions/mv3/promises/index.md
awfuchs: Added separate async/await section at end of doc.

site/en/docs/extensions/mv3/promises/index.md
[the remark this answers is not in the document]
awfuchs: Sam replied on another thread that this is coming soon (in about n days) but is not ready yet. So I've suppressed the screenshot & pointer for the moment and we'll reinstate that when there is actual rendering on the API docs for a screenshot.

site/en/docs/extensions/mv3/promises/index.md
```diff
@@ -0,0 +1,214 @@
...
+Extensions can use promises beginning with Manifest V3. Many API methods support promises, and we
+are progressively adding promise support to additional API methods.
+
+{% Aside 'gotchas' %}
+Promises are not available for extensions using Manifest V2, and are not available on all API methods.
+{% endAside  %}
+
+Promises can and should be used in many circumstances. However, there are times (for example, event
+listeners) when a promise won't work and a callback is more appropriate. Methods that support
+promises also support callbacks when invoked asynchronously.
```
dotproto, line 47:
```suggestion
promises also support callbacks for backwards compatibility reasons.
```

"when invoked asynchronously" – JavaScript doesn't allow you to choose whether a method is invoked synchronously or asynchronously. I'm not quite sure what this clause is attempting to address, so my current inclination is to drop it. Alternatively, we could provide more clarification on why we support both (as I've suggested here).

site/en/docs/extensions/mv3/promises/index.md
```diff
@@ -0,0 +1,214 @@
...
+
+Promises are designed to deliver asynchronous results, both success and failure.  A failure in a
+promise (a promise rejection) is handled differently.  It might look like this:
+
+```js
+chrome.tabs.create({...})
+  .then((result) => {
+    // success case
+  })
+  .catch((e) => {
```
dotproto, line 169:
```suggestion
  .catch((error) => {
```

Rename for clarity. While 'e' or 'err' are often used for error variable names, I think it's best to err on the side of being explicit here.

site/en/docs/extensions/mv3/promises/index.md
```diff
@@ -0,0 +1,214 @@
...
+    // failure case
+  });
+```
+
+Instead of setting `chrome.runtime.lastError`, we deliver the error directly to the "catch" of the
+promise. We don't set the `lastError` in the success case, because the onFulfilled handler (the
+argument in the `.then())` is never invoked in the failure case.
```
dotproto, line 176: Pinged Tim for clarification on how Chrome handles `lastError` in promises, as the current copy left a bit of a gap for me. 

The second sentence seems a little confused. Is `lastError` referring to `chorme.runtime.lastError` or the argument passed into the `.catch()` handler? It's also not clear to me why the onFufilled handler (this is the first time that name has appeared, might be better to say "success handler"?) is mentioned.

EDIT: I tend to read pretty linearly and I only just say the below aside. I think this paragraph can be simplified.

site/en/docs/extensions/mv3/promises/index.md
```diff
@@ -0,0 +1,214 @@
...
+
+{% Aside %}
+Extensions APIs don't set `chrome.runtime.lastError()` with promises.
```
dotproto, line 179:
```suggestion
Extensions APIs don't set `chrome.runtime.lastError` with promises.
```

It's a property, not a method.

site/en/docs/extensions/mv3/promises/index.md
```diff
@@ -0,0 +1,214 @@
...
+{% endAside %}
+
+This lets you write async function calls like sync function calls, verifying the failure cases (the
+errors) in the same way.
```
dotproto, line 183: I think it may be better to say that "This lets you write async logic in a more synchronous style". 

> …verifying the failure cases (the errors) in the same way.

I don't this this is accurate. Typically errors (e.g. `throw new Error('some description')`) are handled with try/catch. Promises do allow you to _throw_ errors in a synchronous style (throwing inside an async function or a `new Promise(…)` call will route the error throught he promise chain), but _handling_ them requires the use of a error handler in the promise chain.

async/await (specifically async) is the language feature that allows you to use try/catch with promises. That's why above I suggested saying "in a more synchronous style".

dotproto (approved):
Added some suggestions that I think we should integrate. Flagging as approve in order to not bock you from merging once these are resolved.

site/en/docs/extensions/mv3/promises/index.md
awfuchs: Pruned this for simplicity in 9e0faef.

site/en/docs/extensions/mv3/promises/index.md
awfuchs: Modified and simplified this along the suggested lines in 9e0faef.

PR #534, opened 2021-03-17
Update browserAction to action for MV3 docs

tomasdev:
Fixes #SOME_ISSUE_NUMBER

-
-
-

### DIFF
--- a/site/en/docs/extensions/mv3/manifest/activeTab/index.md
+++ b/site/en/docs/extensions/mv3/manifest/activeTab/index.md
@@ -43,7 +43,7 @@ See the [Page Redder][2] sample extension:
     "scripts": ["background.js"],
     "persistent": false
   },
-  "browser_action": {
+  "action": {
     "default_title": "Make this page red"
   },
   "manifest_version": 3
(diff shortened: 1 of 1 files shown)


dotproto:
Thanks for the suggestion. This is a good change, but it only covers half of the API update. We also need to update `manifest.json` to move from `browser_action` to `action`. Could you update line 46 as well?

dotproto (changes requested):
Thanks for the suggestion. This is a good change, but it only covers half of the API update. We also need to update `manifest.json` to move from `browser_action` to `action`. Could you update line 46 as well?

tomasdev:
@dotproto done!

site/en/docs/extensions/mv3/manifest/activeTab/index.md
```diff
@@ -46,13 +46,13 @@ See the [Page Redder][2] sample extension:
   "browser_action": {
```
dotproto, line 46:
```suggestion
  "action": {
```

`browser_action` has been replaced by `action` in MV3.

dotproto (changes requested):
Good catch, manifest_version also needed to be updated. One more change and I think we're good.

site/en/docs/extensions/mv3/manifest/activeTab/index.md
tomasdev: Oops!

dotproto (approved):
LGTM, thanks again!

PR #537, opened 2021-03-18
Change enterprise.platformKeys usage example

omorsi:
Currently SHA-1, SHA-256, SHA-384 and SHA-512 are supported in enterprise.platformKeys API as hash algorithms that can be used for signing using RSA keys. Here's a [link](https://source.chromium.org/chromium/chromium/src/+/master:chrome/browser/extensions/api/platform_keys/platform_keys_api_ash.cc;l=308?q=platform.*keys%20sign%20f:ash&ss=chromium) to the related chromium code. Although all of them are supported, but SHA-1 is the weakest. This PR changes the API usage example to SHA-256 instead of SHA-1 as it is stronger and not cryptographically broken.


dotproto (approved):
Simple and straightforward. No objections here :)

PR #291, opened 2021-01-22
Migrate the DevTools docs

kaycebasques:

PR #546, opened 2021-03-23
Devtools2

samthor:
Supersedes #291 but only adds minor fixes.

- various TOC links were invalid (to 404 pages)
- server was modified to serve on 8081, fixed now

cc @petele


samthor:
Going to submit this as discussed.

@petele It's possible that there are still orphan pages that @kaycebasques has imported (or _not_ imported, and they're just on the old site) that are not referenced in the TOC.

But otherwise it LGTM.

Issue #90, opened 2020-12-17
Add reusable tabs component


robdodson:
@robdodson commented on [Tue Oct 27 2020](https://github.com/GoogleChrome/dcc-private/issues/184)

There are multiple places where we'll need to use a tabs component in our markup. I think we might be able to steal the one from web.dev

robdodson:
This is implemented in web.dev as a pretty complex shortcode and I think it might be good for us to revisit that and see if we can simplify it.

It's not uncommon for authors to want to do tabs for comparing code in different languages or versions of a library.

dotproto:
Off the top of my head, a few extensions use cases come to mind.

* Promises & callbacks snippets
* MV2 & MV3 manifest/scripts
* Multiple files shown together in a walkthrough, especially for embedding the final version in the page.

PR #563, opened 2021-03-24
Examples of common Tabs API operations

dotproto:
Adds 3 examples of common operations that developers want to perform to the Tabs API reference page.

### DIFF
--- a/site/en/docs/extensions/reference/tabs/index.md
+++ b/site/en/docs/extensions/reference/tabs/index.md
@@ -8,7 +8,7 @@ extra_permissions_html:
 
 You can use most `chrome.tabs` methods and events without declaring any permissions in the
 extension's [manifest][manifest] file. However, if you require access to the [`url`][prop-url],
-[`pendingUrl`][3], [`title`][prop-title], or [`favIconUrl`][prop-favIconUrl] properties of
+[`pendingUrl`][3], [`title`][prop-title], or [`favIconUrl`][prop-faviconurl] properties of
 [`tabs.Tab`][tab], you must declare the `"tabs"` permission in the manifest, as shown below:
 
 ```json
@@ -94,9 +94,9 @@ directory of the [chrome-extensions-samples][samples-repo] repository.
 
 [manifest]: /docs/extensions/mv3/manifest/
 [prop-url]: #property-Tab-url
-[prop-pendingUrl]: #property-Tab-pendingUrl
+[prop-pendingurl]: #property-Tab-pendingUrl
 [prop-title]: #property-Tab-title
-[prop-favIconUrl]: #property-Tab-favIconUrl
+[prop-faviconurl]: #property-Tab-favIconUrl
 [tab]: #type-Tab
 [mv2-tabs-samples]: https://github.com/GoogleChrome/chrome-extensions-samples/tree/master/mv2-archive/api/tabs/
 [samples-repo]: https://github.com/GoogleChrome/chrome-extensions-samples
--- a/site/en/docs/extensions/reference/tabs/index.md
+++ b/site/en/docs/extensions/reference/tabs/index.md
@@ -24,11 +24,57 @@ permission in the manifest, as shown below:
 
 ## Examples
 
+Opening the extension's onboarding page in new tab after the extension is installed for the first
+time. Must be called from a background context.
+
+```js
+chrome.runtime.onInstalled.addListener((reason) => {
+  if (reason === chrome.runtime.OnInstalledReason.INSTALL) {
+    chrome.tabs.create({
+      url: 'onboarding.html'
+    });
+  }
+});
+```
+
+Retrieve the user's currently focused tab. Requires Manifest V3. Cannot be used in a content
+script.
+
+```js
+async function getCurrentTab() {
+  let queryOptions = { active: true, currentWindow: true };
+  let [tab] = await chrome.tabs.query(queryOptions);
+  return tab;
+}
+```
+
+Content script asking a background script to navigate the current tab.
+
+```js
+// Content script
+function requestNavigation(url) {
+  chrome.runtime.sendMessage({ action: 'navigate', url });
+}
+
+// Background script
+chrome.runtime.onMessage.addListener(messageListener);
+
+function messageListener(request, sender, _sendResponse) {
+  if (request.action === 'navigate') {
+    // Sanitize URL in to guard against compromised content scripts
+    let url = new URL(request.url);
+    chrome.tabs.update(sender.tab.id, { url: url.toString() });
+  }
+}
+```
+
+### More samples
+
 ![Two tabs in a window](tabs.png)
 
-You can find simple examples of manipulating tabs with the `chrome.tabs` API in the
-[examples/api/tabs][7] directory. For other examples and for help in viewing the source code, see
-[Samples][8].
+More samples simple examples of manipulating tabs with the Tabs API can be found in the
+[mv2-archive/api/tabs][7] directory of the [chrome-extensions-samples][samples-repo] repository. For
+other examples and for help in viewing the source code, see [Samples][8].
 
 [1]: /docs/extensions/mv2/tabs
 [2]: #property-Tab-url
@@ -38,3 +84,5 @@ You can find simple examples of manipulating tabs with the `chrome.tabs` API in 
 [6]: #type-Tab
 [7]: https://github.com/GoogleChrome/chrome-extensions-samples/tree/master/mv2-archive/api/tabs/
 [8]: /docs/extensions/mv2/samples
+
+[samples-repo]: https://github.com/GoogleChrome/chrome-extensions-samples
--- a/site/en/docs/extensions/reference/tabs/index.md
+++ b/site/en/docs/extensions/reference/tabs/index.md
@@ -90,8 +90,7 @@ function toggleMuteState(tabId) {
 ### More samples
 
 For more examples that demonstrate the Tabs API, see the [mv2-archive/api/tabs][mv2-tabs-samples]
-directory of the [chrome-extensions-samples][samples-repo] repository. For other examples and for
-help in viewing the source code, see [Samples][mv2-samples].
+directory of the [chrome-extensions-samples][samples-repo] repository.
 
 [manifest]: /docs/extensions/mv3/manifest/
 [prop-url]: #property-Tab-url
@@ -100,5 +99,4 @@ help in viewing the source code, see [Samples][mv2-samples].
 [prop-favIconUrl]: #property-Tab-favIconUrl
 [tab]: #type-Tab
 [mv2-tabs-samples]: https://github.com/GoogleChrome/chrome-extensions-samples/tree/master/mv2-archive/api/tabs/
-[mv2-samples]: /docs/extensions/mv2/samples
 [samples-repo]: https://github.com/GoogleChrome/chrome-extensions-samples
--- a/site/en/docs/extensions/reference/tabs/index.md
+++ b/site/en/docs/extensions/reference/tabs/index.md
@@ -7,9 +7,9 @@ extra_permissions_html:
 ## Manifest
 
 You can use most `chrome.tabs` methods and events without declaring any permissions in the
-extension's [manifest][manifest] file. However, if you require access to the [`url`][prop-url], [`pendingUrl`][3],
-[`title`][prop-title], or [`favIconUrl`][prop-favIconUrl] properties of [`tabs.Tab`][tab], you must declare the `"tabs"`
-permission in the manifest, as shown below:
+extension's [manifest][manifest] file. However, if you require access to the [`url`][prop-url],
+[`pendingUrl`][3], [`title`][prop-title], or [`favIconUrl`][prop-favIconUrl] properties of
+[`tabs.Tab`][tab], you must declare the `"tabs"` permission in the manifest, as shown below:
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


site/en/docs/extensions/reference/tabs/index.md
```diff
@@ -24,11 +24,57 @@ permission in the manifest, as shown below:
...
+chrome.runtime.onInstalled.addListener((reason) => {
+  if (reason === chrome.runtime.OnInstalledReason.INSTALL) {
+    chrome.tabs.create({
+      url: 'onboarding.html'
+    });
+  }
+});
+```
+
+Retrieve the user's currently focused tab. Requires Manifest V3. Cannot be used in a content
```
samthor, line 40: Why does it need V3? Purely because it uses `Promise`?

site/en/docs/extensions/reference/tabs/index.md
```diff
@@ -24,11 +24,57 @@ permission in the manifest, as shown below:
...
+function requestNavigation(url) {
+  chrome.runtime.sendMessage({ action: 'navigate', url });
+}
+
+// Background script
+chrome.runtime.onMessage.addListener(messageListener);
+
+function messageListener(request, sender, _sendResponse) {
+  if (request.action === 'navigate') {
+    // Sanitize URL in to guard against compromised content scripts
```
samthor, line 64: I mean.. we construct any URL we like from a string input. I don't think this counts as sanitizing.

site/en/docs/extensions/reference/tabs/index.md
dotproto: Yep, exactly. TBH I'm a bit conflicted here. On one hand I want to provide snippets that are as broadly useful as possible, on the other I want to encourage modern language features. Perhaps it's enough if I just clarify this line to say "Requires Manifest V3 due to use of promises"?

site/en/docs/extensions/reference/tabs/index.md
dotproto: The idea was that an owned content script might try to pass in a non-URL payload in order to exploit either the extension's or Chrome's handling of this value. I thought that since `new URL()` will throw on invalid URLs, it could provide some basic sanitization. 

I did just did a little experimentation, though, and it looks like this constructor is more permissive than I originally thought. For example, I assumed new URL would throw on `javascript:alert(1)` or `data:text/html,<doctype html><script>alert(1)</script>` but neither cause the constructor to throw.

In general I want to encourage developers to think about how their inputs could be abused, especially messages from content scripts. If you have any thoughts on how to accomplish this I'd love to hear 'em.

site/en/docs/extensions/reference/tabs/index.md
dotproto: I should note that I also considered displaying both a promise & callback version of this snippet, but decided against it because it took up too much vertical space. Another use case in support of #90.

site/en/docs/extensions/reference/tabs/index.md
tophf: There was an article by someone from Chromium on CORB topic noting that instead of sending full URLs in a message from a content script extensions should only pass an id or a path which will be expanded to a full URL in the background script. Like, this is the only way to make sure the background script doesn't act like an open proxy once the content script is compromised.

site/en/docs/extensions/reference/tabs/index.md
samthor: We talked about this; we want this to be a going forward document, so I think `Promise` only (with a note) makes sense.

Over the next year I presume we'll stop including the note as much.

site/en/docs/extensions/reference/tabs/index.md
samthor: So I thought about this a bit more: _is_ there a security risk? This is just telling the user's browser to navigate somewhere, no different than if they typed it in the URL bar itself.

Like this is a pointless extension, I could just type in the URL. But I don't think this actually has a security problem. Am I missing something?

site/en/docs/extensions/reference/tabs/index.md
dotproto: Yep, I think that makes sense.

site/en/docs/extensions/reference/tabs/index.md
tophf: It could be a restricted chrome:// URL like `chrome://crash` or other dangerous ones listed in `chrome://chrome-urls` page.

dotproto: That's a good point, @tophf, thanks for calling that out.

@samthor, it's not as big an issue as, say, transmitting credit card info over HTTP, but it's still very much part of the extensions threat model. Content scripts run in the page's process, not the extensions. If a malicious actor is able to compromise the page's process, they also gains access to whatever capabilities extensions have on the page. That's why we limit what APIs are exposed to content scripts.

In that spirit, I want to encourage developers not to assume that data form content scripts can be trusted, and to think about ways to limit the severity of a potential compromise. To that end, I'm not as interested in the specifics of this example as I concise ways to demonstrate this line of thinking.

site/en/docs/extensions/reference/tabs/index.md
```diff
@@ -24,11 +24,57 @@ permission in the manifest, as shown below:
 
 ## Examples
 
+Opening the extension's onboarding page in new tab after the extension is installed for the first
+time. Must be called from a background context.
```
awfuchs, line 28:
```suggestion
A common pattern for extensions is to open an onboarding page in a new tab when the extension is installed. The following example shows how to do this. {% Aside %} Note that this code must be called from a background context, because tabs cannot use `chrome.tabs.create()`{% end Aside %}.
```

awfuchs, line 28:
```suggestion
The following sections demonstrate several common use cases for the chrome.tabs API.

### Opening an extension page in a new tab

A common pattern for extensions is to open an onboarding page in a new tab when the extension is installed. The following example shows how to do this. {% Aside %} Note that this code must be called from a background context, because tabs cannot use `chrome.tabs.create()`{% end Aside %}.
```

site/en/docs/extensions/reference/tabs/index.md
awfuchs, line 40:
```suggestion
### Get the current tab

This example demonstrates how the background script can retrieve the currently focused tab. {% Aside %} Requires Manifest V3. Cannot be used in a content script. {% end Aside %}

{# Editor's note: what happened about converting this example to a promise with a note? #}
```

site/en/docs/extensions/reference/tabs/index.md
```diff
@@ -24,11 +24,57 @@ permission in the manifest, as shown below:
...
+  if (reason === chrome.runtime.OnInstalledReason.INSTALL) {
+    chrome.tabs.create({
+      url: 'onboarding.html'
+    });
+  }
+});
+```
+
+Retrieve the user's currently focused tab. Requires Manifest V3. Cannot be used in a content
+script.
```
awfuchs, line 41:
```suggestion
```

site/en/docs/extensions/reference/tabs/index.md
```diff
@@ -24,11 +24,57 @@ permission in the manifest, as shown below:
...
+
+```js
+async function getCurrentTab() {
+  let queryOptions = { active: true, currentWindow: true };
+  let [tab] = await chrome.tabs.query(queryOptions);
+  return tab;
+}
+```
+
+Content script asking a background script to navigate the current tab.
```
awfuchs, line 51:
```suggestion
### Request background script to navigate tab

This example shows how a content script could ask the background script to navigate the current tab.
{# Editor's note: I'd like to frame this as a use case; is there an obvious one? #}
```

site/en/docs/extensions/reference/tabs/index.md
```diff
@@ -24,11 +24,57 @@ permission in the manifest, as shown below:
...
+
+```js
+// Content script
+function requestNavigation(url) {
+  chrome.runtime.sendMessage({ action: 'navigate', url });
+}
+
+// Background script
+chrome.runtime.onMessage.addListener(messageListener);
```
awfuchs, line 60:
```suggestion
  chrome.runtime.sendMessage({ action: 'navigate', url });
}
<three back ticks>

<three back ticks>js
// Background script
chrome.runtime.onMessage.addListener(messageListener);
```

site/en/docs/extensions/reference/tabs/index.md
```diff
@@ -24,11 +24,57 @@ permission in the manifest, as shown below:
...
+    // Sanitize URL in to guard against compromised content scripts
+    let url = new URL(request.url);
+    chrome.tabs.update(sender.tab.id, { url: url.toString() });
+  }
+}
+```
+
+### More samples
+
 ![Two tabs in a window](tabs.png)
```
awfuchs, line 73: I don't see the image on the preview...

site/en/docs/extensions/reference/tabs/index.md
```diff
@@ -24,11 +24,57 @@ permission in the manifest, as shown below:
...
 
-You can find simple examples of manipulating tabs with the `chrome.tabs` API in the
-[examples/api/tabs][7] directory. For other examples and for help in viewing the source code, see
-[Samples][8].
+More samples simple examples of manipulating tabs with the Tabs API can be found in the
```
awfuchs, line 75:
```suggestion
For more examples that demonstrate the Tabs API, see the
```

awfuchs (commented):
Suggested some additional headings, main section intro text, and rewording. Also recommend splitting the // background.js and // contentscript.js into separately fenced boxes as a standard. Adding the headings helps disambiguate this.

site/en/docs/extensions/reference/tabs/index.md
dotproto: That image wasn't migrated in the transition to the new site. For the moment I'm just going to delete the image reference.

site/en/docs/extensions/reference/tabs/index.md
dotproto: Replaced this demo with a muting a given tab. I'd like to come back to this section to add a better message passing demo, but at the moment I no good use cases are leaping to mind. If anyone has thoughts, please add a suggestion to issue #665.

site/en/docs/extensions/reference/tabs/index.md
```diff
@@ -24,17 +24,79 @@ permission in the manifest, as shown below:
...
+```
+
+### Get the current tab
+
+This example demonstrates how the background script can retrieve the currently focused tab.
+
+{% Aside %}
+Requires Manifest V3 due to the use of Promises. Cannot use `tabs.query` in content
+scripts.
+{% endAside %}
```
dotproto, line 57: > what happened about converting this example to a promise with a note?

@awfuchs, you included the above editor's note in a previous version, but I didn't quite follow the comment. The code uses Promises by awaiting the result of `chrome.tabs.query(queryOptions)`. Your suggested change also called out that the sample requires MV3. In a subsequent commit I clarified this line to say "Requires Manifest V3 due to the use of Promises."

Does this address your concerns?

site/en/docs/extensions/reference/tabs/index.md
```diff
@@ -24,17 +24,79 @@ permission in the manifest, as shown below:
...
+//// background.js
+
+function toggleMuteState(tabId) {
+  chrome.tabs.get(tabId, async (tab) => {
+    let muted = !tab.mutedInfo.muted;
+    await chrome.tabs.update(tabId, { muted });
+    console.log(`Tab ${tab.id} is ${ muted ? 'muted' : 'unmuted' }`);
+  });
+}
+```
```
dotproto, line 88: > I'd like to frame this as a use case; is there an obvious one? 

FYI @awfuchs, I replaced the navigation request example with this new example and opened #665 to revisit a message passing example in the future.

site/en/docs/extensions/reference/tabs/index.md
```diff
@@ -24,17 +24,79 @@ permission in the manifest, as shown below:
...
-[4]: #property-Tab-title
-[5]: #property-Tab-favIconUrl
-[6]: #type-Tab
-[7]: https://github.com/GoogleChrome/chrome-extensions-samples/tree/master/mv2-archive/api/tabs/
-[8]: /docs/extensions/mv2/samples
+A common pattern for extensions is to open an onboarding page in a new tab when the extension is
+installed. The following example shows how to do this.
+
+{% Aside %}
+Cannot use `chrome.tabs.create()` in content scripts.
```
awfuchs, line 35: Is a sentence with no subject. Consider any of:

- Content scripts cannot use `chrome.tabs.create()`
- `chrome.tabs.create()` is not available in content scripts.
- The `chrome.tabs.create()` method is not available in content scripts.

site/en/docs/extensions/reference/tabs/index.md
```diff
@@ -24,17 +24,79 @@ permission in the manifest, as shown below:
...
+  }
+});
+```
+
+### Get the current tab
+
+This example demonstrates how the background script can retrieve the currently focused tab.
+
+{% Aside %}
+Requires Manifest V3 due to the use of Promises. Cannot use `tabs.query` in content
```
awfuchs, line 55: Are two sentences with no subject. (One less egregious.) Isn't "due to the use of promises" an implementation detail? Consider:

This example requires Manifest V3. Note also that the `tabs.query()` method isn't available in content scripts.

site/en/docs/extensions/reference/tabs/index.md
awfuchs: Yes, I left new comments.

awfuchs (approved):

PR #564, opened 2021-03-24
Follow-up tweaks to promises page

awfuchs:
Some additional cleanup of the new promises page:

- Missing heading: There's a heading "Error handling with callbacks" that introduces the callback version, but the promises version that follows it has no intervening heading.

- There are a few opportunities for minor phrasing improvements.

### DIFF
--- a/site/en/docs/extensions/mv3/promises/index.md
+++ b/site/en/docs/extensions/mv3/promises/index.md
@@ -29,7 +29,7 @@ operation. For more about promises and their use, see the MDN documentation on
 ## Introduction
 
 Promises were introduced into Chrome not long after they were included in the ES6 specification.
-They are an important feature of modern JavaScript, providing several benefits, such as:
+They are an important feature of modern JavaScript, providing benefits such as:
 
 * Streamlined error handling
 * Coding in a synchronous style for invoking asynchronous functions
@@ -140,7 +140,8 @@ function openTabOnRight() {
 
 ### Error handling
 
-Returning errors works differently depending on if the extension is using a callback or a promise.
+Returning errors works differently depending on whether the extension is using a callback or a
+promise.
 
 #### Error handling with callbacks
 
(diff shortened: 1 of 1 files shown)


site/en/docs/extensions/mv3/promises/index.md
```diff
@@ -158,6 +157,8 @@ chrome.tabs.create({...}, (result) => {
 });
 ```
 
+#### Error handling with promises
+
 Promises are designed to deliver asynchronous results, both success and failure.  A failure in a
 promise (a promise rejection) is handled differently.  It might look like this:
```
chromatim, line 165: We may want to mention after this example something like "Alternatively, the error handler can be passed as a second argument to the .then() method".

chromatim (changes requested):
Added a comment about the alternative way error handling can be done on a .then(), do you think that's worth including?

site/en/docs/extensions/mv3/promises/index.md
samthor: I'd rather link to e.g. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise than get into the business of fully documenting `Promise`.

awfuchs: Done. PTAL.

samthor (approved):

site/en/docs/extensions/mv3/promises/index.md
```diff
@@ -171,18 +172,21 @@ chrome.tabs.create({...})
   });
 ```
 
-Instead of setting `chrome.runtime.lastError`, we deliver the error directly to the "catch" of the
-promise.
+Extensions APIs don't set `chrome.runtime.lastError` when you use a promise, instead returning the
```
chromatim, line 175: Is "returning" fine here grammatically? It reads to me like it should be something like "they return".

chromatim (approved):
Looks good to me, per the one grammar question I brought up

site/en/docs/extensions/mv3/promises/index.md
awfuchs: It's OK grammatically but not optimal (thinking non-native speakers, localization, etc.)

Rephrased.

PR #566, opened 2021-03-25
add warnings to ISSUE_TEMPLATE

samthor:
We're getting a few issues about Chrome/Chromium, not the site itself.


dotproto (approved):
I like it.

PR #567, opened 2021-03-25
Don't resize SVGs

### LINKED ISSUE #557: Images not shown properly
**Describe the bug**
It seems that images are not shown properly

**To Reproduce**
1. Go to 'https://developer.chrome.com/docs/devtools/evaluate-performance/'
2. Scroll down and watch those images with a blue frame highlight.
3. See error

**Expected behavior**
Images shown.

**Screenshots**
image

 - OS: macOS Big Sur 11.2.3 (20D91)
 - Browser Chrome
 - Version 89.0.4389.82 (Official Build) (x86_64)

**Additional context**
No

samthor:
Fixes #557.

See comment, but basically, imgix can't render these SVGs. They don't really make sense as SVGs to begin with, though: the embedded PNGs clearly have a known fixed size, so ideally we would re-render them as purely a PNG and then change them all.

This fixes the immediate issue though.

### DIFF
--- a/site/_shortcodes/Img.js
+++ b/site/_shortcodes/Img.js
@@ -50,22 +50,31 @@ const Img = function (args) {
     lazy = true;
   }
 
-  // https://docs.imgix.com/apis/rendering
-  params = {auto: 'format', ...params};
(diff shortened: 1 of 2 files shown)


chybie (approved):
LGTM

PR #568, opened 2021-03-25
Remove CWS payments section of FAQ

awfuchs:
Removes CWS payments section of FAQ because it's FAQ of a deprecated feature and contains broken links that cannot be resolved today.

- b/143559554


site/en/docs/webstore/faq/index.md
```diff
@@ -345,85 +345,6 @@ address in Play and CWS. If there is an exact match, the "Available for Android"
 automatically show up on your item detail page the next time the script runs. At this point, there
 is no way to enter the information into your listing.
 
-## Payments and license server
```
CharyCh, line 348: the payment is deprecated, but the licensing API is not. Can you rewrite this part instead of deleting all?

awfuchs: As agreed offline, the mentions in this section about licensing API are not useful. One discusses how to verify payment status using the licensing API, the other just says the licensing API is almost finished.
So it should be OK to delete this section.

CharyCh (approved):

Issue #576, opened 2021-03-27
Links from MV2 docs to MV3 docs are not specific enough


cashpipeplusplus:
**Describe the bug**
Links from MV2 docs to MV3 should ideally link to a corresponding MV3 doc.  For example, the [manifest format doc for MV2](https://developer.chrome.com/docs/extensions/mv2/manifest/) should link to the [corresponding doc for MV3](https://developer.chrome.com/docs/extensions/mv3/manifest/), not [the general overview of MV3](https://developer.chrome.com/docs/extensions/mv3/intro/) as it does today.

**Expected behavior**
It would be nice if someone could audit all of the links to https://developer.chrome.com/docs/extensions/mv3/intro/ and replace them with more helpful links to the correct docs.

AmySteam:
Fixed in #6872

PR #590, opened 2021-04-01
Link MV2 docs to their MV3 counterparts

dotproto:
Issue #576 noted that MV2 docs don't actually link to the MV3 versions of the same material. This PR aims to address that by adding some additional text to the deprecation warning that links the user to the appropriate MV3 page.

The initial implementation here is largely for discussion purposes. Credit to @robdodson on most of the heavy lifting here.

### DIFF
--- a/site/_includes/partials/extensions/mv2-legacy-page.md
+++ b/site/_includes/partials/extensions/mv2-legacy-page.md
@@ -1,5 +1,17 @@
+{% set found = false %}
+{% set mv3url = page.url | replace('/mv2/', '/mv3/') %}
+{% for item in collections.all %}
+  {% if item.url == mv3url %}
+    {% set found = true %}
+  {% endif %}
+{% endfor %}
+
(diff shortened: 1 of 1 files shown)


robdodson:
A few thoughts:

- There's no way to `break` a loop in nunjucks so this snippet will always loop through 1500+ pages for every mv2 page, which feels inefficient.
- Alternatively there are 35 pages of MV2 documentation by my count (correct me if I'm wrong), so we could instead put the preferred mv3 link into the YAML frontmatter of those 35 pages. This would be a bit of manual work upfront (maybe an hour of work?) but doesn't introduce complexity or inefficiencies. I'd vote for this approach if folks are ok with it.

samthor:
I'm with Rob. I have another idea too but I think just putting a manual link in is best for now.

dotproto:
I discussed this with @awfuchs and he expressed interest in taking over this change. He may start fresh from a new PR for this effort.

PR #596, opened 2021-04-02
Update Storage API examples

dotproto:
https://deploy-preview-596--developer-chrome-com.netlify.app/docs/extensions/reference/storage/#examples

### DIFF
--- a/site/en/docs/extensions/mv2/content_scripts/index.md
+++ b/site/en/docs/extensions/mv2/content_scripts/index.md
@@ -152,14 +152,83 @@ They can include JavaScript files, CSS files or both. All auto-run content scrip
 }
 ```
 
-<table class="simple"><tbody><tr><th>Name</th><th>Type</th><th>Description</th></tr><tr id="matches"><td><code>matches</code></td><td>array of strings</td><td><em>Required.</em> Specifies which pages this content script will be injected into. See <a href="/docs/extensions/mv2/match_patterns">Match Patterns</a> for more details on the syntax of these strings and <a href="#matchAndGlob">Match patterns and globs</a> for information on how to exclude URLs.</td></tr><tr id="css"><td><code>css<code></code></code></td><td>array of strings</td><td><em>Optional.</em> The list of CSS files to be injected into matching pages. These are injected in the order they appear in this array, before any DOM is constructed or displayed for the page.</td></tr><tr id="js"><td><code>js<code></code></code></td><td><nobr>array of strings</nobr></td><td><em>Optional.</em> The list of JavaScript files to be injected into matching pages. These are injected in the order they appear in this array.</td></tr><tr id="match_about_blank"><td><code>match_about_blank</code></td><td>boolean</td><td><em>Optional.</em> Whether the script should inject into an <code>about:blank</code> frame where the parent or opener frame matches one of the patterns declared in <code>matches</code>. Defaults to false.</td></tr></tbody></table>
+<table class="simple">
+  <tbody>
+    <tr>
+      <th>Name</th>
+      <th>Type</th>
+      <th>Description</th>
+    </tr>
+    <tr id="matches">
+      <td><code>matches</code></td>
+      <td>array of strings</td>
+      <td><em>Required.</em> Specifies which pages this content script will be injected into. See <a
+          href="/docs/extensions/mv2/match_patterns">Match Patterns</a> for more details on the
+        syntax of these strings and <a href="#matchAndGlob">Match patterns and globs</a> for
+        information on how to exclude URLs.</td>
+    </tr>
+    <tr id="css">
+      <td><code>css<code></code></code></td>
+      <td>array of strings</td>
+      <td><em>Optional.</em> The list of CSS files to be injected into matching pages. These are
+        injected in the order they appear in this array, before any DOM is constructed or displayed
+        for the page.</td>
+    </tr>
+    <tr id="js">
+      <td><code>js<code></code></code></td>
+      <td>
+        <nobr>array of strings</nobr>
+      </td>
+      <td><em>Optional.</em> The list of JavaScript files to be injected into matching pages. These
+        are injected in the order they appear in this array.</td>
+    </tr>
+    <tr id="match_about_blank">
+      <td><code>match_about_blank</code></td>
+      <td>boolean</td>
+      <td><em>Optional.</em> Whether the script should inject into an <code>about:blank</code> frame
+        where the parent or opener frame matches one of the patterns declared in
+        <code>matches</code>. Defaults to false.</td>
+    </tr>
+  </tbody>
+</table>
 
 #### Exclude matches and globs {: #matchAndGlob }
 
 Specified page matching is customizable by including the following fields in the manifest
 registration.
 
-<table class="simple"><tbody><tr><th>Name</th><th>Type</th><th>Description</th></tr><tr id="exclude_matches"><td><code>exclude_matches</code></td><td>array of strings</td><td><em>Optional.</em> Excludes pages that this content script would otherwise be injected into. See <a href="/docs/extensions/mv2/match_patterns">Match Patterns</a> for more details on the syntax of these strings.</td></tr><tr id="include_globs"><td><code>include_globs</code></td><td>array of strings</td><td><em>Optional.</em> Applied after <code>matches</code> to include only those URLs that also match this glob. Intended to emulate the <a href="http://wiki.greasespot.net/Metadata_Block#.40include"><code>@include</code> </a>Greasemonkey keyword.</td></tr><tr id="exclude_globs"><td><code>exclude_globs</code></td><td>array of string</td><td><em>Optional.</em> Applied after <code>matches</code> to exclude URLs that match this glob. Intended to emulate the <a href="http://wiki.greasespot.net/Metadata_Block#.40include"><code>@exclude</code> </a>Greasemonkey keyword.</td></tr></tbody></table>
+<table class="simple">
+  <tbody>
+    <tr>
+      <th>Name</th>
+      <th>Type</th>
+      <th>Description</th>
+    </tr>
+    <tr id="exclude_matches">
+      <td><code>exclude_matches</code></td>
+      <td>array of strings</td>
+      <td><em>Optional.</em> Excludes pages that this content script would otherwise be injected
+        into. See <a href="/docs/extensions/mv2/match_patterns">Match Patterns</a> for more details
+        on the syntax of these strings.</td>
+    </tr>
+    <tr id="include_globs">
+      <td><code>include_globs</code></td>
+      <td>array of strings</td>
+      <td><em>Optional.</em> Applied after <code>matches</code> to include only those URLs that also
+        match this glob. Intended to emulate the <a
+          href="http://wiki.greasespot.net/Metadata_Block#.40include"><code>@include</code>
+        </a>Greasemonkey keyword.</td>
+    </tr>
+    <tr id="exclude_globs">
+      <td><code>exclude_globs</code></td>
+      <td>array of string</td>
+      <td><em>Optional.</em> Applied after <code>matches</code> to exclude URLs that match this
+        glob. Intended to emulate the <a
+          href="http://wiki.greasespot.net/Metadata_Block#.40include"><code>@exclude</code>
+        </a>Greasemonkey keyword.</td>
+    </tr>
+  </tbody>
+</table>
 
 The content script will be injected into a page if its URL matches any `matches` pattern and any
 `include_globs` pattern, as long as the URL doesn't also match an `exclude_matches` or
@@ -280,7 +349,43 @@ preffered and default field is `"document_idle"`, but can also be specified as `
 }
 ```
 
-<table class="simple"><tbody><tr><th>Name</th><th>Type</th><th>Description</th></tr><tr id="document_idle"><td><code>document_idle</code></td><td>string</td><td><em>Prefered.</em> Use <code>"document_idle"</code> whenever possible.<br><br>The browser chooses a time to inject scripts between <code>"document_end"</code> and immediately after the <a href="http://www.whatwg.org/specs/web-apps/current-work/#handler-onload"><code>window.onload </code></a>event fires. The exact moment of injection depends on how complex the document is and how long it is taking to load, and is optimized for page load speed.<br><br>Content scripts running at <code>"document_idle"</code> do not need to listen for the <code>window.onload</code> event, they are guaranteed to run after the DOM is complete. If a script definitely needs to run after <code>window.onload</code>, the extension can check if <code>onload</code> has already fired by using the <a href="http://www.whatwg.org/specs/web-apps/current-work/#dom-document-readystate"><code>document.readyState </code></a>property.</td></tr><tr id="document_start"><td><code>document_start</code></td><td>string</td><td>Scripts are injected after any files from <code>css</code>, but before any other DOM is constructed or any other script is run.</td></tr><tr id="document_end"><td><code>document_end</code></td><td>string</td><td>Scripts are injected immediately after the DOM is complete, but before subresources like images and frames have loaded.</td></tr></tbody></table>
+<table class="simple">
+  <tbody>
+    <tr>
+      <th>Name</th>
+      <th>Type</th>
+      <th>Description</th>
+    </tr>
+    <tr id="document_idle">
+      <td><code>document_idle</code></td>
+      <td>string</td>
+      <td><em>Preferred.</em> Use <code>"document_idle"</code> whenever possible.<br><br>The browser
+        chooses a time to inject scripts between <code>"document_end"</code> and immediately after
+        the <a
+          href="http://www.whatwg.org/specs/web-apps/current-work/#handler-onload"><code>window.onload
+          </code></a>event fires. The exact moment of injection depends on how complex the document
+        is and how long it is taking to load, and is optimized for page load speed.<br><br>Content
+        scripts running at <code>"document_idle"</code> do not need to listen for the
+        <code>window.onload</code> event, they are guaranteed to run after the DOM is complete. If a
+        script definitely needs to run after <code>window.onload</code>, the extension can check if
+        <code>onload</code> has already fired by using the <a
+          href="http://www.whatwg.org/specs/web-apps/current-work/#dom-document-readystate"><code>document.readyState
+          </code></a>property.</td>
+    </tr>
+    <tr id="document_start">
+      <td><code>document_start</code></td>
+      <td>string</td>
+      <td>Scripts are injected after any files from <code>css</code>, but before any other DOM is
+        constructed or any other script is run.</td>
+    </tr>
+    <tr id="document_end">
+      <td><code>document_end</code></td>
+      <td>string</td>
+      <td>Scripts are injected immediately after the DOM is complete, but before subresources like
+        images and frames have loaded.</td>
+    </tr>
+  </tbody>
+</table>
 
 #### Specify frames {: #frames }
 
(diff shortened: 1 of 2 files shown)


site/en/docs/extensions/reference/storage/index.md
```diff
@@ -81,50 +82,115 @@ tubes can be filled, and if they are filled when you put your message in, it get
...
+</form>
+
+//// options.js ////
+
+// In-page cache of the user's options
+let options = {};
+
+// Initialize the form with the user's option settings
+chrome.storage.sync.get('options', (data) => {
+  Object.assign(options, data?.options);
```
tophf, line 128: FWIW optional chaining in `data?.options` is unnecessary and looks confusing because `data` will never be null/undefined unless an internal error occurs e.g. a disk failure, for which a proper error handling should be added in your code like maybe checking chrome.runtime.lastError.

site/en/docs/extensions/reference/storage/index.md
```diff
@@ -81,50 +82,115 @@ tubes can be filled, and if they are filled when you put your message in, it get
...
+  options.debug = event.target.checked;
+  chrome.storage.sync.set({options});
+});
+
+//// background.js ////
+
+// Watch for changes to the user's options & apply them
+chrome.storage.onChanged.addListener((changes, area) => {
+  if (area === 'sync' && changes.options) {
+    console.log('enable debug mode?', !!changes.options?.newValue?.debug);
```
tophf, line 143: same here: `changes.options?.` is already known to be truthy so `?.` is redundant and confusing.

site/en/docs/extensions/reference/storage/index.md
```diff
@@ -81,50 +82,115 @@ tubes can be filled, and if they are filled when you put your message in, it get
...
+    setDebugMode(!!changes.options?.newValue?.debug);
```
tophf, line 144: ditto

site/en/docs/extensions/reference/storage/index.md
```diff
@@ -81,50 +82,115 @@ tubes can be filled, and if they are filled when you put your message in, it get
...
+below snippet uses an async `action.onClicked` event handler that waits for the `storageCache`
+global to be populated before executing it's logic.
+
+```js
+//// background.js ////
+
+// Where we will expose all the data we retrieve from storage.sync
+let storageCache = {};
+// Asynchronously retrieve data from storage.sync, then cache it
+let initStorageCache = getAllStorageSyncData().then(items => {
```
tophf, line 160: While the idea looks promising on paper but it doesn't actually work as the API request is made in the same initial task shared by the onClicked listener so this code doesn't make any difference to just reading the storage inside onClicked.

site/en/docs/extensions/reference/storage/index.md
```diff
@@ -81,50 +82,115 @@ tubes can be filled, and if they are filled when you put your message in, it get
...
+  /* normal action handler logic */
+});
+
+// Reads all data out of storage.sync and exposes it via a promise
+//
+// Note: Once the Storage API gains promise support, this function
+// can be greatly simplified.
+function getAllStorageSyncData() {
+  // Immediately return a promise and start asynchronous work
+  return new Promise((resolve, reject) => {
```
tophf, line 176: Why using callbacks for ManifestV3 code if the API returns a Promise inherently?

site/en/docs/extensions/reference/storage/index.md
tophf: I guess this idea can help when performing a heavy synchronous operation at start before `chrome` listener runs:

```js
preloadStorage();
superHeavySynchronousOperation(); // or importScripts('super-big-library.js');
chrome.action.onClicked.addListener(fnUsingPreloadedStorage);
```

An example of such operation would be loading of dependencies but unfortunately this won't help with modern complex apps produced by webpack and such because the main part of a script runs only after all its complex/big dependencies were executed, meaning that storage preloader would run at the end of the task when it's already useless...

In such cases the solution would be to write a wrapper bg.js manually:

```js
try {
  importScripts('preload-storage.js', 'bg-main-webpack-chunk.js');
} catch (e) {
  console.error(e);
}
```

...where preload-storage.js either uses a global `const` or a global `self.something`.

I suggest describing this in detail, maybe even make a demo webpack-based extension, otherwise the idea won't be used properly.

site/en/docs/extensions/mv2/content_scripts/index.md
```diff
@@ -152,14 +152,83 @@ They can include JavaScript files, CSS files or both. All auto-run content scrip
...
+<table class="simple">
```
samthor, line 155: Sorry for a meta-comment, but, these tables aren't specially interesting. Could it be worth converting them to Markdown proper?

site/en/docs/extensions/reference/storage/index.md
```diff
@@ -81,50 +82,115 @@ tubes can be filled, and if they are filled when you put your message in, it get
...
 ```
 
-If you're interested in tracking changes made to a data object, you can add a listener to its
-`onChanged` event. Whenever anything changes in storage, that event fires. Here's sample code to
-listen for saved changes:
+We can take this idea even further. In this example we have an [options page][options-page] that
+allows the user to toggle a "debug mode" in the extension (implementation not shown here). Changes
+to this setting are immediately saved to sync storage by the options page and the background script uses `storage.onChanged` to apply the setting as soon as possible.
 
 ```js
```
samthor, line 110: The highlighting on this is going to be a bit weird. Can you just have three code blocks next to each other, with the correct tag, with their filename still listed as a comment?

site/en/docs/extensions/reference/storage/index.md
samthor: Yes, the underlying types say `data` will always be a real value. So no `?` needed.

site/en/docs/extensions/reference/storage/index.md
```diff
@@ -81,50 +82,115 @@ tubes can be filled, and if they are filled when you put your message in, it get
...
+
+//// options.js ////
+
+// In-page cache of the user's options
+let options = {};
+
+// Initialize the form with the user's option settings
+chrome.storage.sync.get('options', (data) => {
+  Object.assign(options, data?.options);
+  optionsForm.debug.checked = data?.options?.debug;
```
samthor, line 129: technically this would be `data.options?.debug`, _but_, since you've just assigned it to the local `options`: you can make this line:

```js
optionsForm.debug.checked = options.debug;

// or even better, in case it's undefined
optionsForm.debug.checked = Boolean(options.debug);
```

site/en/docs/extensions/reference/storage/index.md
samthor: I slightly disagree. I think a clearer way to write this is:

```js
  if (area === 'sync' && changes.options?.newValue) {
    const debugMode = Boolean(changes.options.newValue.debug);
    console.log('blah', debugMode);
    setDebugMode(debugMode);
  }
```

The condition in the `if` will e.g., make VSCode/tsc happy to allow the access inside the block.

site/en/docs/extensions/reference/storage/index.md
```diff
@@ -81,50 +82,115 @@ tubes can be filled, and if they are filled when you put your message in, it get
...
-[1]: https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage
-[2]: /docs/extensions/mv2/manifest/incognito
-[3]: /docs/extensions/mv2/manifest/storage
-[4]: /docs/extensions/mv2/tabs
-[5]: #property-sync
-[6]: #property-local
+Since service workers are not always running, Manifest V3 extensions sometimes need to
+asynchronously load data from storage before they execute their event handlers. To do this, the
+below snippet uses an async `action.onClicked` event handler that waits for the `storageCache`
+global to be populated before executing it's logic.
```
samthor, line 152: grammar nit: it's => its

site/en/docs/extensions/reference/storage/index.md
```diff
@@ -81,50 +82,115 @@ tubes can be filled, and if they are filled when you put your message in, it get
...
+  <label for="debug">
+    <input type="checkbox" name="debug" id="debug">
+    Enable debug mode
+  </label>
+</form>
+
+//// options.js ////
+
+// In-page cache of the user's options
+let options = {};
```
samthor, line 124: this should be `const` - it never changes, you just assign to it.

site/en/docs/extensions/reference/storage/index.md
samthor: It's called out above. For reasons, this API does not yet support the Promise syntax, see [here](https://chromium.googlesource.com/chromium/chromium/+/master/chrome/common/extensions/api/storage.json). (You'd expect to see `returns_async` keys, rather than callbacks.)

site/en/docs/extensions/reference/storage/index.md
samthor: I don't really understand your second comment here @tophf. Service Workers currently do not run in module mode, so there's no way to do top-level work (no top-level `await`). No amount of rearranging dependencies can change that.

As for your first comment: yes, the loader will happen at the same time. But in general this approach is fine because it's not clear if there's going to be additional listeners added.

I'm generally not a huge fan of making event listeners async, as it gives folks a false sense of—they can do async things without worrying about exceptions—but I'm growing a bit more fine with it.

samthor:
This is a good PR @dotproto, stick with it.

site/en/docs/extensions/reference/storage/index.md
tophf: The API supports Promise since Chrome 88:
https://chromiumdash.appspot.com/commit/84869dc2d4381fe85fd71d5f41289de26caeb406

Your link shows an old storage.json so maybe there's a bug in chromium.googlesource.com?

This is the current storage.json:
https://source.chromium.org/chromium/chromium/src/+/master:extensions/common/api/storage.json

samthor:
Thanks for calling this out—I did a naïve Google search and the result
looked about right so I didn't check that. (I actually don't know what
chromium.googlesource.com is meant to be showing then, the last commit on
that repo is in 2012.)

Yes, we should update this.

site/en/docs/extensions/reference/storage/index.md
tophf: > I don't really understand your second comment here @tophf. Service Workers currently do not run in module mode, so there's no way to do top-level work (no top-level await). 

They will soon but my second comment is not related to that.

> No amount of rearranging dependencies can change that.

It's about the multi-process architecture of the API: the earlier the request is sent to the browser process the sooner it will be processed **there**. So, after the API request is sent, the extension process will be busy compiling and running the code of its synchronously executed dependencies/libraries. For example, it may take 100ms or even 1000ms if there's a lot of code so by the time the listener runs, the storage will be ready in the browser process and quickly transferred to the extension process.

site/en/docs/extensions/reference/storage/index.md
tophf: @samthor, nope, your variant won't react to `false` values.

tophf: Er, sorry, I missed the fact that `debug` is not a separate value but a property nested inside `options`...

site/en/docs/extensions/reference/storage/index.md
samthor: Thanks for the explanation, and I see your point. However, for a simple example page—this is getting a bit in-depth.

I'd prefer module service workers (I suspect they'll work fine—they won't complete loading until the async work is done, so nothing will try to answer the incoming event). It seems like Chrome is about to make them stable. The question though is it worth updating to that model ASAP or showing the older approach for longer.

site/en/docs/extensions/reference/storage/index.md
samthor: Thanks for calling this out—I did a naïve Google search and the result looked about right so I didn't check that. (I actually don't know what chromium.googlesource.com is meant to be showing then, the last commit on that repo is in 2012.)

Yes, we should update this.

site/en/docs/extensions/reference/storage/index.md
dotproto: Ack

site/en/docs/extensions/reference/storage/index.md
dotproto: Thanks, both of you. @samthor's suggestion here looks solid, I'll add that now.

site/en/docs/extensions/reference/storage/index.md
dotproto: While preparing this PR (and retesting before I wrote this) I was not able to a promise back from `chrome.runtime.sync.get()`. I'm planning to file a crbug about this a bit later this evening.

dotproto: Issue filed: https://bugs.chromium.org/p/chromium/issues/detail?id=1197387

site/en/docs/extensions/mv2/content_scripts/index.md
dotproto: There are a couple bits that are a bit novel.

* Cell content spans multiple lines. The only way to do this with our current table support is by having each row in a single line. While I like single line paragraphs in markdown diffs, I find tabular data harder to parse in a single line. Here's a relevant discussion in the [markdown-it repo](https://github.com/markdown-it/markdown-it/issues/406). 
* We currently have IDs assigned to rows. While we can't do this with in markdown, `markdown-it-attrs` does allow us to get close (e.g. ``| `matches` {: #matches } | … | … |``). Perhaps that's sufficient?
* These tables currently have a class `simple` set on them, but it doesn't appear to be used. Good news is it seems `markdown-it-attrs` [allow you to style tables](https://github.com/arve0/markdown-it-attrs#:~:text=32.-,Similar%20for%20tables,-%2C%20attributes).

Here's the before and after for one table. To be honest, I don't see either approach as a clear win. I'm happy to go with whatever you prefer, @samthor.

```
<table class="simple">
  <tbody>
    <tr>
      <th>Name</th>
      <th>Type</th>
      <th>Description</th>
    </tr>
    <tr id="matches">
      <td><code>matches</code></td>
      <td>array of strings</td>
      <td><em>Required.</em> Specifies which pages this content script will be injected into. See <a
          href="/docs/extensions/mv2/match_patterns">Match Patterns</a> for more details on the
        syntax of these strings and <a href="#matchAndGlob">Match patterns and globs</a> for
        information on how to exclude URLs.</td>
    </tr>
    <tr id="css">
      <td><code>css</code></td>
      <td>array of strings</td>
      <td><em>Optional.</em> The list of CSS files to be injected into matching pages. These are
        injected in the order they appear in this array, before any DOM is constructed or displayed
        for the page.</td>
    </tr>
    <tr id="js">
      <td><code>js</code></td>
      <td>
        <nobr>array of strings</nobr>
      </td>
      <td><em>Optional.</em> The list of JavaScript files to be injected into matching pages. These
        are injected in the order they appear in this array.</td>
    </tr>
    <tr id="match_about_blank">
      <td><code>match_about_blank</code></td>
      <td>boolean</td>
      <td><em>Optional.</em> Whether the script should inject into an <code>about:blank</code> frame
        where the parent or opener frame matches one of the patterns declared in
        <code>matches</code>. Defaults to false.</td>
    </tr>
  </tbody>
</table>
```

```
| Name | Type | Description |
| --- | --- | --- |
| `matches` {: #matches } | array of strings | _Required._ Specifies which pages this content script will be injected into. See [Match Patterns](/docs/extensions/mv2/match_patterns) for more details on the  syntax of these strings and [Match patterns and globs](#matchAndGlob) for information on how to exclude URLs. |
| `css` {: #css } | <nobr>array of strings</nobr> | _Optional._ The list of CSS files to be injected into matching pages. These are injected in the order they appear in this array, before any DOM is constructed or displayed for the page. |
| `js` {: #js } | <nobr>array of strings</nobr> | _Optional._ The list of JavaScript files to be injected into matching pages. These are injected in the order they appear in this array. |
| `match_about_blank` {: #match_about_blank } | boolean | _Optional._ Whether the script should inject into an `about:blank` frame where the parent or opener frame matches one of the patterns declared in `matches`. Defaults to false. |
```

samthor: Yeah, thanks for the work. I think Markdown is probably better here.

site/en/docs/extensions/reference/storage/index.md
dotproto: The preview didn't look too bad to me, but I'll split them.

site/en/docs/extensions/reference/storage/index.md
dotproto: > While the idea looks promising on paper but it doesn't actually work as the API request is made in the same initial task shared by the onClicked listener so this code doesn't make any difference to just reading the storage inside onClicked.

I believe you are mistaken.

First, there is no guarantee that the storage get call will complete before an event handler is executed. In many cases that may be true for the reasons you described, but it is not always true. In my local testing I observed a number of cases where `storageCache` was empty at the beginning of the `action.onClicked` event handler call but it was populated after awaiting the resolution of `initStorageCache`.

To reproduce this, use the below code and allow ~30 second after the service worker is inactive before clicking the browser action button. Then, open DevTools for the service worker and observe the logs.

```js
let storageCache = {};
let initStorageCache = (() => {
  return new Promise((resolve) => {
    chrome.storage.sync.get(null, (data) => {
      Object.assign(storageCache, data);
      resolve(storageCache);
    });
  });
})();

chrome.runtime.onInstalled.addListener(() => {
  chrome.storage.sync.set({foo: 'bar'});
});

chrome.action.onClicked.addListener(async (tab) => {
  console.log(initStorageCache, storageCache.foo);
  await initStorageCache;
  console.log(initStorageCache, storageCache.foo);
});
```

Second, even if the `initStorageCache` promise is resolved before the event handler is executed, that doesn't mean "doesn't actually work", just that in that case it is superfluous. The pattern I'm demonstrating here guarantees consistent behavior. We do not know where these samples will end up being used, so we should be resilient against slow disks, large amounts of stored data, etc.

> I'm generally not a huge fan of making event listeners async, as it gives folks a false sense of—they can do async things without worrying about exceptions—but I'm growing a bit more fine with it.

I'm for it if only because there's no other way to await promise resolution. I'm bullish on the pattern of deferring the execution of an event listener's body until you're sure all work necessary to handle the event has completed.  To your point, though, it may be best to include a try/catch to better set expectations.

```js
chrome.action.onClicked.addListener(async (tab) => {
  try {
    await initStorageCache;
  } catch (e) {
    // handle error that occurred during storage initialization
  }
  // normal action handler logic
});
```

I'll integrate the above block.

> However, for a simple example page—this is getting a bit in-depth.

IMO the discussion is all relevant. I'm not just aiming for these snippets to be useful in isolation, I also want them to demonstrate good development practices and be usable in a variety of existing extensions. Even if we can't detail ever decision, we can still demonstrate resilient patterns.

> The question though is it worth updating to that model ASAP or showing the older approach for longer.

Generally speaking, my approach has been to target the capabilities of Chrome stable and call out where they're not backwards compatible & why. Based on the timing of the CL that landed SW module support, I don't expect extensions will support module SWs until M91 at the earliest.

site/en/docs/extensions/reference/storage/index.md
dotproto: I simplified another section to use the first pattern you suggested here. Also, while preparing the initial PR I tested to make sure `undefined` was handled the same as `false`. That said, I think your second suggestion is cleaner and clearer. I'll update both blocks to use this style.

site/en/docs/extensions/reference/storage/index.md
tophf: I'm not mistaken but it seems like I didn't explain my point well. I didn't say that the storage **call** will complete, but the internal implementation of the storage which runs inside the separate process (the browser process).

To clarify, there are two observations I made. The first one is that the idea of preloading the storage at the beginning of the script has exactly the same performance as reading the storage inside the listener so it's useless. And then I amended this observation by finding a case where such preloading will be useful.

site/en/docs/extensions/reference/storage/index.md
dotproto: In that case I'll also change `storageCache` and `initStorageCache` in the last sample to also use const.

site/en/docs/extensions/reference/storage/index.md
dotproto: > Oh and another benefit of preloading would be the case where multiple API events are registered so each listener can start with await initStorageCache

Yep, that's exactly the use case I had in mind. 

This is actually a simplification of a storage library I've been toying with to provide developers with a write-through cache backed by IndexedDB or the Storage API. The goal is to make it so devs can read/write data on a `storage` object they import and it handles the Storage API calls behind the scenes. I'm aiming for the ergonomics of a simple global object with the resilience of the Storage API.

site/en/docs/extensions/reference/storage/index.md
dotproto: Fixed

robdodson (approved):

dotproto:
At this point I think we've addressed the majority of the substantive issues in this PR, so I'm going to merge it in its current state. Feel free to continue discussions here and please open new issues as appropriate. I'm sure we'll continue to update these and other examples.

PR #598, opened 2021-04-02
[POC only] Api not just ref

awfuchs:
Proof of concept for some structural reorg on the API references:

- The API reference singleton in the nav becomes "extensions API" with API reference as first child
- Additional children for some split-out API intro content
- Split out the API intros in a rough way into these three new how-tos


samthor:
I'm generally happy to move more out of the "preamble" for API references and instead move it to howto guides. But this PR otherwise doesn't look ready to review as it still contains a bunch of lorem ipsum.

awfuchs:
Superseded by #663 and its predecessors. Closing, no further work in this direction.

PR #609, opened 2021-04-06
Service worker migration cleanup

dotproto:
This PR is primarily intended for discussion purposes. It introduces two primary changes:

1. Remove the terminal period from one-sentence comments in JS
2. Add a comment to JS blocks to indicate where the example might live

@awfuchs, I'd like to get your thoughts on these changes and the adoption of these patterns across other documentation pages.

(1) is largely personal preference and reflects my personal style. The broader question this touches on is whether or not code comments in both inline snippets and our examples repo should always conform to standard English grammar.

(2) The initial motivation for this change was seeing an exchange where one developer included a code snippet in their comment, but it wasn't clear from the developer's comment where that snippet would be used. While in the context of this page I think it's clear, in another page (e.g. message passing) it may not be.

### DIFF
--- a/site/en/docs/extensions/mv3/migrating_to_service_workers/index.md
+++ b/site/en/docs/extensions/mv3/migrating_to_service_workers/index.md
@@ -49,6 +49,7 @@ The below snippet shows how an existing extension initializes its browser action
 persistent background page.
 
 ```js
+//// background.js
 chrome.storage.local.get(["badgeText"], ({ badgeText }) => {
   chrome.action.setBadgeText({ text: badgeText });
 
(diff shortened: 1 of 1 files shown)


awfuchs (approved):
I'm in favor of both these changes. An additional point is that we should avoid things like:

```
//// background.js
chrome.action.onClicked.addListener((tab) => {
  chrome.scripting.executeScript({
    target: { tabId: tab.id },
    files: ['foo.js'],
  });
});

////foo.js
function doFoo() {
  let result=document.body.innerText;
  . . .
}
```

And instead put each file's fragments in a separate fence:

```
//// background.js
chrome.action.onClicked.addListener((tab) => {
  chrome.scripting.executeScript({
    target: { tabId: tab.id },
    files: ['saycontent.js'],
  });
});
```

```
////foo.js
function doFoo() {
  let result=document.body.innerText;
  . . .
}
```

dotproto:
I was hesitant about splitting multi-file examples into several blocks as it ads more vertical height to the page, but I think the clarity it provides makes this a clear win. I'll use this approach moving forward.

PR #615, opened 2021-04-07
Updated information about inclusion/exclusion from FLoC calculation

samdutton:
As discussed here: WICG/floc/issues/85.

Also updated here: GoogleChrome/web.dev/pull/5039.

### DIFF
--- a/site/en/blog/floc/index.md
+++ b/site/en/blog/floc/index.md
@@ -110,7 +110,7 @@ For example, a site can opt out of all FLoC cohort calculation by sending the HT
 Permissions-Policy: interest-cohort=()
 ```  
 
-During the FLoC origin trial, for websites that don't opt out, pages will be included in the FLoC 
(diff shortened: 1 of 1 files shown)


site/en/blog/floc/index.md
```diff
@@ -110,7 +110,10 @@ For example, a site can opt out of all FLoC cohort calculation by sending the HT
 Permissions-Policy: interest-cohort=()
 ```  
 
-During the FLoC origin trial, websites that don't opt out will be included in the FLoC calculation if Chrome detects that they are sites which load ads-related resources.   
+During the FLoC origin trial, for websites that don't opt out, pages will be included in the FLoC
```
jpmedley, line 113:
```suggestion
During the FLoC origin trial, pages on websites that don't opt out will be included in the FLoC 
```

jpmedley (changes requested):

jpmedley (approved):

PR #617, opened 2021-04-08
Runtime API examples

dotproto:
This PR introduces 3 examples to the chrome.runtime API reference page

* An example of how to use getURL
* A TCP handshake style exchange between a content script and background page
* Retrieving a published extension's key for local development

### DIFF
--- a/site/en/docs/extensions/reference/runtime/index.md
+++ b/site/en/docs/extensions/reference/runtime/index.md
@@ -4,13 +4,15 @@ api: runtime
 
 ## Examples
 
-### Use [`getURL`][getURL] to add an extension image to a page {: #example-getURL }
+### Use [`getURL`][getURL] to add an extension image to a page {: #example-get-url }
 
-Extensions can access their own resources using relative paths, but in order for the extension's
-assets to appear in a web page, you will need to specify the full URL of the asset (and expose it
-using the [web_accessible_resources][war] field).
+In order for a web page to access an asset hosted on another domain, it must specify the resource's
+full URL (e.g. `<img src-"https://example.com/logo.png">`). The same is true for when a web page
+wants to include assets included in an extension. The two main differences here are that the
+extension's assets must be exposed as [web_accessible_resources][war] and that typically content
+scripts are responsible for injecting extension assets.
 
-This example shows how a [content script][content] could add an image in the extension's package to
+This example shows how a [content script][content] can add an image in the extension's package to
 the page that the content script has been [injected][content-inject] into.
 
 ```js
--- a/site/en/docs/extensions/reference/runtime/index.md
+++ b/site/en/docs/extensions/reference/runtime/index.md
@@ -23,49 +23,43 @@ the page that the content script has been [injected][content-inject] into.
 }
 ```
 
-### Use runtime messaging to communicate between background and content scripts {: #example-background-content-messaging }
+### Getting background data into a content script {: #example-content-msg }
 
-A basic example of content script injected into a page and background script passing messages to
-each other. This exchange is modeled on a [TCP handshake][handshake].
+Extensions often need to load data such as user settings or application state in a page shortly
+after a content script is injected. This example shows a simple pattern where a page's content
+script requests data and the background replies.
 
 ```js
 //// content.js ////
 
-// 1. Send the background a message requesting receipt confirmation.
-console.log('[content] sending content-syn');
-chrome.runtime.sendMessage('content-syn');
-
-chrome.runtime.onMessage.addListener((message, sender) => {
-  if (message === 'bg-syn-ack') {
-    // 3. BG syn-ack received, acknowledged receipt.
-    console.log('[content] received bg-syn-ack');
-    console.log('[content] sending content-ack');
-    chrome.runtime.sendMessage('content-ack');
-  }
+// 1. Send the background a message requesting the user's data
+chrome.runtime.sendMessage('get-user-data', (response) => {
+  // 3. Got an asynchronous response with the data from the background
+  initializeUI(response);
 });
 ```
 
 ```js
 //// background.js ////
 
-chrome.runtime.onMessage.addListener((message, sender) => {
-  if (message.type === 'content-syn') {
-    // 2. Content syn received, acknowledge receipt & request confirmation.
-    console.log('[bg] received content-syn');
-    console.log('[bg] sending bg-syn-ack');
-    chrome.tabs.sendMessage(sender.tab.id, 'bg-syn-ack');
-  }
-  if (message.type === 'content-ack') {
-    // 4. Content ack received. Both ends have verified that message
-    //    passing works as expected.
-    console.log('[bg] received content-ack');
+// Example of a simple user data object
+let user = {
+  username: 'demo-user'
+}
+
+chrome.runtime.onMessage.addListener((message, sender, sendResponse) => {
+  // 2. A page requested user data, respond with a copy of the user object
+  if (message === 'get-user-data') {
+    sendResponse(user);
   }
 });
 ```
 
 ### Getting an extension's key from it's manifest {: #example-key }
 
-Retrieve a production extension's [public key][key-prop] for use in development. Execute this snippet in a DevTools console for the production extension paste the value in your unpacked extension's manifest.json.
+Retrieve a production extension's [public key][key-prop] for use in development. Execute this
+snippet in a DevTools console for the production extension to copy the extension's key to your
+clipboard, then open your unpacked extension's manifest.json and paste the key on a new line.
 
 ```js
 //// DevTools console for a published extension
--- a/site/en/docs/extensions/reference/runtime/index.md
+++ b/site/en/docs/extensions/reference/runtime/index.md
@@ -9,7 +9,7 @@ api: runtime
 In order for a web page to access an asset hosted on another domain, it must specify the resource's
 full URL (e.g. `<img src-"https://example.com/logo.png">`). The same is true for when a web page
 wants to include assets included in an extension. The two main differences here are that the
-extension's assets must be exposed as [web_accessible_resources][war] and that typically content
+extension's assets must be exposed as [web accessible resources][war] and that typically content
 scripts are responsible for injecting extension assets.
 
 This example shows how a [content script][content] can add an image in the extension's package to
--- a/site/en/docs/extensions/reference/runtime/index.md
+++ b/site/en/docs/extensions/reference/runtime/index.md
@@ -63,33 +63,21 @@ chrome.runtime.onMessage.addListener((message, sender, sendResponse) => {
 });
 ```
 
-### Getting an extension's key from it's manifest {: #example-key }
+### Gathering feedback on uninstall {: #example-uninstall-url }
 
-When building an extension, there are occasionally times where you may need your unpacked extension
-to have the same extension ID as your production version. For example, [Google OAuth][oauth]
-identifies extensions by their IDs.
+Many extensions use post-uninstall surveys to understand how the extension could better serve its
+users and improve retention. The below example shows how one can add this functionality to their
+extension.
 
-
-Sometimes during development you may need the development version of your extension to have the same
-extension ID as the production version of your extension.
-
-production extension's [public key][key-prop].
-
-
-
-The following example shows a snippet that retrieves the public key; you can execute this in the
-DevTools console...
-
-
-
-Retrieve a production extension's [public key][key-prop] for use in development. Execute this
-snippet in a DevTools console for the production extension to copy the extension's key to your
-clipboard, then open your unpacked extension's manifest.json and paste the key on a new line.
 
 ```js
-//// DevTools console for a published extension
-copy(`"key": "${chrome.runtime.getManifest().key}"`)
+chrome.runtime.onInstalled.addListener(reason => {
+  if (reason === chrome.runtime.OnInstalledReason.INSTALL) {
+    chrome.runtime.setUninstallURL('https://example.com/extension-survey');
+  }
+});
 ```
+
 
 [content-inject]: https://developer.chrome.com/docs/extensions/mv3/content_scripts/#functionality
 [content]: /docs/extensions/mv3/content_scripts/
--- a/site/en/docs/extensions/reference/runtime/index.md
+++ b/site/en/docs/extensions/reference/runtime/index.md
@@ -27,9 +27,13 @@ the page that the content script has been [injected][content-inject] into.
 
 ### Getting background data into a content script {: #example-content-msg }
 
-Extensions often need to load data such as user settings or application state in a page shortly
-after a content script is injected. This example shows a simple pattern where a page's content
-script requests data and the background replies.
+It's common for an extension's content scripts to need data managed by another part of the
+extension, like the extension's background script. Much like two browser windows opened to the same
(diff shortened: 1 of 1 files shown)


site/en/docs/extensions/reference/runtime/index.md
```diff
@@ -3,3 +3,61 @@ api: runtime
 ---
 
 <!-- Intentionally blank -->
```
samthor, line 5: This is no longer blank :)

site/en/docs/extensions/reference/runtime/index.md
```diff
@@ -3,3 +3,61 @@ api: runtime
...
+
+## Examples
+
+Using a content script to inject an extension asset on a page.
```
samthor, line 9: This needs a better intro, sorry. How is this script started? How do I add it to my extension? This is an example of `getURL` but you should open with that.

site/en/docs/extensions/reference/runtime/index.md
```diff
@@ -3,3 +3,61 @@ api: runtime
...
+
+chrome.runtime.onMessage.addListener((message, sender) => {
+  if (message === 'bg-syn-ack') {
+    // 3. BG syn-ack received, acknowledged receipt.
+    console.log('[content] received bg-syn-ack');
+    console.log('[content] sending content-ack');
+    chrome.runtime.sendMessage('content-ack');
+  }
+});
+
```
samthor, line 37: break into two separate blocks (I think @awfuchs had this comment)

site/en/docs/extensions/reference/runtime/index.md
samthor, line 9: Maybe add a heading for each of the three examples.

```
### Use `getURL` to find an image inside the extension

### Use runtime messaging to communicate between ...

### Retrieve arbitrary data from an extension's manifest
```

samthor (approved):
The demos are fine but need explanations.

site/en/docs/extensions/reference/runtime/index.md
dotproto: Lol, very true

dotproto:
> The demos are fine but need explanations.

Can you expand on this? What do you expect / what would you like to see? The only one that feels a bit sparse to me is "Using a content script to inject an extension asset on a page."

Do you have thoughts on the verbosity of inline comments? For the sake of comparison, the Storage API has an example with [a lot of inline comments](https://github.com/GoogleChrome/developer.chrome.com/pull/596/files#diff-4f2912ef8fac7bc6ffdcf6392f2d0cc13fa45a3a62cc910fe04616ef8efe976bR159) and another [with very few](https://github.com/GoogleChrome/developer.chrome.com/pull/596/files#diff-4f2912ef8fac7bc6ffdcf6392f2d0cc13fa45a3a62cc910fe04616ef8efe976bR93).

site/en/docs/extensions/reference/runtime/index.md
dotproto: Done.

site/en/docs/extensions/reference/runtime/index.md
dotproto: No worries. The current copy is terse to a fault. 

I agree that we should express call out `getURL` and better describe the general use case. I'm more hesitant on the other points as I don't think it's appropriate to go into detail about how to use content scripts or how they are executed. That information is better conveyed by more targeted docs. That said, we can and should link to those resources.

Take a look at the updated copy and let me know what you think.

dotproto: TBH I've been on the fence about adding headers to these sections. I think you've pushed me over the edge.

site/en/docs/extensions/reference/runtime/index.md
```diff
@@ -2,4 +2,79 @@
...
+// 1. Send the background a message requesting receipt confirmation.
+console.log('[content] sending content-syn');
+chrome.runtime.sendMessage('content-syn');
+
+chrome.runtime.onMessage.addListener((message, sender) => {
+  if (message === 'bg-syn-ack') {
+    // 3. BG syn-ack received, acknowledged receipt.
+    console.log('[content] received bg-syn-ack');
+    console.log('[content] sending content-ack');
+    chrome.runtime.sendMessage('content-ack');
```
tophf, line 43: This is an anti-pattern. This approach doesn't utilize the one-time messaging mechanism properly: extensions should use sendResponse() provided to the callback as the third parameter as shown in the [message passing](https://developer.chrome.com/docs/extensions/mv3/messaging/#simple) documentation.

Your examples might be suitable for port-based messaging but even then they'll be incorrect in case there are two cycles going on in two tabs. You need to keep a map of tabId-to-messageId - such map is maintained by the built-in one-time messaging automatically.

site/en/docs/extensions/reference/runtime/index.md
```diff
@@ -2,4 +2,79 @@
 api: runtime
 ---
 
-<!-- Intentionally blank -->
+## Examples
+
+### Use [`getURL`][getURL] to add an extension image to a page {: #example-getURL }
```
awfuchs, line 7: I recommend adding text between these headings* between these titles, such as:

"The following sections demonstrate a number of ways to use the chrome.runtime API."

* https://developers.google.com/style/headings#formatting-a-heading-or-title

site/en/docs/extensions/reference/runtime/index.md
```diff
@@ -2,4 +2,79 @@
...
+
+Extensions can access their own resources using relative paths, but in order for the extension's
```
awfuchs, line 9: This is kind of a run-on sentence that rushes past some assumed knowledge. Consider something like:

Extensions can access their own resources using relative paths (like "images/logo.png") but web pages can't, because they are running in a separate namespace. To let web pages access these resources, you need to specify...

site/en/docs/extensions/reference/runtime/index.md
```diff
@@ -2,4 +2,79 @@
...
+{ // Block used to avoid setting global variables
+  let img = document.createElement('img');
+  img.src = chrome.runtime.getURL('logo.png');
+  document.body.append(img);
+}
+```
+
+### Use runtime messaging to communicate between background and content scripts {: #example-background-content-messaging }
+
+A basic example of content script injected into a page and background script passing messages to
```
awfuchs, line 28: Consider:

A common pattern is for a background script to inject a content script into a page. In this case, the content and background scripts are running in separate contexts, and cannot invoke each other directly. Instead, the scripts can communicate by [passing messages](/docs/extensions/mv3/messaging/ but in footnote). The following example shows how...

site/en/docs/extensions/reference/runtime/index.md
```diff
@@ -2,4 +2,79 @@
...
+    // 4. Content ack received. Both ends have verified that message
+    //    passing works as expected.
+    console.log('[bg] received content-ack');
+  }
+});
+```
+
+### Getting an extension's key from it's manifest {: #example-key }
+
+Retrieve a production extension's [public key][key-prop] for use in development. Execute this snippet in a DevTools console for the production extension paste the value in your unpacked extension's manifest.json.
```
awfuchs, line 68: Consider:

Sometimes in development you may need to use the production extension's [public key][key-prop]. The following example shows a snippet that retrieves the public key; you can execute this in the DevTools console...

BUT -- is "execute this in the DevTools console" the kind of example we want?

site/en/docs/extensions/reference/runtime/index.md
dotproto: Forked this discussion into issue #647 as I don't think this is a blocker for merging this PR.

tophf: Well, this code is typical for people who started learning the extensions API, and that's definitely an antipattern, and it will introduce bugs in certain cases so I don't see any reason why you'd want to add it...

dotproto: After some additional reflection I think I now understand your concerns and agree that we shouldn't merge this PR with the current example. I was too focusing on emulating an arbitrary request flow rather than real world scenarios where different extensions contexts need to pass messages. I'll try to get an update in for this tonight.

tophf: Yay!

BTW, for your backlog, there's another problematic example - the first one in the [message passing](https://developer.chrome.com/docs/extensions/mv3/messaging/#simple) article I've linked. It uses chrome.tabs.query() to find the active tab. The problem is that people copypaste it in the wrong places such as tabs.onUpdated or webNavigation listeners instead of using the event's tab or tabId parameter so their code works only if the activity happens in the currently focused tab. I guess this situation can be rectified by adding an example for a listener with a short explanation.

dotproto: This example has been replaced with a simpler content script message and background script `sendResponse()` call.

site/en/docs/extensions/reference/runtime/index.md
dotproto: Took a shot at rewriting this intro. Let me what you think.

site/en/docs/extensions/reference/runtime/index.md
dotproto: Ack. Rewriting the intro.

site/en/docs/extensions/reference/runtime/index.md
```diff
@@ -2,4 +2,82 @@
...
+### Getting background data into a content script {: #example-content-msg }
+
+It's common for an extension's content scripts to need data managed by another part of the
+extension, like the extension's background script. Much like two browser windows opened to the same
+web page, these two contexts cannot directly access each other's values. Instead, the extension can
+use [message passing][message-passing] to coordinate across these different contexts.
+
+In this example, the content script needs some data from the extension's background script in order
+to initialize it's UI. To get this data, it passes a `get-user-data` message to the background, and
+the background responds with a copy of the user's information.
```
dotproto, line 37: @awfuchs, I rewrote the intro to this section. Let me know what you think.

site/en/docs/extensions/reference/runtime/index.md
```diff
@@ -2,4 +2,82 @@
...
+### Use [`getURL`][get-url] to add an extension image to a page {: #example-get-url }
+
+In order for a web page to access an asset hosted on another domain, it must specify the resource's
+full URL (e.g. `<img src-"https://example.com/logo.png">`). The same is true for when a web page
+wants to include assets included in an extension. The two main differences here are that the
+extension's assets must be exposed as [web_accessible_resources][war] and that typically content
+scripts are responsible for injecting extension assets.
+
+This example shows how a [content script][content] can add an image in the extension's package to
+the page that the content script has been [injected][content-inject] into.
```
dotproto, line 16: @awfuchs, I rewrote the intro to this section. Let me know what you think.

site/en/docs/extensions/reference/runtime/index.md
dotproto: That's a fair question. I've found this to be very useful, but I need to do more key manipulation than the average extension developer. I'm inclined to keep it for the moment, but I'm open to dropping it if we haver other examples that would make better use of the page space.

dotproto: On second thought, I'm not wild about encouraging people to copy-paste stuff into their console. I'm going to replace this example with something else.

site/en/docs/extensions/reference/runtime/index.md
```diff
@@ -2,4 +2,88 @@
...
 
-<!-- Intentionally blank -->
+## Examples
+
+### Use [`getURL`][get-url] to add an extension image to a page {: #example-get-url }
+
+In order for a web page to access an asset hosted on another domain, it must specify the resource's
+full URL (e.g. `<img src-"https://example.com/logo.png">`). The same is true for when a web page
+wants to include assets included in an extension. The two main differences here are that the
+extension's assets must be exposed as [web_accessible_resources][war] and that typically content
```
samthor, line 12: Are the underscores important here?

site/en/docs/extensions/reference/runtime/index.md
```diff
@@ -2,4 +2,88 @@
...
+scripts are responsible for injecting extension assets.
+
+This example shows how a [content script][content] can add an image in the extension's package to
+the page that the content script has been [injected][content-inject] into.
+
+```js
+//// content.js
+
+{ // Block used to avoid setting global variables
+  let img = document.createElement('img');
```
samthor, line 22: nit: `const` over `let`

site/en/docs/extensions/reference/runtime/index.md
```diff
@@ -2,4 +2,88 @@
...
+extension, like the extension's background script. Much like two browser windows opened to the same
+web page, these two contexts cannot directly access each other's values. Instead, the extension can
+use [message passing][message-passing] to coordinate across these different contexts.
+
+In this example, the content script needs some data from the extension's background script in order
+to initialize it's UI. To get this data, it passes a `get-user-data` message to the background, and
+the background responds with a copy of the user's information.
+
+```js
+//// content.js ////
```
samthor, line 40: super nit: you _end_ the filename here with ////'s, but you don't do the same above.

site/en/docs/extensions/reference/runtime/index.md
```diff
@@ -2,4 +2,88 @@
...
+  console.log('received user data', response);
+  initializeUI(response);
+});
+```
+
+```js
+//// background.js ////
+
+// Example of a simple user data object
+let user = {
```
samthor, line 54: nit: should be

```js
const user = {
  username: 'demo-user',
};
```

samthor (approved):

site/en/docs/extensions/reference/runtime/index.md
```diff
@@ -2,4 +2,82 @@
...
+
+### Getting background data into a content script {: #example-content-msg }
+
+It's common for an extension's content scripts to need data managed by another part of the
+extension, like the extension's background script. Much like two browser windows opened to the same
+web page, these two contexts cannot directly access each other's values. Instead, the extension can
+use [message passing][message-passing] to coordinate across these different contexts.
+
+In this example, the content script needs some data from the extension's background script in order
+to initialize it's UI. To get this data, it passes a `get-user-data` message to the background, and
```
awfuchs, line 36: it's -> its

awfuchs (approved):
Approved but one grammar comment

site/en/docs/extensions/reference/runtime/index.md
dotproto: @awfuchs indicated offline that he has some nits, but will address them in another PR he's planning to work on for this page.

site/en/docs/extensions/reference/runtime/index.md
dotproto: Done. I'll follow up with you over chat to discuss const/let conventions more generally.

site/en/docs/extensions/reference/runtime/index.md
dotproto: The convention I've been following is to use both. Fixed.

site/en/docs/extensions/reference/runtime/index.md
dotproto: Fixed.

site/en/docs/extensions/reference/runtime/index.md
dotproto: Ugh, I'm bad at that one. I'm just so used to typing 's for possessives 😞

PR #623, opened 2021-04-08
Styling `<code>`

chybie:
Example: Refer to `/blog/migrating-to-typescript/#making-any-call` for the updated styles.

### DIFF
--- a/site/_scss/_typography.scss
+++ b/site/_scss/_typography.scss
@@ -125,9 +125,9 @@
 
   :not(pre)>code {
     background: var(--color-code-bg);
-    margin: 0 0.25em;
     border: 1px solid var(--color-hairline);
     border-radius: 3px;
+    margin: 0 0.25em;
     padding: 0.125em 0.25em;
   }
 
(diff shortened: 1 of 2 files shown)


TimvdLippe (approved):
Styling looks good!

robdodson (approved):

TimvdLippe:
When will this PR be released? It seems like the styling on https://developer.chrome.com/blog/migrating-to-typescript/ is still using the old styles.

chybie:
Oh, it's merged. @robdodson @samthor , did a quick check:
- It's also not caching issue. 
- It works in dev but after production build, the styles were purged somehow, it's not included in the .html file. 

Any ideas of where might this goes wrong before I take a closer look into this?

robdodson:
It's possible that the clever style that we're using is not being properly detected by purgeCss's regex
https://github.com/GoogleChrome/developer.chrome.com/blob/main/site/_transforms/purify-css.js#L61

so it gets purged when we do a production build.

We may need to try rewriting the CSS so it's just:

```scss
code {
  ...
}

pre > code {
  // undo the background color and border radius on code
}
```

chybie:
Created a new PR #633

robdodson:
that seems to have fixed it. Thank you @chybie !

PR #632, opened 2021-04-13
Fix devtools pullapprove label.

robdodson:
Previously the `reviewer: devtools` label wasn't working for blog posts because they weren't part of the AND condition. This fixes that. cc @chybie

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -56,7 +56,8 @@ groups:
       - >
         files.include('site/en/docs/devtools/*') or
         files.include('site/_data/docs/devtools/*') or
-        files.include('site/_data/i18n/docs/devtools.yml')
+        files.include('site/_data/i18n/docs/devtools.yml') or
+        files.include('site/en/blog/*').exclude('site/en/blog/blog.11tydata.js')
 
   # Content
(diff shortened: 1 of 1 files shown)


.pullapprove.yml
```diff
@@ -56,7 +56,8 @@ groups:
       - >
         files.include('site/en/docs/devtools/*') or
         files.include('site/_data/docs/devtools/*') or
-        files.include('site/_data/i18n/docs/devtools.yml')
+        files.include('site/_data/i18n/docs/devtools.yml') or
+        files.include('site/en/blog/*').exclude('site/en/blog/blog.11tydata.js')
```
samthor, line 60: can this include "site/en/blog/*/**.md"? Or do we want it for every file?

samthor (approved):

.pullapprove.yml
robdodson: Good question, I was leaving it as every file for now in case folks (for whatever reason) ever need to link to something like a .zip or a .txt. But yeah, we could probably change them all to *.md as well, I think we'd want to update everything in the file though.

samthor: Sure, this is fine then.

PR #640, opened 2021-04-14
updates types to pull from Chrome directly/ show Promise

samthor:
This updates our types system to depend on `chrome-types-helpers` (see [chrome-types](https://github.com/GoogleChrome/chrome-types)).

This generates Chrome's types from head as part of its build, which is important to the Chrome team—any changes they make to the underlying types should show up immediately.

Version data is generated less regularly, as it reflects only released stable data. We pull this from NPM where it is published on change.


site/_filters/md.js
```diff
@@ -0,0 +1,113 @@
+const md = require('markdown-it')();
```
samthor, line 1: This file should be renamed something like "reference-support.js". It renders inline MD but also turns `{@link ...}` into useful things.

MichaelSolati (approved):

PR #684, opened 2021-04-22
Runtime API intro

awfuchs:
Adds standard content to the introduction for the runtime API.

* [Current page](https://developer.chrome.com/docs/extensions/reference/runtime)
* [PR preview](https://deploy-preview-684--developer-chrome-com.netlify.app/docs/extensions/reference/runtime)


samthor (commented):
This is fine but actually `<dl>` can be written differently, e.g., as we do on [the proxy page](https://raw.githubusercontent.com/GoogleChrome/developer.chrome.com/main/site/en/docs/extensions/reference/proxy/index.md). tl;dr:

```md

Accessing extension and platform metadata

: * getBackgroundPage ...
  * foo

```

Basically a single line on its own followed by a line prefixed by `: ` tells our renderer to do a definition list.

Sifat1998 (approved):

site/en/docs/extensions/reference/runtime/index.md
```diff
@@ -1,6 +1,50 @@
...
+
+<dl>
+<dt>Message passing</dt>
+<dd><ul>
+<li>connect runtime.connect(extensionId: string, connectInfo: object): Port</li>
+<li>connectNative runtime.connectNative(application: string): Port</li>
+<li>sendMessage runtime.sendMessage(extensionId?: string, message: any, options: object, 
+responseCallback: function)</li>
+<li>sendNativeMessage runtime.sendNativeMessage(application: string, message: object, 
+responseCallback: function)</li>
```
rdcronin, line 14: Should we link out here to the message passing article?  That seems like it would be a good standard for intros (linking to the more in-depth articles and / or how-tos)

site/en/docs/extensions/reference/runtime/index.md
```diff
@@ -1,6 +1,50 @@
...
+
+<dt>Managing extension execution and lifecycle</dt>
+<dd><ul>
+<li>reload runtime.reload()</li>
+<li>requestUpdateCheck runtime.requestUpdateCheck(callback: function)</li>
+<li>restart runtime.restart()</li>
+<li>restartAfterDelay runtime.restartAfterDelay(seconds: number, callback: function)</li>
+<li>setUninstallURL runtime.setUninstallURL(url: string, callback: function)</li>
+<li>openOptionsPage runtime.openOptionsPage(callback: function)</li>
+</ul></dd>
```
rdcronin, line 33: I think this category is the "fuzziest".  It's not clear to me how all of these are execution and lifecycle related (e.g., openOptionsPage()).

Also worth noting: restart() and restartAfterDelay() refer to restarting *the device*, not the extension - these are only available in kiosk mode on CrOS.  I think they probably warrant a slightly different section.

rdcronin (commented):
Thanks, Andy!

Chatted briefly offline, but dropping a few notes here for posterity:
- I like that this splits up the API into conceptual groups; I think, particularly for things like message passing, this makes it more digestible.
- I think this is probably already on your radar, but formatting the entries here to be more code-style would help with legibility

(A few more notes inline below)

site/en/docs/extensions/reference/runtime/index.md
awfuchs: Done.

site/en/docs/extensions/reference/runtime/index.md
awfuchs: Renamed the category "Managing extension lifecycle and options"; WDYT? Also moved the restart methods to their own category.

awfuchs:
Removing @dotproto in favor of tagging them in as optional.

site/en/docs/extensions/reference/runtime/index.md
```diff
@@ -1,6 +1,59 @@
...
+Message passing
+
+: These methods support message passing between tasks within the extension (such as background
+  scripts) and with the native system on which the browser is running. See [Message
+  Passing][message-passing] for an overview of the subject.
+  
+  * [runtime.connect(extensionId: string, connectInfo: object): Port](/docs/extensions/reference/runtime/#method-connect)
+  * [runtime.connectNative(application: string): Port]()
+  * [runtime.sendMessage(extensionId?: string, message: any, options: object, responseCallback: function)]()
+  * [runtime.sendNativeMessage(application: string, message: object, responseCallback: function)]()
```
rdcronin, line 16: Thinking about this more, I worry that these are just going to get out of date.  Any time we add a new method, change a signature, change how we represent the signatures in the docs, etc, they are invalidated.

What do you think about:
- Keeping the categories
- Having each category reference and link to "such as" methods

Example:

Message passing
: These methods support passing messages between different contexts.  These can be used either for contexts within your own extension (with [runtime.connect()] and [runtime.sendMessage()], or with native applications on the machine leveraging [native messaging](https://developer.chrome.com/docs/apps/nativeMessaging/).  
See [Message  Passing][message-passing] for an overview of the subject.

site/en/docs/extensions/reference/runtime/index.md
```diff
@@ -1,6 +1,59 @@
...
+
+: This utility lets you convert the path of a resource in your extension to a URL that the browser
+  can resolve.
+
+  * [runtime.getURL(path: string): string]()
+
+
+## Manifest
+
+There are no permissions or other manifest requirements for the runtime API.
```
rdcronin, line 54:
```suggestion
There are no permissions or other manifest requirements to use the runtime API.
```

rdcronin (changes requested):
Thanks, Andy!

site/en/docs/extensions/reference/runtime/index.md
awfuchs: I agree with this concern. I did something similar to your suggestion, but a little more boilerplate-like, so that

- instead of a bullet list of methods (with signatures!), it has a run-in list of method names
- it says each time "Methods in this category include"

WDYT?

site/en/docs/extensions/reference/runtime/index.md
```diff
@@ -1,6 +1,56 @@
 ---
 api: runtime
 ---
+The runtime API provides methods to support a number of areas of functionality that your extensions
+can use:
+
+Message passing
+
+: These methods support message passing between tasks within the extension (such as background
+  scripts) and with the native system on which the browser is running. See [Message
```
rdcronin, line 10: To me, this almost reads as if this only supports passing messages between extension bits and the native system, rather than intra-extension tasks *and* the system.  Also, messaging can be used to communicate with other extensions.  The intra-extension messaging is probably most popular.  What about:

```suggestion
: These methods support message passing so that you can communicate with different parts of your extension (such as an extension popup and background scripts), other extensions, or native applications on the user's device. See [Message
```

site/en/docs/extensions/reference/runtime/index.md
```diff
@@ -1,6 +1,56 @@
...
+Helper utilities
+
+: This utility provide utility such as the conversion of internal resource representations to
+  external formats. Methods in this category include 
+  [getURL](/docs/extensions/reference/runtime/#method-getURL).
+
+
+## Manifest
+
+There are no permissions or other manifest requirements to use the runtime API.
```
rdcronin, line 51:
```suggestion
Most methods on the runtime API do not require any permission to use.  However, sendNativeMessage[linky] and connectNative[linky] require the `nativeMessaging` permission to be declared in your manifest.
```

rdcronin (approved):
LGTM!  This looks much better.  Thanks Andy!

PR #711, opened 2021-04-28
Adds defer publish option to CWS publish page

awfuchs:
Preview for this change: https://deploy-preview-711--developer-chrome-com.netlify.app/docs/webstore/publish/#deferred-publishing

b/186426814

### DIFF
--- a/site/en/docs/webstore/publish/index.md
+++ b/site/en/docs/webstore/publish/index.md
@@ -140,7 +140,7 @@ publish** option from the developer console's &#x22EE; menu as shown below:
 {% Img src="image/SHhb2PDKzXTggPGAYpv8JgR81pX2/yoMNFt1ht6qSLXzFyrWj.png", alt="Screenshot showing
 the 'more' menu's defer publish option", width="386", height="284" %}
 
-This lets you pause the rollout of a submitted item if you discover an error after submitting it, or
+This lets you pause the rollout of a submitted item if you discover an error after submitting it or
 if you simply want to change your rollout time.
 
 
--- a/site/en/docs/webstore/publish/index.md
+++ b/site/en/docs/webstore/publish/index.md
@@ -121,10 +121,28 @@ you uncheck the checkbox, your item will **not** be published immediately after 
 complete. Instead, you'll be able to manually publish it at a time of your choosing once the review
 is complete.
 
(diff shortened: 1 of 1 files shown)


site/en/docs/webstore/publish/index.md
```diff
@@ -121,11 +121,29 @@ you uncheck the checkbox, your item will **not** be published immediately after
 complete. Instead, you'll be able to manually publish it at a time of your choosing once the review
 is complete.
 
+{% Aside %}
```
dasli, line 124: Is there a reason we split the instructions between two sections? What if we removed the aside in the "Deferred publishing option" section and just add the instructions there rather than having them in the "Review of submitted items" section?

awfuchs: The idea was that the "deferred publish option" section is happening before you click "submit for review". And going to the pending review to click on "Defer publish" can only happen once the review is pending.

That was my rationale; it also makes "oh, and if you want to change your mind" appear in an aside rather than as instructions inside the how to submit process -- these instructions are pushed out until later in case you need them then.

Thoughts on this?

dasli: Ok, I don't feel strongly so I'm fine with this organization of content as well.

site/en/docs/webstore/publish/index.md
```diff
@@ -121,11 +121,29 @@ you uncheck the checkbox, your item will **not** be published immediately after
...
 After you submit the item for review, it will undergo a review process. The time for this review
 depends on the nature of your item. See the [FAQ on review times][16] for more details.
 
+While an item is pending review, you can still turn off automatic publishing by choosing **Defer
+publish** option from the developer console's &#x22EE; menu as shown below:
+
+{% Img src="image/SHhb2PDKzXTggPGAYpv8JgR81pX2/yoMNFt1ht6qSLXzFyrWj.png", alt="Screenshot showing
+the 'more' menu's defer publish option", width="386", height="284" %}
+
+This lets you pause the rollout of a submitted item if you discover an error after submitting it, or
```
dasli, line 143: Grammar: no comma needed after "submitting it"

dasli (approved):
These changes look good.

PR #738, opened 2021-05-05
Adjust WebNavigation reference with regards to back/forward cache.

dtapuska:
The back forward cache does not fire onDOMContentLoaded event when 
a page is restored from the cache. Explicitly call this out in the documentation.


dotproto (approved):
Thanks for the update!

PR #765, opened 2021-05-12
Update app deprecation blog URL

dotproto:
Updates the app deprecation banner to use the latest blog post regarding the deprecation. While the current URL has more context on the depreciation strategy, the URL in this PR has the material content developers are often looking for: the current deprecation timeline.


awfuchs (approved):

robdodson:
@dotproto do you want to land this one?

PR #772, opened 2021-05-13
Suppress spurious extension in screenshot

awfuchs:
Corrected per b/182050070


dotproto (approved):
The new screenshot looks good. Two quick items to address before we merge this.

1. Please update the alt text for users that rely on assistive technologies
2. This screenshot also needs to be updated in the mv3 version of the page

PR #775, opened 2021-05-13
Expand documentation on webRequest extra_headers

dotproto:
Docs counterpart to https://crbug.com/1208466

The meaningful content changes begin here: https://deploy-preview-775--developer-chrome-com.netlify.app/docs/extensions/reference/webRequest/#:~:text=chrome%2084


awfuchs (approved):

karan173:
We should revert this. 
- We added an exception for ParsedHeaders in M93 so that they can be modified w/o extraHeaders, see https://chromium-review.googlesource.com/c/chromium/src/+/2927311. Even before that CL, I think these headers could be viewed w/o extraHeaders.
- We had already added an exception for CSP in M91 allowing modification w/o extraHeaders, see https://chromiumdash.appspot.com/commits?commit=67b5d1265dc828c84abf32cd4711da3fa052f424&platform=Windows.

dotproto:
Thanks for catching this! I'll revert this right away.

PR #783, opened 2021-05-14
Add detail for how to see if a method supports promises

awfuchs:
Fixes b/183555452

[Using promises](https://deploy-preview-783--developer-chrome-com.netlify.app/docs/extensions/mv3/promises/) page on the preview site.

### DIFF
--- a/site/en/docs/extensions/mv3/promises/index.md
+++ b/site/en/docs/extensions/mv3/promises/index.md
@@ -55,12 +55,14 @@ Not all methods in extensions APIs support promises. Sometimes that's because we
 promise support on the method yet; in many cases it's because using a promise isn't feasible for the
 method.
 
-{% if false %}
 You can check if an API method support promises by checking its API reference page:
 
-{% Img src="image/SHhb2PDKzXTggPGAYpv8JgR81pX2/lDjGweNENKyqGSA2nZrp.png", alt="Screenshot of an API
-reference page with promises support noted", width="800", height="467" %}
-{% endif %}
+{% Img src="image/SHhb2PDKzXTggPGAYpv8JgR81pX2/3gwYoRgPRWOrs5N2j5mP.png", alt="Screenshot showing a
+method that supports promises", width="800", height="280" %}
+
+The screenshot above shows a method from the `chrome.tabs` API. You can see that this method
+supports promises because it displays the `Promise` pill, and also shows syntax for both the promise
+and callback forms.
 
 ## How to use promises
 
--- a/site/en/docs/extensions/mv3/promises/index.md
+++ b/site/en/docs/extensions/mv3/promises/index.md
@@ -55,14 +55,14 @@ Not all methods in extensions APIs support promises. Sometimes that's because we
 promise support on the method yet; in many cases it's because using a promise isn't feasible for the
 method.
 
-You can check if an API method support promises by checking its API reference page:
+You see if an API method support promises by checking its API reference page:
  
 {% Img src="image/SHhb2PDKzXTggPGAYpv8JgR81pX2/AYQVtnh19vNMHoXzxZB1.png", alt="Screenshot showing a
 method that supports promises", width="800", height="280" %}
 
-The screenshot above shows a method from the `chrome.tabs` API. You can see that this method
-supports promises because it displays the `Promise` pill, and also shows syntax for both the promise
-and callback forms.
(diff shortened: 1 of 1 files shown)


dotproto (changes requested):
Approving to

site/en/docs/extensions/mv3/promises/index.md
```diff
@@ -55,12 +55,14 @@ Not all methods in extensions APIs support promises. Sometimes that's because we
...
 method.
 
-{% if false %}
 You can check if an API method support promises by checking its API reference page:
 
-{% Img src="image/SHhb2PDKzXTggPGAYpv8JgR81pX2/lDjGweNENKyqGSA2nZrp.png", alt="Screenshot of an API
-reference page with promises support noted", width="800", height="467" %}
-{% endif %}
+{% Img src="image/SHhb2PDKzXTggPGAYpv8JgR81pX2/3gwYoRgPRWOrs5N2j5mP.png", alt="Screenshot showing a
+method that supports promises", width="800", height="280" %}
```
dotproto, line 61: At the moment it's a bit hard to see what's going on in this page when viewed on desktop. On first glance it's hard to tell if the image is a screenshot of another page, or if it's an embedded block fo text that has slightly different rendering. At mobile sizes it's more obvious because the image scaling makes it more obvious that it's not text, but it still doesn't read to me as a screenshot.

I think we could address this by adding some UI treatment to the screenshot. A low effort solution would be to add, say, a 1px solid black outline to the image to more clearly offset it from the page's text content. A higher effort approach would be to capture a screenshot of a Chrome window that's sized to show off the specific content we want to highlight on an API doc page.

I just realized that even after looking at this image for a while and writing up this feedback, I still thought the "Methods" header in the screenshot was part of the page's text content. 🤦

awfuchs: Done.

site/en/docs/extensions/mv3/promises/index.md
```diff
@@ -55,12 +55,14 @@ Not all methods in extensions APIs support promises. Sometimes that's because we
 promise support on the method yet; in many cases it's because using a promise isn't feasible for the
 method.
 
-{% if false %}
 You can check if an API method support promises by checking its API reference page:
```
dotproto, line 58:
```suggestion
You see if an API method support promises by checking its API reference page:
```

Slight rewording to remove the double-use of "check" in this sentence.

site/en/docs/extensions/mv3/promises/index.md
```diff
@@ -55,12 +55,14 @@ Not all methods in extensions APIs support promises. Sometimes that's because we
...
+ 
+{% Img src="image/SHhb2PDKzXTggPGAYpv8JgR81pX2/AYQVtnh19vNMHoXzxZB1.png", alt="Screenshot showing a
+method that supports promises", width="800", height="280" %}
 
-{% Img src="image/SHhb2PDKzXTggPGAYpv8JgR81pX2/lDjGweNENKyqGSA2nZrp.png", alt="Screenshot of an API
-reference page with promises support noted", width="800", height="467" %}
-{% endif %}
+The screenshot above shows a method from the `chrome.tabs` API. You can see that this method
+supports promises because it displays the `Promise` pill, and also shows syntax for both the promise
+and callback forms.
```
dotproto, line 65:
```suggestion
The screenshot above shows a method from the [`chrome.tabs`][api-tabs-methods] API. You can see that
this method supports promises because one of the method's signatures returns a promise. To make this
easier to see at a glance, the reference docs also display a `Promise` pill below the signatures.
```

I'm not entirely happy with this suggestion. I like that the current text is that it highlights the pill as the primary signal, as I expect that will be the easiest way for people to see promise support when skimming the docs. On the other hand, I don't think the average reader will be familiar with the "pill", so I wanted to highlight that it appears below the method signatures. Unfortunately, adding that ended up making the paragraph feel repetitive. At the moment, this suggestion is my best effort to blend those concerns. 

Also, I figured that since we're calling out tabs API methods, it made sense to link to them as well. This will let users jump straight to the docs to see non-screenshot examples if they wish. (I can't suggest to the end of the document because nothing changed downt here.)

```
[api-tabs-methods]: /docs/extensions/reference/tabs#method-captureVisibleTab
```

dotproto (approved):
Two relatively minor edits. Approving now to unblock the merge.

PR #799, opened 2021-05-18
Fixing link in "Offline-First TWA" article

demianrenzulli:
Hi @jpmedley!

I've noticed that in the article we published earlier this week, the link included in the first note is broken:

https://developer.chrome.com/docs/android/trusted-web-activity/offline-first/

The reason is that used the same markdown as in Web.dev, but in developer.chrome.com is different. I just fixed this and now it looks good. I might need your review and approval for this PR.

Thanks in advance.


jpmedley (approved):

jpmedley:
Members of DevRel merge their own PRs. Sometimes I forget and who merges and who doesn't.

PR #800, opened 2021-05-19
Remove redundant empty tags

ajhsu:
Just found that there's an redundant empty tag within the Name field.


ajhsu:
@googlebot I signed it!

awfuchs (approved):

PR #804, opened 2021-05-19
guard against missing tags

robdodson:
If someone adds a tag to a post that isn't supported the site will currently throw in a cryptic way.


robdodson:
Sadly there's no try/catch in nunjucks syntax so there isn't an easy way to tell the user "hey, you used a missing tag."

https://twitter.com/rob_dodson/status/1395095457383075842

MichaelSolati (approved):

PR #828, opened 2021-05-26
Minor tweaks to wording

samdutton:
Also: not urgent, but I think the code samples should probably use `const` where appropriate, rather than `let`.

Fixes #SOME_ISSUE_NUMBER

-
-
-


dotproto (approved):
Thanks for the fix!

PR #830, opened 2021-05-27
Typo fix

Klestofer:


Klestofer:
@googlebot I signed it!

jpmedley:
@robdodson There's clearly a mistake to be fixed here. Since it's not visible to readers, I worry more about breaking the link than leaving the error. 

Would you rather leave this or should he add something to [redirects.yaml](https://github.com/GoogleChrome/developer.chrome.com/blob/main/redirects.yaml).

robdodson:
I just did a quick test and it looks like our server redirects don't support hashes, so if we did added a redirect I don't think it would work.

I think we might be ok just fixing the typo. It's unclear how many folks are linking directly to this header, but if someone clicks on a link it won't 404, it'll still take them to the correct page. My concern with not fixing the typo is that other folks will raise a similar PR in the future.

robdodson (approved):

dotproto:
> I just did a quick test and it looks like our server redirects don't support hashes, so if we did added a redirect I don't think it would work.

As I recall hashes resolution in URLs is handled browser side, so this isn't surprising. We could implement redirects with client side JS, but it may not be worth the hassle.

dotproto (approved):
TBH I've been debating whether or not to fix this typo myself. I really don't like breaking links (URLs are one of the web's superpower, after all), but I get bothered every time I see this typo.

I think that most scenarios where a user follows a link to the CSP section of this document there will be external context that clarifies what content the user should see. If they don't see what they expect, they will either skim the document or search the page for an appropriate keyword. As such, I think it's better to fix the typo and (regrettably) break some existing links.

PR #838, opened 2021-06-01
Suggest using ninja -t commands

pcc:


robdodson:
@petele are you able to ask the Chromium Chronicle authors what they think about this change?

petele:
I've pinged the author, will update if/when I hear anything.

petele:
From the author:

> I took a quick look at the "ninja -t commands" and while I'm sure it could be used for preprocessing it's not clear to me that it would be an improvement. This method is not perfect, but it is relatively straightforward

Thanks for the suggestion, going to leave it as is for now.

PR #861, opened 2021-06-08
Fix typo

bangseongbeom:


dotproto:
I believe the current language is correct. The concept of "web accessible resources" was introduced with Manifest V2. You can still see references to it in the [Manifest V2 migration tutorial](https://developer.chrome.com/docs/extensions/mv2/tut_migration_to_manifest_v2/).

PR #863, opened 2021-06-08
[Extensions][Tabs] Give an example for mutating tabs api calls.

solomonkinard:
Showcase how to update tabs while a drag in progress. Promise error can be used instead of chrome.runtime.lastError.

### DIFF
--- a/site/en/docs/extensions/reference/tabs/index.md
+++ b/site/en/docs/extensions/reference/tabs/index.md
@@ -99,7 +99,11 @@ This example shows how to move a tab while a drag may or may not be in progress.
 
 {% Aside %}
 
-Manifest V3 required due to the use of Promises and chrome.tabs.onActivated(), replacing chrome.tabs.onSelectionChanged(). The use of catch(error) in a Promise context is a way to ensure that an error that otherwise populates chrome.runtime.lastError is not unchecked. chrome.tabs.move is used in this example, but the same waiting pattern can be used for other calls that modify tabs while a drag may be in progress.
+Manifest V3 required due to the use of Promises and chrome.tabs.onActivated(), replacing
+chrome.tabs.onSelectionChanged(). The use of catch(error) in a Promise context is a way to ensure
+that an error that otherwise populates chrome.runtime.lastError is not unchecked. chrome.tabs.move
+is used in this example, but the same waiting pattern can be used for other calls that modify tabs
+while a drag may be in progress.
 
 {% endAside %}
 
--- a/site/en/docs/extensions/reference/tabs/index.md
+++ b/site/en/docs/extensions/reference/tabs/index.md
@@ -93,6 +93,31 @@ function toggleMuteState(tabId) {
 }
 ```
 
+### Move the current tab to the first position when clicked
+
+This example shows how to move a tab while a drag may or may not be in progress.
+
+{% Aside %}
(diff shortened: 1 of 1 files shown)


xanderade248:
__l

site/en/docs/extensions/reference/tabs/index.md
```diff
@@ -93,6 +93,31 @@ function toggleMuteState(tabId) {
 }
 ```
 
+### Move the current tab to the first position when clicked
+
+This example shows how to move a tab while a drag may or may not be in progress.
+
+{% Aside %}
+
+Manifest V3 required due to the use of Promises and chrome.tabs.onActivated(), replacing chrome.tabs.onSelectionChanged(). The use of catch(error) in a Promise context is a way to ensure that an error that otherwise populates chrome.runtime.lastError is not unchecked. chrome.tabs.move is used in this example, but the same waiting pattern can be used for other calls that modify tabs while a drag may be in progress.
```
dotproto, line 102:
```suggestion
Manifest V3 required due to the use of Promises and chrome.tabs.onActivated(), replacing
chrome.tabs.onSelectionChanged(). The use of catch(error) in a Promise context is a way to ensure
that an error that otherwise populates chrome.runtime.lastError is not unchecked. chrome.tabs.move
is used in this example, but the same waiting pattern can be used for other calls that modify tabs
while a drag may be in progress.
```

Wrap at 100 characters.

site/en/docs/extensions/reference/tabs/index.md
```diff
@@ -93,6 +93,31 @@ function toggleMuteState(tabId) {
...
+
+{% endAside %}
+
+```js
+// background.js
```
dotproto, line 107:
```suggestion
//// background.js
```

site/en/docs/extensions/reference/tabs/index.md
```diff
@@ -93,6 +93,31 @@ function toggleMuteState(tabId) {
...
+chrome.tabs.onActivated.addListener(activeInfo => move(activeInfo));
+
+function move(activeInfo) {
+  chrome.tabs.move(activeInfo.tabId, {index: 0})
+    .then(response => console.log("Success."))
+    .catch(error => {
+      if (error == "Error: Tabs cannot be edited right now (user may be dragging a tab).")
+        setTimeout(() => move(activeInfo), 500);
+    });
+}
```
dotproto, line 118:
```suggestion
async function move(activeInfo) {
  try {
    await chrome.tabs.move(activeInfo.tabId, {index: 0});
    console.log('Success.');
  } catch (error) {
    if (error == 'Error: Tabs cannot be edited right now (user may be dragging a tab).') {
      setTimeout(() => move(activeInfo), 50);
    }
  }
}
```

Restructured callback to use async/await. Also reduced the timeout period to make drag handling feel as responsive (or at least closer to) keyboard input.

site/en/docs/extensions/reference/tabs/index.md
solomonkinard: Accepted. 4 slashes doesn't seem useful other than just because others are doing it.

dotproto (approved):
Thanks, @solomonkinard!

solomonkinard:
Thanks.

PR #866, opened 2021-06-09
Fix to address incorrect terminology

dotproto:
The original post incorrectly used the term "declarative" where I meant to use "dynamic".


rdcronin (approved):
LGTM!

rdcronin (approved):

PR #892, opened 2021-06-16
Quick fix for api-tabs-methods reference

awfuchs:
Previous merge missed out the definition of the api-tabs-methods xref.


awfuchs:
Was not based on current main. Starting over.

PR #895, opened 2021-06-17
Add Tabs component

### LINKED ISSUE #90: Add reusable tabs component

devnook:
Fixes #90

### DIFF
--- a/site/en/docs/handbook/components/index.md
+++ b/site/en/docs/handbook/components/index.md
@@ -683,6 +683,7 @@ Attribute `title` becomes the title of the corresponding tab panel.
   <web-tab title="Tab 3 (with code)">
 
     This is Tab 3. It has a code snippet inside.
+
     ```js
     const hello = 'world';
     ```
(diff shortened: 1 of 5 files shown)


site/en/docs/handbook/components/index.md
```diff
@@ -663,6 +663,38 @@ Second Term
...
 
+## Tabs
+
+Use `web-tabs` component to display information in a form of horizontal tabs.
+Attribute `title` becomes the title of the corresponding tab panel.
+
+```html
+<web-tabs>
+  <web-tab title="Tab 1">
+    <p>I'm content of Tab 1</p>
```
robdodson, line 674: Could you show an example that adds a code snippet? I think if you add a newline character it should be possible. Something like this:

````html
<web-tab title="Tab 1">

```js
const hello = 'world';
```

</web-tab>
````

robdodson (approved):
LGTM with one suggestion

PR #897, opened 2021-06-17
Note availibility of 'func' and 'args' parameters

alcor:
Couple notes: In the current builds the 'func' parameter seems to be called 'function'

Also, 'args' is not implemented:

background.js:155 Uncaught (in promise) TypeError: Error in invocation of scripting.executeScript(scripting.ScriptInjection injection, optional function callback): Error at parameter 'injection': Unexpected property: 'args'.
    at copyLink (background.js:155)

Perhaps add a note that these are upcoming APIs?

Fixes #SOME_ISSUE_NUMBER

-
-
-

### DIFF
--- a/site/en/blog/crx-scripting-api/index.md
+++ b/site/en/blog/crx-scripting-api/index.md
@@ -168,7 +168,7 @@ chrome.action.onClicked.addListener(async (tab) => {
 
   chrome.scripting.executeScript({
     target: {tabId: tab.id},
-    func: greetUser,
+    function: greetUser,
     args: [givenName],
   });
 });


alcor:
@googlebot I signed it!

robdodson:
@dotproto does this change look correct to you?

samthor:
Sorry for the delay.

It looks like it is named 'func' in Chrome 92, and 'args' is also supported from 92. The change was made to 'func' [here](https://chromium.googlesource.com/chromium/src/+/1f6eb205f66d2d75b61490a2f5b74fb15c4e2825). The way the API docs are generated, we assume that APIs are deprecated for a long time rather than just removed&mdash;in this case it seems like 'function' was removed in favor of 'func'.

This is even though the API was marked stable [6+ months ago](https://chromium.googlesource.com/chromium/src/+/d931ed0b503904f2e17545ba1ed6b2139a5f6b58) and it shouldn't change.

So I might ping @rdcronin about what is going on here.... but this PR isn't valid, because the current docs are correct with the eventual 92+ release.

rdcronin:
Hi there!  Thanks for filing the bug!

So, two pieces here:

The `func` property was originally `function`, but we changed it in M92 (just reaching stable now) order to be compatible with JS object declaration shorthand.  This was tracked in crbug.com/1166438.  For now, both parameters are supported - passing `function` is still allowed, and will be for another couple milestones (to allow developers to migrate to `func` once 92 reaches stable).  Unfortunately, the docserver doesn't provide a good way to indicate this in the generated documentation - instead, `func` now simply says "92+".  This was supposed to be announced in https://developer.chrome.com/docs/extensions/whatsnew/, but it looks like we never did; apologies for that!

The `args` property was also added in M92.  This is reflected in the docs accurately (or appears to be?), where it indicates "Chrome 92+".  92 is rolling out to stable now, so you should be able to use it on stable channel soon.  And, of course, you can experiment with it on Canary, Dev, or Beta channels.

Hope this clears it up, and sorry for the confusion!

PR #901, opened 2021-06-18
Add share button

### LINKED ISSUE #88: Docs and blog posts need share button in the title bar
@robdodson commented on [Thu Nov 05 2020](https://github.com/GoogleChrome/dcc-private/issues/230)

image

devnook:
Fixes #88


site/_includes/layouts/doc-post.njk
```diff
@@ -1,7 +1,7 @@
 {% extends "layouts/base.njk" %}
 
 {% block title_bar %}
-  {% include 'partials/breadcrumbs.njk' %}
+  {% include 'partials/breadcrumbs-post.njk' %}
```
robdodson, line 4: Initially I wasn't so sure about naming this "-post" because then it seems like it is just for posts, but we use it for docs as well.

We could keep the same breadcrumb include and make the share button configurable.

```njk
{% set sharingEnabled = true %}
{% include 'partials/breadcrumbs.njk' with context %}
```

By adding `with context` we give the partial access to variables in the parent scope.

Alternatively we could convert the breadcrumbs partial into a macro and pass it as an argument.

```njk
{% set options = { sharingEnabled: true } %}
{{ breadcrumbs(options) }}
```

Note that nunjucks has some issues with using set and object sub properties (https://github.com/mozilla/nunjucks/issues/313)

robdodson:
There's a small issue where the anchor does not take up the full size of the share button, so you can hover and click it but nothing happens.

Kapture 2021-06-25 at 11 14 46

site/_includes/layouts/doc-post.njk
devnook: I think "with context" works only with import, not include. with include the variable should just be available inside the template as-is.

devnook:
PTAL

devnook:
PTAL

site/_scss/blocks/_share-button.scss
```diff
@@ -0,0 +1,25 @@
...
+
+  &:hover {
+    background: var(--color-bg-shade);
+  }
+
+  .share-button__link {
+    display: inline-block;
+    height: 32px;
+    padding-left: 5px;
+    width: 28px;
```
samthor, line 15: I'm confused, this still has the issue Rob mentioned, I can hover the circle without it being clickable. Can you just make this `width: 100%` without padding and set `text-align: center` or use Flex or something?

devnook: Please try now

samthor:
The SVG still feels slightly off, like half a pixel or something. But that could just be me, and the hover area issue is fixed, so LGTM

samthor (approved):

PR #908, opened 2021-06-21
Cleanup best practices page

dotproto:
* Fix incorrect grammar
  * old: "listing should sets the user's expectations and clearly communicates what the extension does"
  * new: "should set the user's expectations and clearly communicate what the extension does"
* Rewrap at col 100
* Replace numeric links with named ones
* Alphabetically sort named links


awfuchs (approved):

codedfiles (commented):
Duplicate of #

PR #923, opened 2021-06-25
improve displaying of search results on small screens

### LINKED ISSUE #917: Search results are overflowing.
If you use mobile emulation on devtools and search for something, the results will overflow and look cropped. This means the preview images are hidden.

image

devnook:
Fixes #917

### DIFF
--- a/site/_scss/blocks/_search-box.scss
+++ b/site/_scss/blocks/_search-box.scss
@@ -111,7 +111,7 @@ search-box[active] {
   margin: get-size(300) auto 0 auto;
 
   @include media-query('md') {
-    display: initial;
+    display: block;
     margin: 0 get-size(300) 0 get-size(300);
   }
 }
--- a/site/_scss/blocks/_search-box.scss
+++ b/site/_scss/blocks/_search-box.scss
@@ -112,6 +112,6 @@ search-box[active] {
 
   @include media-query('md') {
     display: initial;
-    margin: 0 0 0 get-size(300);
+    margin: 0 get-size(300) 0 get-size(300);
   }
 }
--- a/site/_scss/blocks/_search-box.scss
+++ b/site/_scss/blocks/_search-box.scss
@@ -89,12 +89,8 @@ search-box[active] {
   --overlay-background-color: var(--color-primary);
   @extend %overlay;
   align-items: center;
+  display: inline-flex;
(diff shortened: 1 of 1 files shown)


robdodson:
I think there might be a couple of issues.

If you type "qq" into the search box, it should show you the what's new in chrome post. On desktop, the thumbnail is pushed offscreen to the right, but it should be next to the text.

On mobile, the thumbnail is centered underneath the text, but it should either be to the right of the text, or, if there's not enough room, we should consider hiding the thumbnail on mobile.

devnook:
PTAL

site/_scss/blocks/_search-box.scss
```diff
@@ -103,12 +103,15 @@ search-box[active] {
 
 .search-box__snippet {
   color: var(--color-secondary-text);
+  word-break: break-all;
```
robdodson, line 106: You can remove this. Otherwise it causes words to wrap in strange ways like breaking at the last letter. The words seem to wrap correctly now with the margin that you added.

site/_scss/blocks/_search-box.scss
```diff
@@ -103,12 +103,15 @@ search-box[active] {
...
 }
 
 .search-box__thumbnail {
-  margin-left: get-size(200);
+  display: none;
+  margin: get-size(300) auto 0 auto;
```
robdodson, line 111: should the margin be 0 if you're hiding the thumbnail?

site/_scss/blocks/_search-box.scss
```diff
@@ -103,12 +103,15 @@ search-box[active] {
...
 
   @include media-query('md') {
-    margin-left: get-size(300);
+    display: initial;
```
robdodson, line 114:
```suggestion
    display: block;
```

site/_scss/blocks/_search-box.scss
```diff
@@ -103,12 +103,15 @@ search-box[active] {
...
+    margin: 0 0 0 get-size(300);
```
robdodson, line 115:
```suggestion
    margin: 0 get-size(300) 0 get-size(300);
```

This gives the image a bit of space from the vertical scrollbar on smaller desktop screens.

samthor:
I fixed this in #1148.

PR #940, opened 2021-06-29
Set card images to auto height to avoid cropping.

### LINKED ISSUE #891: A title is being cut off on the blog landing page
I think this image speaks for itself.

image

robdodson:
Fixes #891 

On web.dev we use object-fit: cover on our cards and that's usually OK since folks rarely include text in their images. But on d.c.c. folks use text in images a lot, because they're doing blog post series.

This changes makes the card image area `height: auto` (the default) so images don't inadvertently get cropped on smaller screens. We do still have a max-height to prevent the images from getting too tall. So in some situations there will be a tiny bit of cropping, but hopefully nothing too bad.

Separately, we should probably update our docs to include ideal aspect ratios for card thumbnails and hero images.

### DIFF
--- a/site/_scss/blocks/_blog-card.scss
+++ b/site/_scss/blocks/_blog-card.scss
@@ -10,6 +10,6 @@
 
 .blog-card__thumbnail {
   img {
-    height: px-to-rem(184px);
+    max-height: px-to-rem(184px);
   }
 }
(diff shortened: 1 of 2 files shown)


MichaelSolati (approved):

dotproto:
I tested this PR on macOS 11.3 in Chrome 91.0.4472.114, Firefox 89.0.2 (64-bit), and Safari Version 14.1 (16611.1.21.161.3) using each browsers' responsive mode developer tools. To test this change, I visited https://deploy-preview-940--developer-chrome-com.netlify.app/ in each browser, scrolled to the "Blog" card and cycled through the available mobile devices. In all case the "Extension actions in Manifest V3" image was clearly visible.

PR #950, opened 2021-07-01
Fix broken link in Tabs API docs

dotproto:


site/en/docs/extensions/reference/tabs/index.md
```diff
@@ -8,8 +8,9 @@ extra_permissions_html:
 
 You can use most `chrome.tabs` methods and events without declaring any permissions in the
 extension's [manifest][manifest] file. However, if you require access to the [`url`][prop-url],
-[`pendingUrl`][3], [`title`][prop-title], or [`favIconUrl`][prop-faviconurl] properties of
-[`tabs.Tab`][tab], you must declare the `"tabs"` permission in the manifest, as shown below:
+[`pendingUrl`][prop-pendingurl], [`title`][prop-title], or [`favIconUrl`][prop-faviconurl]
```
rachelandrew, line 11: these still don't seem to work, looking at the preview https://deploy-preview-950--developer-chrome-com.netlify.app/docs/extensions/reference/tabs and the complete file it look as if there are no IDs in the document of these values, other internal links just link to the main Tab section.

dotproto: Ah, for clarity I wasn't trying to fix the missing anchor issue (I believe that's a bug in the API doc generator, just opened #954). I think our options are to either…

1. Merge it as is to fix the rendering issue on `pendingUrl` ([`pendingUrl`][3] vs. [`pendingUrl`](#property-Tab-pendingUrl))
2. Update this PR to remove the links form these property names since they don't currently work.

What do you think?

rachelandrew: I think remove them if they don't work, we can always add them back in again once they will.

samthor: It took me a while to understand this issue. This is trying to deep-link into a property of a class. We don't support that yet (and the old site was actually broken for this case anyway), I have a plan on this but I suggest just removing the specific links for now.

awfuchs (approved):

PR #958, opened 2021-07-03
Removes Manifest image on Cookies page

factoidforrest:
mentioning the manifest pane and having an image of it was irrelevant and distracting.  We just want to know where to head, and that leads the reader to believe that they should be in the manifest pane before they continue reading.  It's like telling someone "You see a fork in the road, and you will want to head not left".

Fixes #SOME_ISSUE_NUMBER

-
-
-

### DIFF
--- a/site/en/docs/devtools/storage/cookies/index.md
+++ b/site/en/docs/devtools/storage/cookies/index.md
@@ -16,18 +16,11 @@ page's cookies with Chrome DevTools.
 ## Open the Cookies pane {: #open }
 
 1.  [Open Chrome DevTools][2].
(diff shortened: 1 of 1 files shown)


factoidforrest:
@googlebot I signed it!

rachelandrew (changes requested):
I agree with this suggestion, having just worked through the steps in the original article, it really doesn't matter which pane opens up, the reader just needs to know to select storage.

However, the figure numbering will need to be corrected throughout the document, are you able to make that change?

samthor:
@factoidforrest we're happy to accept the PR but can you make the suggested change?

samthor:
Took this over in #1235.

PR #965, opened 2021-07-07
Changed incorrect property func to function

shahednasser:
In the [chrome.commands](https://developer.chrome.com/docs/extensions/reference/commands/#action-command) API reference, the example that uses the method `chrome.scripting.executeScript` uses an incorrect property in the object passed. Instead of `function`, it passes `func`. When this code of the example is used, an error is thrown.

This pull request fixes this issue by changing `func` to `function`

**To Reproduce**
1. Go to [chrome.commands](https://developer.chrome.com/docs/extensions/reference/commands/#action-command) documentation
2. Copy the code for the Action command example for `background.js`
3. Create a chrome extension with a service worker `background.js` and paste the content in it
4. get an error in service worker when command is triggered:

```
Error in event handler: TypeError: Error in invocation of scripting.executeScript(scripting.ScriptInjection injection, optional function callback): Error at parameter 'injection': Unexpected property: 'func'.
    at chrome-extension://enhjmpkahklpmbndljologkmjjinpihf/background.js:2:22
```

### DIFF
--- a/site/en/docs/extensions/reference/commands/index.md
+++ b/site/en/docs/extensions/reference/commands/index.md
@@ -250,7 +250,7 @@ keyboard shortcut.
 chrome.action.onClicked.addListener((tab) => {
   chrome.scripting.executeScript({
     target: {tabId: tab.id},
-    func: contentScriptFunc,
+    function: contentScriptFunc,
     args: ['action'],
   });
 });


shahednasser:
@googlebot I signed it!

dotproto:
This seems like a good change to help developers targeting the current stable version of Chrome.

For context, this is a bit of an odd situation where we changed the property names after the feature was first implemented. When support for functions and arguments first landed, the options object for `scripting.executeScript` took the properties `function` and `arguments`, respectively. Shortly after that, we realized that since "function" is a keyword in JavaScript, it wasn't possible to assign a variable named function, which in turn meant that developers couldn't use object shorthand to assign that property ([issue 1166438](https://bugs.chromium.org/p/chromium/issues/detail?id=1166438)). For example

```js
let function = () => { console.log('done') }; // this line throws
chrome.action.onClicked.addListener((tab) => {
  chrome.scripting.executeScript({
    target: {tabId: tab.id}  
    function
  });
});
```

In order to enable this pattern, we had to rename `function` to something else and we settled on `func`. This change was made in [commit 1f6eb20](https://source.chromium.org/chromium/chromium/src/+/1f6eb205f66d2d75b61490a2f5b74fb15c4e2825), which ships in Chrome 92 ([OmahaProxy](https://storage.googleapis.com/chromium-find-releases-static/1f6.html#1f6eb205f66d2d75b61490a2f5b74fb15c4e2825)). 

Another minor complication is that our generated docs reflect commits to the Chromium project, or more practically speaking what extension developers can see in Canary. Because of this, we (myself and @awfuchs) have tended to use Canary-compatible code samples in extensions documentation. We may need to revisit that discussion, though.

dotproto (approved):

jen801 (commented):
>

jen801 (commented):
#965

jen801 (commented):
#970

PR #979, opened 2021-07-13
Fix invalid xmlns for gupdate tag

dzsibi:
The document uses both the HTTP and the HTTPS version of the URL in the xmlns attribute, but only the HTTP is valid. I replaced the appearances of the HTTPS version with the HTTP version.

Relevant code here:

https://source.chromium.org/chromium/chromium/src/+/main:extensions/browser/updater/safe_manifest_parser.cc;l=186;drc=d2f37a73b35d282c9b8aa553f01faa1bfbf453c8


samthor:
Sorry, we can't accept this PR unless you sign the CLA.

dzsibi:
@googlebot I signed it!

samthor (approved):

PR #988, opened 2021-07-15
Update externally_connectable example to use HTTPS

alesandroortiz:
The `externally_connectable` example in https://developer.chrome.com/docs/extensions/mv3/messaging/#external-webpage allows any protocol including HTTP, which allows MITM attackers to connect to the extension. To promote secure development, this PR changes the example to only allow HTTPS. This reduces risk if developer copy/pastes example and then edits only the hostname (like I did, before realizing the security implications).

See [this comment](#issuecomment-880395786) for other instances of similarly-insecure config examples in the docs.

- Update `externally_connectable` example to use HTTPS

### DIFF
--- a/site/en/docs/extensions/mv3/messaging/index.md
+++ b/site/en/docs/extensions/mv3/messaging/index.md
@@ -208,7 +208,7 @@ which websites you want to communicate with. For example:
 
 ```json
 "externally_connectable": {
-  "matches": ["*://*.example.com/*"]
+  "matches": ["https://*.example.com/*"]
 }
 ```
 


alesandroortiz:
The only other instance of this configuration is in:
* https://github.com/GoogleChrome/developer.chrome.com/blob/main/site/en/docs/extensions/mv3/manifest/index.md

However, given that location is unlikely to be copy/pasted into code, and it's an example of the most flexible allowable syntax, it seems appropriate to leave it there.

There's generous usage of `http` and wildcard protocols in example patterns on some pages, but I'll open a separate issue for that in a day or so.
Identified locations for separate issue:
* https://github.com/GoogleChrome/developer.chrome.com/blob/main/site/en/docs/extensions/mv3/match_patterns/index.md
* https://github.com/GoogleChrome/developer.chrome.com/blob/main/site/en/docs/extensions/mv3/content_scripts/index.md
* https://github.com/GoogleChrome/developer.chrome.com/blob/main/site/en/docs/extensions/mv3/faq/index.md

samthor:
Thanks, this seems sane. Remember, a huge amount of this documentation is 8-9 years old, and it was definitely a different time then.

samthor (approved):

alesandroortiz:
Thanks for review and merge!

PR #992, opened 2021-07-15
Fix match pattern typos in example configs in Security page

alesandroortiz:
Fixes invalid match patterns in Security page.

~~As a related issue, I'm not sure if the first pattern (`/*`) is written as intended. I think it's intended to be `https://developer.chrome.com/*` but it's unclear due to lack of full file history. If someone lets me know in comments what it's intended to be, I can make the fix as part of this PR.~~ Also fixed per [comment](#issuecomment-885186796).

- Fix match pattern typos in example configs in Security page

### DIFF
--- a/site/en/docs/extensions/mv3/security/index.md
+++ b/site/en/docs/extensions/mv3/security/index.md
@@ -49,14 +49,14 @@ specified in the permissions.
   "description": "Example of a Secure Extension",
   "permissions": [
     "https://developer.chrome.com/*",
-    "https://*.google.com/"
+    "https://*.google.com/*"
   ],
   "manifest_version": 2
 }
 ```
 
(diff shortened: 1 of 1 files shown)


samthor:
I think the first pattern is probably wrong. Do you want to fix it?

alesandroortiz:
> I think the first pattern is probably wrong. Do you want to fix it?

Updated the `https://developer.chrome.com/*` match pattern and also fixed the path patterns to include wildcard (end pattern in `/*` instead of `/`) since otherwise it would only match `/` path AFAIK.

samthor:
Thanks!

alesandroortiz:
Thanks for the review and merge. Have a great week!

PR #996, opened 2021-07-16
Add further language support to d.c.c

devnook:
This PR:

1. Adds a LanguageSelect component in footer
2. Sets up the hreflang tags in the head of each document if alternative language versions are available. Right now we have alternative copies for pages:  / and /404/ if you want to test.
3. Adds scaffold for 5 new languages (ja, ko, ru, pt, zn). The language select will work on pages: / and /404/.


site/_scss/blocks/_language-select.scss
```diff
@@ -0,0 +1,56 @@
...
+    overflow: hidden;
+    padding: 0;
+    position: absolute;
+    white-space: nowrap;
+    width: 1px;
+  }
+
+  select {
+    /* stylelint-disable property-no-vendor-prefix */
+    -moz-appearance: none;
```
myakura, line 18: Firefox supports the unprefixed `appearance` property since Firefox 80 (released in August 2020). Unless there's much traffic from Firefox <80, you can remove this line.

site/_scss/blocks/_language-select.scss
```diff
@@ -0,0 +1,56 @@
+.language-select {
+
+  // Temporary class until web.dev gets rid of w- prefix.
+  .w-visually-hidden {
```
robdodson, line 4: d.c.c. has a `visually-hidden` class. I'm wondering if you can write this as:

```
.w-visually-hidden {
  @extend .visually-hidden;
}
```

site/_scss/blocks/_language-select.scss
```diff
@@ -0,0 +1,56 @@
+.language-select {
```
robdodson, line 1: Rather than scope all of these styles to .language-select, i think it might be worth having a `_select.scss`, similar to what web.dev does. I think we generally want all of our select elements to look like this.

robdodson (approved):
I think this generally looks ok to me, with a few comments. Marking it as approved to prevent blocking you since I'll be out Monday.

It might also be good to have @samthor take a quick look now that he's back.

robdodson:
All of the warnings in the output are coming from line 45 of `find.js`. It seems like it's trying to build a bunch of pages that don't exist and maybe that's causing the builds to fail?

site/_scss/blocks/_language-select.scss
devnook: Thanks @myakura !

devnook: Thanks @myakura

devnook:
PTAL

site/_scss/blocks/_select.scss
```diff
@@ -0,0 +1,36 @@
...
+// This will remove the dotted outline inside of the select when it is focused
+// in Firefox.
+// The trick is to make the text transparent (which makes the dotted outline
+// transparent) and then "display" the text again using a 0 offset text-shadow.
+// Note that this forces the text to be black when it is focused in Firefox.
+// To change that color, change the value of the text-shadow.
+select:-moz-focusring {
+  color: transparent;
+  text-shadow: 0 0 0 #000;
+}
```
myakura, line 32: I think you don't have to do this anymore. Firefox recently had a visual refresh which changes the default style of form elements. I've checked with Firefox on macOS and Windows and they don't show dotted outlines inside select box.

devnook: Done!

site/_js/web-components/language-select.js
```diff
@@ -0,0 +1,54 @@
...
+    if (lang !== this.current) {
+      const pathParts = location.pathname.split('/');
+      const replace = this.supportedLanguages.includes(pathParts[1]) ? 1 : 0;
+      pathParts.splice(1, replace, lang);
+      location.href = pathParts.join('/');
+    }
+  }
+}
+
+// @ts-ignore
```
samthor, line 53: Does TS not think `LanguageSelect_` is a HTML element constructor? Maybe add a note, we can fix it later.

devnook: I think it is similar to this one, but I was unable to pinpoint it: https://github.com/microsoft/TypeScript/issues/35951

site/_shortcodes/Hreflang.js
```diff
@@ -0,0 +1,33 @@
...
+ * @param {EleventyCollectionObject} collections Eleventy collections object
+ */
+function Hreflang(url, site, collections) {
+  if (!url) {
+    return;
+  }
+  const urlParts = url.split('/');
+  const replace = site.locales.includes(urlParts[1]) ? 1 : 0;
+  const hreflangs = site.locales.reduce((out, locale) => {
+    urlParts.splice(1, replace, locale);
```
samthor, line 18: I don't think this is right. If the source URL was not localized (and `urlParts[1]` is not a language code), then `replace` will always be zero, so we'll _add_ a new language code for each iteration of this `reduce`.

I kind of forget whether _all_ URLs will have a language code at "/<lang>": if so, then the 1/0 check isn't needed, and your code is odd but will always work correctly.

I think ideally you'd strip the previous language code first, outside the loop, then slice the array every iteration and insert something new into it.

devnook: You're right, the replace bit should be inside the loop. I moved this logic to a separate helper for better readability, and added a test, just to be sure.

PR #1009, opened 2021-07-20
add chromeOS warning as a pill

samthor:
This updates the "Availability" section of APIs to include a notice about Chrome OS-only APIs. This is already a _giant_ warning at the top of the page, but this section should contain all the data you need to understand whether an API is available.

image

### DIFF
--- a/site/_includes/macros/render-type.njk
+++ b/site/_includes/macros/render-type.njk
@@ -506,12 +506,18 @@
     {% elif feature.unknownVersion %}
       {# This is stable but not yet visible in the version history data. It's probably coming next
         Chrome release. #}
-      <span class="aside--warning tag-pill type--label rounded-lg">
+      <span class="aside--warning tag-pill type--label rounded-lg" title="No version data available">
         Unknown version
       </span>
     {% elif feature.availableFromVersion %}
       <span class="aside--default tag-pill type--label rounded-lg">
         Chrome {{ feature.availableFromVersion }}+
+      </span>
+    {% endif %}
+
+    {% if feature.chromeOsOnly %}
+      <span class="aside--key-term tag-pill type--label rounded-lg">
+        Chrome OS only
       </span>
     {% endif %}
 


site/_includes/macros/render-type.njk
```diff
@@ -506,7 +506,7 @@
     {% elif feature.unknownVersion %}
       {# This is stable but not yet visible in the version history data. It's probably coming next
         Chrome release. #}
-      <span class="aside--warning tag-pill type--label rounded-lg">
+      <span class="aside--warning tag-pill type--label rounded-lg" title="No version data available">
```
devnook, line 509: You might want to use i18n filter here to have this string be translatable.

site/_includes/macros/render-type.njk
```diff
@@ -515,6 +515,12 @@
       </span>
     {% endif %}
 
+    {% if feature.chromeOsOnly %}
+      <span class="aside--key-term tag-pill type--label rounded-lg">
+        Chrome OS only
```
devnook, line 520: Same here

devnook:
Do you have example url where this is rendered? Can't find the pill:)

site/_includes/macros/render-type.njk
samthor: This is normally a good suggestion but our reference documentation has no plan to be translated, since the raw source in Chrome is far away and English-only.

Regardless it's weird to just fix it for a single string in a world where there's lots of them :)

samthor:
One example (for the whole API): https://deploy-preview-1009--developer-chrome-com.netlify.app/docs/extensions/reference/enterprise_deviceAttributes

There's also an example of a single method being only supported there, inside an API with wider support: https://deploy-preview-1009--developer-chrome-com.netlify.app/docs/extensions/reference/idle#method-getAutoLockDelay

robdodson (approved):

PR #1033, opened 2021-07-26
makes desktop toc sticky + highlight top link

samthor:
This makes the ToC on big screens sticky—we're wasting a huge amount of whitespace otherwise—and adds a WC that highlights the top visible heading on the page.

~~cc @jimoong for design opinions - we discussed this, but is bolding the link the right choice? DevSite, for instance, has gray ToC links and gives the active link a color and bold weight.~~

_Update: @jimoong is happy from a design point of view._

This is really helpful on reference pages:
https://deploy-preview-1033--developer-chrome-com.netlify.app/docs/extensions/reference/browserAction#method-getBadgeBackgroundColor

But it also works on any page with a TOC (just picked a random one):
https://deploy-preview-1033--developer-chrome-com.netlify.app/docs/multidevice/ios/links/


robdodson:
Seems like it can cause a bit of a weird shift on longer words.

Kapture 2021-07-26 at 15 44 19

Also looks like it's failing some linter checks.

samthor:
@robdodson do a hard refresh, I just pushed a change which brings `letter-spacing` down a touch to try to prevent this due to bolding.

In general this matches DevSite which also uses bolding.

site/_scss/blocks/_toc.scss
```diff
@@ -30,6 +30,35 @@
...
+  top: get-size(200);
+
+  &::-webkit-scrollbar {
+    height: 8px;
+    width: 8px;
+  }
+  &::scrollbar {
+    height: 8px;
+    width: 8px;
+  }
```
robdodson, line 48: Kind of wondering if we should reuse the scrollbar mixin and style from web.dev

https://github.com/GoogleChrome/web.dev/blob/main/src/styles/utilities/_scrollbars.scss
https://github.com/GoogleChrome/web.dev/blob/main/src/styles/tools/_mixins.scss#L65

site/_scss/blocks/_toc.scss
```diff
@@ -30,6 +30,35 @@
...
+
+  &::-webkit-scrollbar-thumb {
+    background: rgba(128,134,139,.26);
```
robdodson, line 51: on web.dev we use $GREY_500 for scrollbars. On d.c.c. we tend to use custom properties for colors so we can (eventually) support a dark mode theme. It would be good to add this color to our theme and use the custom property in case we want to change our custom scrollbar colors. https://github.com/GoogleChrome/developer.chrome.com/blob/main/site/_scss/_theme.scss

site/_includes/partials/toc-side.njk
```diff
@@ -7,6 +7,8 @@
   still have the right margin. Otherwise the article will shift out of place.
 #}
 <div class="toc-container display-none xl:display-block pad-left-400 pad-right-400 {% if not tocContents %}visibility-hidden{% endif %}">
-  <div class="type--h6 color-secondary-text">{{ 'i18n.common.toc' | i18n(locale) }}</div>
-  <div class="toc__wrapper">{{ tocContents | safe }}</div>
+  <toc-active>
```
robdodson, line 10: hm... Ewa wrote a similar component for the Learn CSS course (web-scroll-spy)
https://github.com/GoogleChrome/web.dev/blob/main/src/lib/components/ScrollSpy/index.js

It seems ungood to have both

samthor: Well, there you go.

I could merge them and move it to webdev-infra. I'd probably rather get this in then move them though.

jimoong:
Can we use something like "overflow-wrap: normal;" to prevent word wrapping rather then touching letter spacing? I guess it's better to keep the word together even if it overflows the box for legibility, but not sure if it makes sense in other edge cases.

site/_includes/partials/toc-side.njk
robdodson: Yeah ideally we should have one component that does it. I'm fine not blocking this PR if you plan to do a follow up that merges and moves them to infra.

site/_scss/blocks/_toc.scss
samthor: I've mirrored them for now.

samthor:
@jimoong, rather than that, I've just extended the ToC to take up more of the available space. PTAL at the original link!

We had a huge amount of margin on the right, so instead I've allowed the links to extend into that area. I think this will cover the vast majority of long method names, and I think if they're absolutely huge we should wrap them rather than have them disappear off the screen.

samthor:
The linters are now passing here.

robdodson (approved):

PR #1040, opened 2021-07-27
Add sitemap.xml

robdodson:
We received an internal request to add a sitemap.xml so Googlers can index their docs internally.

- Adds a `formatDateNumeric` function to the date helpers.
- Adds a sitemap.xml.njk which is pretty similar to the one used on web.dev, but doesn't need the luxon dependency for dates.

### DIFF
--- a/site/_data/lib/date.js
+++ b/site/_data/lib/date.js
@@ -60,6 +60,7 @@ const formatDateShort = (date, locale = 'en', opts = {}) => {
 /**
  * Returns a date string formatted like "2021-12-31".
  * @param {Date} date
+ * @returns {string}
  */
 const formatDateNumeric = date => {
   return date.toISOString().split('T')[0];
--- a/site/_data/lib/date.js
+++ b/site/_data/lib/date.js
@@ -2,9 +2,9 @@
  * Returns a formatted date string.
  * See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat
  * for a list of options.
- * @param {Date} date
+ * @param {Date|string} date
  * @param {string} [locale="en"]
- * @param {object} opts
+ * @param {object} [opts]
  * @returns {string}
  */
 const formatDate = (date, locale = 'en', opts = {}) => {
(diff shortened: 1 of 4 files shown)


site/_data/lib/date.js
```diff
@@ -1,4 +1,13 @@
-const formatDate = (date, locale = 'en', opts) => {
+/**
+ * Returns a formatted date string.
+ * See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat
+ * for a list of options.
+ * @param {Date} date
+ * @param {string} [locale="en"]
+ * @param {object} opts
```
MichaelSolati, line 7: Could mark this as optional as well...

```js
 * @param {object} [opts]
 ```

File: site/_data/lib/date.js [hunk 2]
```diff
@@ -1,4 +1,13 @@
-const formatDate = (date, locale = 'en', opts) => {
+/**
+ * Returns a formatted date string.
+ * See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat
+ * for a list of options.
+ * @param {Date} date
```
MichaelSolati, line 5:
```
* @param {Date|string} date
```

site/_data/lib/date.js
```diff
@@ -9,6 +18,13 @@ const formatDate = (date, locale = 'en', opts) => {
...
 };
 
+/**
+ * Returns a date string formatted like "Monday, July 26, 2021".
+ * @param {Date} date
+ * @param {string} [locale="en"]
+ * @param {object} opts
+ * @returns {string}
+ */
 const formatDateLong = (date, locale = 'en', opts = {}) => {
```
MichaelSolati, line 28: See above for argument typings.

site/_data/lib/date.js
```diff
@@ -22,6 +38,13 @@ const formatDateLong = (date, locale = 'en', opts = {}) => {
...
 };
 
+/**
+ * Returns a date string formated like "July 26, 2021".
+ * @param {Date} date
+ * @param {string} [locale="en"]
+ * @param {object} opts
+ * @returns {string}
+ */
 const formatDateShort = (date, locale = 'en', opts = {}) => {
```
MichaelSolati, line 48: See above for argument typings.

MichaelSolati (approved):

site/_data/lib/date.js
samthor: FWIW, TS doesn't need this hint when it sees a `=` in the function itself.

PR #1046, opened 2021-07-28
Adds "Wasm modules sharing restricted to same-origin" article

agektmr:
Adds "Wasm modules sharing restricted to same-origin" article

cc: @lutzvahl


lutzvahl:
Thanks Eiji! LGTM

mihajlija (approved):

PR #1094, opened 2021-08-03
Remove dup manifest version page

### LINKED ISSUE #102: Multiple "manifest version" documents
@dotproto commented on [Thu Dec 03 2020](https://github.com/GoogleChrome/dcc-private/issues/400)

**Describe the bug**
There are 2 different "manifest version" documents in our corpus.

* `docs/extensions/mv2/manifestVersion`
* `docs/extensions/mv2/manifest/manifest_version`

Confusingly, we also have `docs/extensions/mv2/manifest/version`, but this this refers to an extension's version field, not the version of the manifest.

These should be consolidated into a single document.

### LINKED ISSUE #207: Local site always redirects / to /docs/extensions/
**Describe the bug**
If you run the site locally it is redirecting / to /docs/extensions/

@samthor I think redirects-yaml was updated a month ago, could that change be affecting things? Today is my first day reviewing the site since then I think.

samthor:
Fixes #102. Technically it wasn't a duplicate page—it simply described the field, rather than the concept itself. But it was confusing. Removing the page allows it to fall through to a parent page which is a much clearer concept.

This also disallows browsers from caching redirects in dev. I think this has been the cause of some frustration for our team over the months. Fixes #207.


dotproto (approved):
This review only applies to the deletion of `site/en/docs/extensions/mv2/manifest/manifest_version/index.md`

I'm good with dropping `/mv2/manifest/manifest_version/`. To +1 @samthor's comment in the PR description, I think `/mv2/manifestVersion/` is the more historically relevant page. 
It's worth noting that with this change `/<VERSION>/manifestVersion/`does not exist in the MV3 doc set while it does in the MV2 doc set. That said, the live site's redirect logic seems to handle this without issue a problem. It might be good to verify if `/mv2/manifest/manifest_version/` also redirects without an explicit rule.

PR #1095, opened 2021-08-04
boop webdev-infra for imgix auto format

samthor:
No specific issue, but we weren't setting "auto=format" in our srcset URLs, so the correctly sized images weren't being served with an optimal format for the current browser.

- bumps the webdev-infra version
- adds some missing license headers
- removes TODOs around types (fixed now)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -41,7 +41,7 @@
         "rollup": "^2.22.2",
         "striptags": "^3.1.1",
         "unistore": "^3.5.2",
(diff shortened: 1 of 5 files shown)


dotproto (approved):
I don't think my approval does anything here, but LGTM :)

robdodson (approved):

samthor:
FWIW this actually fixes _some_ of the LHCI issues, too.

PR #1105, opened 2021-08-05
move anchor icon to left side of heading

### LINKED ISSUE #476: Heading hash anchors can overflow and create extra whitespace
**Describe the bug**
If there isn't enough room on a line for a heading's hash anchor then it will overflow to the next line.

image

**To Reproduce**
Take a look at this article https://developer.chrome.com/blog/hardware-accelerated-animations/ on a laptop screen.

Possible solutions would be to have the hash anchor be absolutely positioned next to the heading, or change its icon and move it over to the left using negative margin. That's what css-tricks does:
image

samthor:
Fixes #476.

We could make the icon visible all the time, as currently it's hover is a bit awkward - hover heading, move cursor to left, copy/click/etc.


robdodson:
Looks like hash anchors are showing up on the ToC

image

samthor:
I was previously stripping a hash from the trailing end of a heading, now I look for it at the start too.

I've done a tricky thing with CSS; I think the "#" is too bit at its regular font size, but it should still have a large touch target. So I've added `content: '#'` to a smaller anchor.

The Percy change is literally because one of its screenshots has the problem where the anchor forced a newline.

site/_filters/toc.js
```diff
@@ -40,8 +40,8 @@ const headingRe = /<(h[23])(|\s.*?)>(.*?)<\/\1>/gm;
 // Matches an ID attribute, returning the value in the 1st group.
 const idRe = /\bid="(.+?)"/;
 
-// Matches e.g., " #" so we can remove it.
-const trailingHashRe = /\s+#$/;
+// Matches e.g., " #" or "# " so we can remove it.
+const trailingHashRe = /(\s+#$)|(^#\s+)/g;
```
dotproto, line 44: Conceptually it's a little weird that the first half of the expression matches against the end of a string and the second part of the expression matches the start of a string. Reordering these might be a little more obvious to future readers.

```suggestion
const trailingHashRe = /(^#\s+)|(\s+#$)/g;
```

Additionally, the new pattern is no longer just a trailing hash. Should this variable be renamed? Perhaps something like "anchorLinkTextRe".

robdodson (approved):

site/_scss/blocks/_heading-link.scss
```diff
@@ -1,5 +1,30 @@
 .heading-link {
```
dotproto, line 1: On issue with the current implement implementation is that mousing over the space where the anchor # should be does not cause it to appear. Since the invisible element is the part that you must click or copy, this makes it a bit trickier than necessary to get the link.

https://user-images.githubusercontent.com/1047763/128947797-808e966d-ec90-428c-80fe-c1c3b3582cf4.mov

The most straightforward way address this is to use opacity instead of visibility, but I assume we're using visibility and a pseudo element to prevent the # from being read aloud by screen readers.

Alternatively, we could use a negative margin and equal positive padding on the header to make the area to the left of the header react to hovers.

https://user-images.githubusercontent.com/1047763/128948453-b5502713-ab90-490e-bf1a-3f961fb11203.mov

dotproto (approved):
LGTM with a nit on the anchor hover implementation.

site/_filters/toc.js
samthor: Ack done.

dotproto (approved):
LGTM

PR #1106, opened 2021-08-05
try to fix lhci

samthor:
This tries to fix LHCI by two things:

- replacing the image on @agektmr recent blog post with a 4:3 one (it was "too big" and Lighthouse complains)
- simplify the DOM on the Extensions page by removing redundant `<div>`

### DIFF
--- a/site/_includes/macros/project-sections.njk
+++ b/site/_includes/macros/project-sections.njk
@@ -64,24 +64,22 @@ have been defined in the _data/docs/*.yml file for this project.
         #}
         {% set isExternal = helpers.isExternalLink(item.url) %}
         <li>
-          <div>
-            {% if type === 'column' %}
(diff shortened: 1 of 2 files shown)


samthor:
Excitingly this seems to work. I'm a bit worried about the complexity check since we're pretty tight at this point. We also clearly have more complex pages but this is just a canonical one we're testing.

robdodson (approved):

robdodson (changes requested):
This seems like it breaks a page.

Compare
https://developer.chrome.com/docs/extensions/
to
https://deploy-preview-1106--developer-chrome-com.netlify.app/docs/extensions/

Look at the introducing manifest v3 section

samthor:
Fixed, continues to pass LHCI.

robdodson (approved):

PR #1108, opened 2021-08-05
[CWS CUJs] Update dates 

AmySteam:
I forgot to add the updated dates on some files, and this PR add the correct dates when these files were updated.


samdutton (approved):
LGTM — but do we want to make all updated dates the same?

AmySteam:
@samdutton I think I worked on "Getting started" earlier than the other two files, I just forgot to add the updated date :confounded: . I'll go ahead and merge :+1:

dotproto (approved):
Thanks for this :)

PR #1134, opened 2021-08-10
show manifest keys required for API

### LINKED ISSUE #622: The chrome.automation API information is not detailed enough.
What is the required permission is not mentioned in the chrome.automation document?

1. I assumed it was `automation`, but there was an error `Permission'automation' is unknown or URL pattern is malformed.`.
2. I assumed that permission might not be needed, and printed `chrome.automation` but the result was indeed `undefined`.

**I performed the above test in the Dev channel according to the documentation.**

https://developer.chrome.com/docs/extensions/reference/automation/

samthor:
We've had an issue or two which come down to the fact that APIs don't appear unless you set the appropriate key in the manifest, and that can be confusing.

I think we usually document this in the fulltext, but not always, so I'd like to add it to the top here. However, I'm not really sure if it needs more explanation—the [manifest page](https://developer.chrome.com/docs/extensions/mv3/manifest/) doesn't make it clear either (although to be fair it just has links to other things).

Fixes #622.

Screen Shot 2021-08-11 at 09 48 39

### DIFF
--- a/site/_includes/layouts/namespace-reference.njk
+++ b/site/_includes/layouts/namespace-reference.njk
@@ -81,11 +81,11 @@
                 </div>
               </li>
             {% endif %}
-            {% if feature.manifestKeys.length %}
+            {% if feature.manifestRequirements.length %}
               <li>
                 <div class="code-sections__label">Manifest Keys</div>
                 <div>
-                  {% for manifestKey in feature.manifestKeys %}
+                  {% for manifestKey in feature.manifestRequirements %}
                     <code>{{ manifestKey }}</code><br />
                   {% endfor %}
                 </div>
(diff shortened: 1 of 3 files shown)


site/_includes/layouts/namespace-reference.njk
```diff
@@ -81,7 +81,16 @@
...
             {% endif %}
-
+            {% if feature.manifestRequirements.length %}
+              <li>
+                <div class="code-sections__label">Manifest Keys</div>
+                <div>
+                  {% for manifestKey in feature.manifestRequirements %}
+                    <code>{{ manifestKey }}</code><br />
+                  {% endfor %}
+                </div>
```
dotproto, line 91: I feel like the current copy may be a little too terse. For the sake of experimentation, I tried another presentation. WDYT?

> Screen Shot 2021-08-10 at 5 05 50 PM

Offhand I'm not sure what other APIs would display this message, so I don't know if the suggested copy is accurate.

dotproto (approved):
IMO the current text is acceptable, but I think it could be more useful if we polish prompt a little.

PR #1143, opened 2021-08-11
don't crash on bad authors

samthor:
We were doing this check in post-author.njk, but not in card-authors.njk.


MichaelSolati:
I'll need to update #1091 for this. That's not a complaint, just adding a reminder 🦦

MichaelSolati (approved):

samthor:
Sounds good. You might be able to crash or make noise another way - I just found it frustrating while testing that I got a weird image error instead of knowing that I just typo'ed the author name.

PR #1148, opened 2021-08-12
render markdown, strip html tags in indexing

### LINKED ISSUE #917: Search results are overflowing.

### LINKED ISSUE #924: Search results show raw content of {% Img %} shortcode
Search results show raw content of {% Img %} shortcode - seems like algolia is getting raw content instead of already rendered html.

samthor:
Fixes #917 
Fixes #924 

I know @devnook tried to solve this in #923. I disagree with @robdodson 's comment, I think we should `break-word` just in case.

Fixes a minor display issue too; if we don't have content on page (which happens for some of the structural pages), fall back to the meta description.

### DIFF
--- a/site/_collections/algolia.js
+++ b/site/_collections/algolia.js
@@ -21,7 +21,6 @@
 
 const {createHash} = require('crypto');
 const {generateImgixSrc} = require('../_shortcodes/Img');
-const {filterDrafts} = require('../_utils/drafts');
 const {stripDefaultLocale} = require('../_filters/urls');
 const striptags = require('striptags');
 
@@ -54,7 +53,8 @@ module.exports = collections => {
   const toIndex = collections.getAllSorted().filter(item => {
     const {data} = item;
 
-    // Filter out pages we don't want to index.
+    // Filter out pages we don't want to index. This includes drafts, which we
+    // always filter (even in dev, since we're not really indexing here).
     if (
       data.disable_algolia ||
       data.noindex ||
@@ -85,11 +85,15 @@ module.exports = collections => {
     };
 
     // The item is dumped to JSON, but we can't get at the underlying post's
-    // templateContent until the rendering state of 11ty. Define a getter
-    // which will be called at the later point.
+    // templateContent until the rendering state of 11ty: the templateContent
+    // prop itself is a getter.
+    // (templateContent returns the final rendered HTML for this page, not
+    // including the layout).
+    // So, define our own getter which will work at that later point.
     Object.defineProperty(algoliaCollectionItem, 'content', {
       get() {
         // Strip HTML tags and limit to a sensible size for indexing.
+        // As stated above, there's no more Markdown in this content.
         const text = striptags(item.templateContent ?? '');
         return limitText(text);
       },
--- a/site/_collections/algolia.js
+++ b/site/_collections/algolia.js
@@ -14,12 +14,16 @@
  * limitations under the License.
  */
 
-const removeMarkdown = require('remove-markdown');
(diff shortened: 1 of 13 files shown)


robdodson:
> I know @devnook tried to solve this in #923. I disagree with @robdodson 's comment, I think we should break-word just in case.

What case are you worried about? One that I could think of is if someone tries to have a `reallyLongMethodName` but I think our site's `code` style may already cause that to break (I _think_ ... would be good to test).

robdodson:
Visual comparison with and without break-all.

With break-all pretty much every word near the right-hand edge will get broken in a weird way.

Screen Shot 2021-08-12 at 4 11 29 PM

If you remove the style then the words wrap naturally

Screen Shot 2021-08-12 at 4 12 55 PM

samthor:
You're demoing `break-all`? The `break-word` option, despite its confusing name, avoids breaking words where possible. It only breaks words too long to render at once. The [example on MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break) is good for this: with `break-word`, the first two words remain on their own lines, it's only the third really huge word that's split up.

robdodson:
ah sorry, I didn't notice that you said `break-word`. I thought you were saying you disagreed with my comment not to use `break-all` in Ewa's PR. Yeah, `break-word` is fine.

site/_utils/drafts.js
```diff
@@ -5,12 +5,11 @@
...
-  } else {
+const filterDrafts = item => {
+  if (process.env.NODE_ENV === 'production') {
     return !item.data.draft;
   }
+  return true; // include everything in non-prod
 };
 
-module.exports = {drafts};
+module.exports = {filterDrafts};
```
devnook, line 15: I'd suggest filterOutDrafts or removeDrafts. Also, this looks like an eleventy filter, should it be in site/_filters?  How does this compare to _filters/removeDrafts?

site/_collections/algolia.js
```diff
@@ -47,38 +51,55 @@ function limitText(content, limit = 7500) {
...
       objectID: createHash('md5').update(item.url).digest('hex'),
     };
 
+    // The item is dumped to JSON, but we can't get at the underlying post's
+    // templateContent until the rendering state of 11ty. Define a getter
+    // which will be called at the later point.
+    Object.defineProperty(algoliaCollectionItem, 'content', {
+      get() {
+        // Strip HTML tags and limit to a sensible size for indexing.
+        const text = striptags(item.templateContent ?? '');
```
devnook, line 93: Is this already rendered content, e.g. with shortcodes rendered in?

samthor:
yes - templateContent is itself a getter which renders the final page
content but only when 11ty is rendering pages. If we get it here, we crash.

site/_utils/drafts.js
samthor: It's not really an 11ty filter, which is confusing. It's just a helper for collection filtering. I've made that clearer, and I like the updated name—thanks.

site/_collections/algolia.js
samthor: yes - templateContent is itself a getter which renders the final page content but only when 11ty is rendering pages. If we get it here, we crash.

robdodson:
@samthor any idea why the linter is failing? it seems to be missing a module

samthor:
I expanded the typechecking a tiny bit. I've fixed the other types issues in an extra commit here.

devnook (approved):

PR #1159, opened 2021-08-13
always render 'authors'

### LINKED ISSUE #496: Add date to blog index page even if there's no author
On the [blog index page](https://developer.chrome.com/blog/), if a post doesn't have an author, only the title and tags appear. But if there's an author, it shows the author, and publication date. See screenshot below:

Screen Shot 2021-03-04 at 11 07 37 AM

Add publication date even if there's no author.

samthor:
Fixes #496. Renders authors even if we have zero, which causes the date to render.

Screen Shot 2021-08-14 at 08 25 20

### DIFF
--- a/site/_includes/macros/cards/blog-card.njk
+++ b/site/_includes/macros/cards/blog-card.njk
@@ -47,11 +47,10 @@
     </div>
   {% endif %}
 
-  {% if pd.authors and pd.authors.length %}
-    <div class="gap-top-300">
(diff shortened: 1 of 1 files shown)


dotproto (approved):
At first glance I was wondering if it even made sense to put the date in an "authors card", but after playing around with the HTML output the this structure makes sense. I don't think it's worth the effort to refactor the "card-authors" element for the ideological purity of a more generic container class.

LGTM.

PR #1160, opened 2021-08-13
fix < > to &lt; &gt; on match_patterns page

### LINKED ISSUE #366: Incorrect match pattern documentation.
Go to https://developer.chrome.com/docs/extensions/mv2/match_patterns/ and https://developer.chrome.com/docs/extensions/mv3/match_patterns/. 

See the section after "Here's the basic syntax:". It is:

>  := ://
>  := '*' | 'http' | 'https' | 'file' | 'ftp'
>  := '*' | '*.' +
>  := '/' 

It should be (as seen at https://3-72-0-dot-chrome-apps-doc.appspot.com/extensions/match_patterns):

`<url-pattern> := <scheme>://<host><path>`
`<scheme> := '*' | 'http' | 'https' | 'file' | 'ftp'`
`<host> := '*' | '*.' <any char except '/' and '*'>+`
`<path> := '/' <any chars>`

We probably just need to escape < and > to display them,

samthor:
Fixes #366. We literally rendered these as HTML elements.

This is a specific case; is there any case where we _want_ HTML elements to appear inside code blocks? Is this really a problem with our Markdown parser?

### DIFF
--- a/site/en/docs/extensions/mv2/match_patterns/index.md
+++ b/site/en/docs/extensions/mv2/match_patterns/index.md
@@ -25,10 +25,10 @@ any URL that starts with a permitted scheme. Each match pattern has 3 parts:
 Here's the basic syntax:
 
 ```text
-<url-pattern> := <scheme>://<host><path>
-<scheme> := '*' | 'http' | 'https' | 'file' | 'ftp' | 'urn'
-<host> := '*' | '*.' <any char except '/' and '*'>+
-<path> := '/' <any chars>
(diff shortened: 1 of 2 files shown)


dotproto (changes requested):
One code block uses &amp;lt; and &gt;, the other uses &amp;tl; and &amp;gt;. Is there a reason for this inconsistency?

dotproto (approved):
LGTM :shipit:

PR #1169, opened 2021-08-16
Breadcrumbs fixes

### LINKED ISSUE #84: Some pages have incorrect heading order.
@robdodson commented on [Sun Nov 29 2020](https://github.com/GoogleChrome/dcc-private/issues/337)

Pages like /tags/ rely on the breadcrumb widget to provide their heading. If the breadcrumb is the primary heading on the page then it should use an h1 (styled to look like a breadcrumb).

I think I just need to go through all of the pages and verify the heading order is correct, and possibly tweak the breadcrumb widget where it's needed.

samthor:
Fixes #84, and lets us set the last breadcrumb as a heading on certain pages.

Fixes some other minor issues:
- Adds a title for "Origin Trials" tag, and fixes a bug that a missing tag would cause a random tag name to appear instead (Nunjucks needs us to set a default)
- makes sure e.g., "API Reference" is highlighted in navigation even on sub-pages
- puts the share button in the middle of the breadcrumbs, otherwise it caused pages with a share button to be "taller"
- removes `.cluster` from breadcrumbs as it did `overflow: hidden`, which caused:
Screen Shot 2021-08-16 at 18 46 42

The Percy changes are actually relevant, they showcase a bunch of these issues.

### DIFF
--- a/site/_data/lib/links.js
+++ b/site/_data/lib/links.js
@@ -34,7 +34,7 @@ function getLinkActiveState(itemUrl, pageUrl) {
  *   parent: SectionLinkItem|undefined,
  *   active: boolean,
  * }}
- * @type {{never}}
+ * @type {never}
  */
 // eslint-disable-next-line no-unused-vars
 let SectionLinkItem;
--- a/site/_data/lib/links.js
+++ b/site/_data/lib/links.js
@@ -28,12 +28,13 @@ function getLinkActiveState(itemUrl, pageUrl) {
 }
 
 /**
- * @type {{
- *   sections?: Section[],
+ * @typedef {{
+ *   sections?: SectionLinkItem[],
  *   url?: string,
  *   parent: SectionLinkItem|undefined,
  *   active: boolean,
  * }}
+ * @type {{never}}
  */
 // eslint-disable-next-line no-unused-vars
 let SectionLinkItem;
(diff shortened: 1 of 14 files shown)


site/_data/lib/links.js
```diff
@@ -28,12 +28,13 @@ function getLinkActiveState(itemUrl, pageUrl) {
...
 /**
- * @type {{
- *   sections?: Section[],
+ * @typedef {{
+ *   sections?: SectionLinkItem[],
  *   url?: string,
  *   parent: SectionLinkItem|undefined,
  *   active: boolean,
  * }}
+ * @type {{never}}
```
devnook, line 37: Why never here?

site/_includes/partials/breadcrumbs.njk
```diff
@@ -3,23 +3,37 @@
...
+        {# This is the last segment and we need to be a large <h1>. #}
+
+        {% if part.url %}
+          <h1 class="type--h5"><a class="link decoration-none no-visited" href="{{ part.url }}">{{ part.title }}</a></h1>
+        {% else %}
+          <h1 class="type--h5">{{ part.title }}</h1>
+        {% endif %}
+
+      {% else %}
+        {# This is a normal segment. #}
```
devnook, line 24: normal -> middle?

devnook (approved):

site/_data/lib/links.js
samthor: I actually made a typo, this is supposed to be `{never}` :)

This means that if you accidentally use SectionLinkItem as a variable (e.g., `const foo = SectionLinkItem`), TSC is more likely to explode. Otherwise the default "type" of the variable is `any`.

Issue #89, opened 2020-12-17
Docs subsections should start with an index page


robdodson:
@robdodson commented on [Fri Oct 30 2020](https://github.com/GoogleChrome/dcc-private/issues/203)

**Is your feature request related to a problem? Please describe.**

We want to ensure that authors create an index page for their subsections, here's an example where the first link in the overview section goes to an index page:

```
- title: i18n.docs.extensions.overview
  sections:
    - url: /docs/extensions/overview
    - url: /docs/extensions/overview/foo
    - url: /docs/extensions/overview/bar
```

**Describe the solution you'd like**
It would be good to have a linter that enforces this structure. Otherwise folks may end up creating subsections with empty urls.

For example, if I did this:

```
- title: i18n.docs.extensions.overview
  sections:
    - url: /docs/extensions/overview/qux
    - url: /docs/extensions/overview/foo
    - url: /docs/extensions/overview/bar
```

A user might logically go to `/docs/extensions/overview/` to see what's there, only to be greeted with a 404.

robdodson:
yeah, thankfully the nacl stuff is all deprecated, so it's less of an issue.

> .//webstore/webstore_api
> .//webstore/payments-iap

Are these real pages? https://developer.chrome.com/docs/webstore/webstore_api/ seems to 404 and /webstore/payments-iap redirects to the /webstore/ page

samthor:
The PR I've sent removes "payments-iap" because it just had images in that folder, no content. "webstore_api" has a bunch of subfolders I haven't worked out if are needed.

I actually suspect most of "webstore_api" can be deleted since it seems to be about in-app payments, which we no longer support? @dotproto @awfuchs

dotproto:
I just did a pass on the contents of the webstore_api directory and included my initial impressions below.

**Keep**
- `docs/webstore/webstore_api/items`
- `docs/webstore/webstore_api/items/get`
- `docs/webstore/webstore_api/items/insert`
- `docs/webstore/webstore_api/items/publish`
- `docs/webstore/webstore_api/items/update`

**Keep until [payments API is fully sunset (TBD)](https://developer.chrome.com/docs/webstore/cws-payments-deprecation/#milestone-dates)**
- `docs/webstore/webstore_api/userLicenses`
- `docs/webstore/webstore_api/userLicenses/getUserLicense`
- `docs/webstore/webstore_api/payments`
- `docs/webstore/webstore_api/payments/list`

[...]

PR #1179, opened 2021-08-17
hoist /overview/ pages up where there's no parent index

samthor:
Addresses (but does not fix) #89.

Fixes a bunch of links _and_ adds redirects to the yaml file, in case these are being linked to externally.

cc @joycetoh8 as I'm moving the /overview/ page of Android stuff up one folder to where there's currently no index.md file.


site/en/docs/multidevice/overview/index.md
```diff
@@ -10,29 +10,26 @@ Google provides three ways to run web content on tablets, smartphones, and other
 - Using Chrome Mobile for devices running Android 4.0 and later, and iOS.
```
devnook, line 10: Should this file be also moved out of /overview/ dir?

devnook (approved):

site/en/docs/multidevice/overview/index.md
samthor: No, because the multidevice folder above it actually has a valid index page.

PR #1181, opened 2021-08-17
Describe about the args property

bangseongbeom:
References which are using the `args`:

- https://developer.chrome.com/blog/crx-scripting-api/#injecting-a-function-with-arguments
- https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration/#executing-arbitrary-strings


samthor (approved):

samthor:
Thank you for this and the other PR. 👍

PR #1225, opened 2021-08-18
add feeds for all tags

### LINKED ISSUE #613: Support RSS feed by tags
Should we should support RSS feed by tag? For example, users might be interested in `New in Chrome` blog posts but not others. @robdodson @samthor 

**Additional context**
Twitter feedback: https://twitter.com/mrleblanc101/status/1379622739867209729

samthor:
Fixes #613. Adds a feed for all blog tags.

- Merges all `chrome-xx` tags into a single `chrome.xml` feed.
- Adds an icon as part of breadcrumbs to hint that this page has a custom feed.
- Updates the `<link>` on those same pages to be the custom feed.

Also moves /style-guide/ into handbook, as we were incorrectly indexing it/treating it special at the top of the site. It feels part of the handbook 🤷 

This puts everything inside /feeds/ in order to remain compatiable with the existing devtools/extensions feeds. If we add author feeds we can put them in a subfolder or something.

for @chybie and @MichaelSolati


site/_data/i18n/tags.yml
```diff
@@ -76,3 +82,9 @@ new-in-devtools:
 
 devtools-engineering:
   en: DevTools Engineering
+
+origin-trials:
```
chybie, line 86: Understand the `chrome` tag is added on purpose. Are these tags as well?

chybie (approved):
LGTM

site/_data/i18n/tags.yml
samthor: Yes, they're actually missing and we're rendering bad data where people have tagged articles with them D:

PR #1230, opened 2021-08-18
Updated origin trial details and wording

samdutton:


samthor (approved):
This makes me wonder if we should tag these pages with the relevant Chrome versions too—right now we just tag blog posts.

No change needed now though.

mihajlija (approved):

PR #1243, opened 2021-08-20
add releases card on home page

samthor:
This actually isn't tracked but we have this data and I wanted to make the homepage more living.

This simply shows the current Chrome release along with the date for the release of the next on the homepage of the site. This could later lead to a complete Releases page. We have designs for both.

See screenshot for this in context:

Screen Shot 2021-08-20 at 16 38 26

### DIFF
--- a/site/_data/chrome.js
+++ b/site/_data/chrome.js
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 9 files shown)


robdodson:
Overall looks really cool. A few quick thoughts:

- I think the date on the second item may be unclear to the reader. Does that mean 93 will be stable on August 31? Or does it mean something else? Maybe it should say Next version: August 31 or..something. Might ask in the larger team chat what they think.
- Small a11y nit, It might be good to use the `visually-hidden` class and a span with some hidden text to make it more accessible to screen readers.

```
<div><span class="visually-hidden">Current stable version, Chrome </span>92</div>
<div aria-hidden="true">Current stable version</div>
```

robdodson (approved):

PR #1245, opened 2021-08-20
Update index.md with Cloud-based emulator mention

testingbot:
Added TestingBot under Cloud-based emulators and simulators.


mihajlija:
Hi, featuring commercial tools on the site is an exception only for cases where there aren't many open-source tools available. We never aim to create exhaustive lists of all available tools, and the tools we do feature must have high usage thresholds. Based on [public stats](https://stackshare.io/stackups/lambdatest-vs-testingbot), TestingBot usage is significantly smaller than usage of other tools mentioned in the post, so we can't include it.

PR #1258, opened 2021-08-24
Fix link in first aside

thijskh:
The link `register` does not exist, `csw-register` does.


thijskh:
@googlebot I signed it!

samthor:
we should probably have a linter for this. Thanks for PR.

Issue #843, opened 2021-06-02
Let posts have custom banners.


jecfish:
At the moment, I include that in the beginning of each blog post. Example: https://developer.chrome.com/blog/new-in-devtools-89

robdodson:
This seems like something we could move into a banner (since it has a clickable call to action). This is _almost_ possible today. If you add:

```yaml
override:banner:
  type: info
  text: Hello world
  actions:
    - text: Apply
      href: https://google.com
    - text: Dismiss
```

to an existing post, it will display a custom banner just for that page. But you **should not** do this yet, because clicking anything in that banner will cause the site to "forget" any other banners that the user has previously dismissed. So I'd suggest we turn this into a feature request to make the site support multiple dismissible banners.

PR #1270, opened 2021-08-25
[devtools] Move banner to partials, localised reach out footer

jecfish:
Fix: #843


samthor (commented):
I'd like to see the language code put as part of the directory path, not the filename. e.g.:

partials/devtools/zh/banner.md

Sorry, I know it's a pain to reorder them. I know in this case y'all link specifically to the right language (because you don't want to fall back to English), but in the _very_ near future I can imagine a macro which pulls in the correct one, either allowing the fallback or just helpfully inserting the locale for you.

PR #1318, opened 2021-08-31
cleanup pagination types, disambiguate 'real' vs fake pagination

samthor:
I think a big problem with the types in #1091 is that we're starting from a point that implies some incorrect properties.

The `pagination` property on `PaginatedPage` was typed as if it was a pure Eleventy object, but it's not: it's just our multidimensional contents that are showed on a single page.

- by renaming it to `contents`, it's clear this is something different
- we modify `addPagination` to clearly take a templated type, for future types that are _not_ just 11ty pages

Eleventy in the "normal" single-dimension mode _does support_ a number of items on the same page, but we just don't use that (because it just lumps every item together, which doesn't work for us), we render a single `PaginatedPage`, which Eleventy helpfully gives to us directly (1 = object, more = you get an array) in our configured variable "paged".


samthor (commented):
@MichaelSolati I also have an extension of this which shows how to use it for author posts. Let's chat when I'm online tomorrow.

PR #1329, opened 2021-09-01
Fix open options example code

benkaiser:
addEventListener takes 2 args, I'm assuming the other one intended here is 'click'

-
-
-


samthor:
That seems likely! Sorry, but even for small changes we need you to sign the CLA. Otherwise we'll make the same change ourselves.

benkaiser:
@googlebot I signed it!

PR #1372, opened 2021-09-10
Add supported language links for blog post

### LINKED ISSUE #1358: Add links to translated content
Making the translated content more discoverable. Design is in! Pending implementation

desktop-blog

chybie:
Fixes #1358

There are a few points which I am not sure what is the best way to deal with it:
1. The current implement is using shortcode, but it could be a web component. The code is similar to `LanguageSelect` component in the `webdev-infra` project. Let me know if web component is preferred (locally in the project or put it in `webdev-infra`)
1. The supported language names are currently hardcoded in `webdev-infra/LanguageSelect`. I did the same now in my shortcode file. Ideally we should export it from the `webdev-infra` or put it in a global variable file... said `_data/site.json`?
1. For localized url that isn't english, I am not able to use relative path in href.  (e.g. if current locale = "es", this link`<a href="/ja/url">` will redirect to `/es/ja/url`. I am using absolute path now as a workaround. Any better option for this?

Test pages:
1. Page with no translation available: http://localhost:8080/blog/new-in-devtools-91/
2. English page with translation: http://localhost:8080/blog/new-in-devtools-93/
3. Translated page in Spanish: http://localhost:8080/es/blog/new-in-devtools-93/

Screen Shot 2021-09-10 at 4 20 50 PM

### DIFF
--- a/site/_includes/partials/post-headline.njk
+++ b/site/_includes/partials/post-headline.njk
@@ -12,9 +12,8 @@
     <span>•</span>
     Updated on <time>{{ helpers.formatDateLong(updated, locale) }}</time>
   {% endif %}
-  </p>
-  <p class="type--caption language-list">
-    Translated to: {% LanguageList page.url, site.locales, collections, locale %}
+  <br>
+  {% LanguageList page.url, site, collections, locale %}
   </p>
 </div>
 {% endif %}
--- a/site/_includes/partials/post-headline.njk
+++ b/site/_includes/partials/post-headline.njk
@@ -5,15 +5,18 @@
 {% endif %}
 
 {% if date %}
-<div class="dates">
-  <p class="type--caption">
+  <p class="type--caption color-secondary-text">
     Published on <time>{{ helpers.formatDateLong(date, locale) }}</time>
-  {% if updated %}
+    {% if updated %}
     <span>•</span>
     Updated on <time>{{ helpers.formatDateLong(updated, locale) }}</time>
   {% endif %}
-  <br>
-  {% LanguageList page.url, site, collections, locale %}
   </p>
-</div>
+{% endif %}
+
+{% set languageListHtml %}{% LanguageList page.url, site, collections, locale %}{% endset %}
+{% if languageListHtml %}
+<p class="type--caption color-secondary-text">
+  {{ languageListHtml | safe }}
+</p>
 {% endif %}
--- a/site/_includes/partials/post-headline.njk
+++ b/site/_includes/partials/post-headline.njk
@@ -13,5 +13,8 @@
     Updated on <time>{{ helpers.formatDateLong(updated, locale) }}</time>
   {% endif %}
   </p>
+  <p class="type--caption language-list">
+    Translated to: {% LanguageList page.url, site.locales, collections, locale %}
+  </p>
 </div>
 {% endif %}
--- a/site/_scss/blocks/_language-list.scss
+++ b/site/_scss/blocks/_language-list.scss
@@ -1,4 +1,0 @@
-.language-list a {
-  color: var(--color-secondary-text);
-  margin-inline-start: 3px;
-}
--- a/site/_scss/blocks/_language-list.scss
+++ b/site/_scss/blocks/_language-list.scss
@@ -1,0 +1,4 @@
+.language-list a {
+  color: var(--color-secondary-text);
+  margin-inline-start: 3px;
+}
--- a/site/_scss/main.scss
+++ b/site/_scss/main.scss
@@ -30,7 +30,6 @@
 @import 'blocks/side-nav';
 @import 'blocks/post-authors';
 @import 'blocks/card-authors';
-@import 'blocks/dates';
 @import 'blocks/project-icon';
 @import 'blocks/project-hero';
 @import 'blocks/project-sections';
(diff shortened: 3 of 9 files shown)


site/_shortcodes/LanguageList.js
```diff
@@ -0,0 +1,53 @@
...
+  es: 'Español',
+  ko: '한국어',
+  zh: '中文',
+  ru: 'Pусский',
+  pt: 'Português',
+  ja: '日本語',
+};
+
+/**
+ * Outputs a list of <link hreflang=""> tags with alternate language versions
```
samthor, line 20: I think you mean `<a href="..."` (did you copy this comment from the head stuff?

site/_shortcodes/LanguageList.js
```diff
@@ -0,0 +1,53 @@
...
+  if (!url) {
+    return;
+  }
+
+  const cleanUrl = stripLocale(url);
+  const hreflangs = getLocalizedPaths(cleanUrl, site.locales).filter(hreflang =>
+    findByUrl(collections.all, hreflang[0])
+  );
+  const links = hreflangs
+    // eslint-disable-next-line no-unused-vars
```
samthor, line 37: You can fix this by just not including `_`, so e.g.:

```js
    .filter(([, code]) => code !== locale)
```

site/_scss/blocks/_language-list.scss
```diff
@@ -0,0 +1,4 @@
+.language-list a {
+  color: var(--color-secondary-text);
+  margin-inline-start: 3px;
```
samthor, line 3: This is here because you seem to not have whitespace between the language choices, so the actual HTML looks like:

```html
Language (foo),Language (bar)
```

... maybe fix the HTML generation instead?

site/_shortcodes/LanguageList.js
```diff
@@ -0,0 +1,53 @@
...
+    .map(([urlPath, code]) => {
+      return `<a href="${path.join(site.url, urlPath)}">${languageNames[code]}
+        (${code})</a>`;
+    });
+
+  let html = '';
+
+  if (links.length) {
+    html = `<span class="language-list">Translated to:
+    ${links.join(',')}</span>`;
```
samthor, line 48: this should be ", " with a space

site/_shortcodes/LanguageList.js
```diff
@@ -0,0 +1,53 @@
...
+  const cleanUrl = stripLocale(url);
+  const hreflangs = getLocalizedPaths(cleanUrl, site.locales).filter(hreflang =>
+    findByUrl(collections.all, hreflang[0])
+  );
+  const links = hreflangs
+    // eslint-disable-next-line no-unused-vars
+    .filter(([_, code]) => code !== locale)
+    .map(([urlPath, code]) => {
+      return `<a href="${path.join(site.url, urlPath)}">${languageNames[code]}
+        (${code})</a>`;
```
samthor, line 41: I don't think Jiwoong's design has the `({$code})` part. I don't think it adds much? Can we drop it?

site/_scss/blocks/_language-list.scss
```diff
@@ -0,0 +1,4 @@
+.language-list a {
+  color: var(--color-secondary-text);
```
samthor, line 2: Rather than this, we can just use `class="color-secondary-text"` on the `<a>` itself. So you can actually remove this whole CSS file.

site/_includes/partials/post-headline.njk
```diff
@@ -12,6 +12,8 @@
     <span>•</span>
     Updated on <time>{{ helpers.formatDateLong(updated, locale) }}</time>
   {% endif %}
+  <br>
+  {% LanguageList page.url, site, collections, locale %}
```
samthor, line 16: This has a couple of problems. If the article has no date, then this won't be shown at all.

You can change this whole section a bit to look maybe like (noting we can drop `.dates` because it just sets a color, wich we can instead do with utility class "color-secondary-text" instead):

```js
{% if date %}
<p class="type--caption color-secondary-text">...</p>
{% endif %}
{% set languageListHtml %}{% LanguageList ... %}{% endset %}
{% if languageListHtml %}
<p class="type--caption color-secondary-text">...</p>
{% endif %}
```

samthor (commented):
Sorry to make you fix lots of unrelated stuff but hey here you are!

This is great code otherwise, thanks for your work! 👍

samthor:
Also I will chime in and say—shortcode is perfect for this. These should be real links, not a dropdown or JS chooser.

Some [suggestions are here](https://www.w3.org/International/questions/qa-site-conneg), but I think you should also add `translate="no"` to the `<a>` here.

site/_shortcodes/LanguageList.js
chybie: Dropped

site/_shortcodes/LanguageList.js
chybie: aha, TIL! I always thought need to put `_` for unused var.

site/_shortcodes/LanguageList.js
chybie: 🙈🙈🙈

chybie:
Thanks @samthor  for the review. Fixed all. Please check again!

site/_scss/main.scss
```diff
@@ -70,7 +70,6 @@
 @import 'blocks/logo';
 @import 'blocks/devtools';
 @import 'blocks/select-loader';
-@import 'blocks/language-list';
```
samthor: Please remove "blocks/_dates". It just had one class that set a color, which we now do with utility CSS.

samthor (approved):

Issue #594, opened 2021-04-01
Allow smart quotes


robdodson:
**Is your feature request related to a problem? Please describe.**
We currently lint for smart quotes and ask authors to remove them. We do this because it's easy for a smart quote to sneak into a code sample (especially if you're copying from Google Docs to markdown). This can break a user's editor and can be hard to debug depending on the editor font that they're using.

However, smart quotes look nice and may aid in reading comprehension, so it would be nice to enable them if we could do so safely. markdown-it has a plugin called `typographer` which may enable us to do this, we just need to test it.

**Describe the solution you'd like**
Authors should be able to use smart quotes in their body content. A smart quote used in a fenced code block should be converted to a straight quote at build time.

**Describe alternatives you've considered**
A short term solution for authors is to add the `<!-- lint disable no-smart-quotes -->` comment to the top of their markdown file to disable this rule.

mathiasbynens:
> Authors should be able to use smart quotes in their body content.

👍 

> A smart quote used in a fenced code block should be converted to a straight quote at build time.

IMHO this is the wrong approach. The files in this repository are the source of truth, and ideally that source is correct. If a code example features a string literal, it should be wrapped in straight quotes, otherwise it’s a syntax error. We shouldn’t try to fix things at build time that should really have been caught during review.

Instead, maybe we can help authors and reviewers to detect these typos early, by building on the lint rule that’s already there?

- Limit the scope of the lint rule to code spans and blocks only (i.e. don’t affect prose).
- Expand the lint rule to include other possibly problematic cases (e.g. `…`).
- Make the lint rule non-blocking (WARN instead of ERR?).

WDYT?

samthor:
I'd be happy to make this only complain inside code blocks.

robdodson:
This came up again today in relation to web.dev.

It seems like we may want to do two things:
- Make this rule only complain inside of code blocks
- Set `typographer` to true in our markdown-it settings so we have consistent quotes across the site (so authors don't mix straight and smart quotes in their prose).

Btw @mathiasbynens do y'all already lint for smart quotes in code blocks on v8.dev? I was curious if there's an existing rule we could borrow.

mathiasbynens:
> Btw @mathiasbynens do y'all already lint for smart quotes in code blocks on v8.dev? I was curious if there's an existing rule we could borrow.

We don’t, unfortunately. We rely on code review to catch issues like this.

PR #1373, opened 2021-09-10
Remove lint rules

devnook:


robdodson:
Fixes #594

robdodson (approved):

samthor:
For context this is to support i18n where we don't get be so restrictive.

PR #1384, opened 2021-09-11
Add missing URL for 'session cookies' link

tannerdolby:
Fixes #1384 

- Adds url definition for [6] in template representing 'session cookies' 

Note: I wasn't entirely sure where the original link was pointing to, so I grabbed one I thought kinda worked and we can go from there, swapping it out for the correct link if need be.


samthor (approved):
Yes this is good, even if the id changes, the Cookies page is a good place to start :)

tannerdolby:
Ok great :) thanks @samthor!

PR #1386, opened 2021-09-11
fix URL switcher links were being mangled

samthor:
The "Translated to:" links were being mangled in a few ways:
- They had a "developer.chrome.com" pathname prefix, which made no sense (the fallback redirect code took over)
- We tried to normalize language code links, disabled now for links marked with `translate="no"`

This also puts some meta strings up for translation ("Translated to:" and so on).

### DIFF
--- a/site/_shortcodes/LanguageList.js
+++ b/site/_shortcodes/LanguageList.js
@@ -28,6 +28,8 @@ function LanguageList(url, site, collections, locale = 'en') {
     return;
   }
 
+  // getLocalizedPath also implicitly strips any locale prefix, but get the clean URL here first
+  // anyway: we want all options, and this makes it clearer to future readers.
   const cleanUrl = stripLocale(url);
   const hreflangs = getLocalizedPaths(cleanUrl, site.locales).filter(
     ([urlPath]) => findByUrl(collections.all, urlPath)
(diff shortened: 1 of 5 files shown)


site/_shortcodes/LanguageList.js
```diff
@@ -36,18 +35,10 @@ function LanguageList(url, site, collections, locale = 'en') {
   const links = hreflangs
     .filter(([, code]) => code !== locale)
     .map(([urlPath, code]) => {
-      return `<a translate="no" lang="${code}"
-        href="${path.join(site.url, urlPath)}">
-        ${languageNames[code]}</a>`;
+      return `<a translate="no" lang="${code}" href="${urlPath}">${languageNames[code]}</a>`;
```
chybie, line 38: can't comment on unedited code. In this case, the `const cleanUrl = stripLocale(url);` step is not necessary. I added the `stripLocale` util in last PR. Can remove that as well.

chybie (approved):

site/_shortcodes/LanguageList.js
samthor: It is redundant because `getLocalizedPaths` does it too, but I've left it in with a comment. I think it makes the code clearer.

PR #1397, opened 2021-09-13
fix post breadcrumbs for 'es'

### LINKED ISSUE #1393: Title shows in breadcrumbs for non-en blog pages
On English pages, the top breadcrumb is just "Blog", since that's where you can go back to. On non-en pages, we show the title of the article, which is redundant since it's literally right there in the page

It should be "Blog" in all cases.

samthor:
Technically fixes #1393 for 'es', but not in the general case.

This will create "/es/tags" and "/es/blog", but not for all languages (just did 'es' as a demo). There's some open questions here:

- 11ty doesn't allow us to programatically generate these pages: we have to just add this for every supported lang, is that fine? (cc @devnook who has worked on i18n content on web.dev)
- should the blog/tags include all posts, even untranslated ones? (cc @chybie and @MichaelSolati who have opinions)
- the feed links are broken (we only generate feeds for English)—is this important?

In general I'm of the view that we will never have full coverage, so it's responsible of us to basically show English fallback content wherever possible.

### DIFF
--- a/site/_data/i18n/tags.yml
+++ b/site/_data/i18n/tags.yml
@@ -87,6 +87,7 @@ devtools:
 
 new-in-devtools:
   en: What's New in DevTools
+  es: Qué hay de nuevo en DevTools
 
 devtools-engineering:
   en: DevTools Engineering
--- a/site/es/blog/blog.11tydata.js
+++ b/site/es/blog/blog.11tydata.js
@@ -1,0 +1,3 @@
+module.exports = {
+  type: 'blogPost',
+};
--- a/site/es/blog/index.md
+++ b/site/es/blog/index.md
@@ -5,6 +5,6 @@ permalink: '{{locale}}/blog/{% if pagination.pageNumber > 0 %}{{ pagination.page
 layout: 'layouts/blog-landing.njk'
 type: landing
 pagination:
-  data: collections.blog-en
+  data: collections.blog-es
   size: 24
 ---
--- a/site/es/tags/index.md
+++ b/site/es/tags/index.md
@@ -1,0 +1,6 @@
+---
+title: 'Tags'
+description: 'Find blog posts tagged to a certain feature'
+permalink: '{{locale}}/tags/index.html'
+layout: 'layouts/tags-landing.njk'
+---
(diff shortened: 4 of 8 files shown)


chybie:
- **English fallback**. I think it's responsible to show the fallback if translated content doesn't exist. However, as a non-english speaker browsing the page, the best scenario would be
    - see native content first, then have a way to show the non-translated one ( a checkbox or so)
    - The reason being is - Given our current speed of translation, the English content would be more than translated content. It could be discouraging when you open an `es/` page but see most of them are English content. Then eye-balling through it to find the translated one.
    - Erm... it's might be a bit tricky to do the above given these are statically generated pages? If that's the case, it would be better to view **just the native content in `/es`**, and have the language switcher place in a more obvious location with an info bar somewhere to inform user that for full blog content, switch to english. I include a similar message at every translated What's New in DevTools post (eg. [link](https://developer.chrome.com/es/blog/new-in-devtools-93/#whats-new))

- **Broken feed link**. I think the feed link can come later, either hide it for other language or leave it as english for now is fine.

- **Broken search**. Not related to this bug, the search isn't working yet for `es`.

devnook:
> * 11ty doesn't allow us to programatically generate these pages: we have to just add this for every supported lang, is that fine? (cc @devnook who has worked on i18n content on web.dev)

I think it's fine, there is always some setup for adding a new language required. I have a draft doc page on how to add a new lang on web.dev, including which files should be copied over from /en/. I'll share on chat.

samthor:
I'm going to update this for the six supported languages today.

samthor:
FYI @chybie and @devnook to take another look now that it's for all langs

site/_data/i18n/tags.yml
```diff
@@ -90,6 +90,7 @@ devtools:
 
 new-in-devtools:
   en: What's New in DevTools
+  es: Qué hay de nuevo en DevTools
```
chybie, line 93: Add other languages here too? To make it easier for you:

```
ja: DevTools の新機能
ko: DevTools 의 새로운 기능
ru: Новинки DevTools
zh: DevTools 新功能
pt: O que há de novo no DevTools
```

PR #1399, opened 2021-09-13
Chrome label fixes

### LINKED ISSUE #1380: Future Chrome versions cause string errors
We match some of them to e.g., Canary, and that string doesn't exist. (It may be worth getting rid of names for anything but Stable and Beta, since I think the others aren't really in common parlance. We can just call them "Future" versions.)

samthor:
Fixes #1380.

Also, makes sure that we show any "imminent" release that occurs while `beta == stable + 2`, which happened recently because of the release schedule shift, but could happen again.

### DIFF
--- a/site/_data/chrome.js
+++ b/site/_data/chrome.js
@@ -70,8 +70,9 @@ async function dataForVersion(channel, version) {
 /**
  * Generates Chrome version information.
  *
- * Returns the current version for each channel (i.e., "stable" => 85) as well as its release
- * information, if available.
+ * While there are many channels, we just care about 'stable' and 'beta'. If we can't find beta, then
+ * assume beta is stable+1. Beta is not always stable+1: it can be stable+2 if Chrome is
+ * mid-release. We add a fake release 'pending' in this case.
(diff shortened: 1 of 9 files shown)


site/_data/chrome.js
```diff
@@ -89,28 +90,52 @@ module.exports = async function buildVersionInformation() {
...
+  const {beta, stable} = raw;
+  if (beta === stable + 1) {
+    // great
+  } else if (beta === stable + 2) {
+    // This can occur if Chrome does something weird with releases, basically the previous beta
+    // is about to be released but it's not done yet.
+    raw['pending'] = stable + 1;
+  } else {
+    console.warn(raw);
+    throw new Error('bad Chrome release data, beta is >+2 from stable');
```
samthor, line 122: One note here, I want a lighter-touch type of an error: crash in prod but warn/defaults in dev. Perhaps that's for another PR. 🤷

MichaelSolati (approved):
Dense, but seems to work and definitely looks nicer. LGTM

PR #1419, opened 2021-09-14
Improve formatting

mathiasbynens:
I think some things went missing when converting the post to Markdown. I tried to correct this here — PTAL.

I think the `styles.replaceSync` code example is still broken but it wasn’t 100% clear to me what it was intended to look like, so I didn’t address that. Please look into that as well!


kriti-sapra:
Hey! Thank you for those changes I didn't notice them! For `styles.replaceSync` I think the official code has the same indentation but for readability I can add spaces.

chybie (approved):

chybie (approved):

PR #1420, opened 2021-09-14
Fix typo in example

yi-gu:


yi-gu:
@agektmr PTAL. Thanks!

PR #1431, opened 2021-09-16
Fix typo where the variable name was a keyword

### LINKED ISSUE #1429: Example causes errors.
**Describe the bug**
Using language keyword as a variable name in the example. 

https://developer.chrome.com/docs/extensions/reference/runtime/#example-get-url

**Screenshots**
image

tannerdolby:
Fixes #1429

- Changes `let const = ..` to `const img = ..` (We could use `let img = ..`, i just reached for `const` first)


samthor:
If only we'd used `var let = ...` instead, which is totally valid in non-strict mode.

samthor (approved):

PR #1432, opened 2021-09-16
Add chrome.action API to the customize extension user interface table

### LINKED ISSUE #1006: MV3 action missing in “Extension Development overview”
**Describe the bug**
- **“Customize extension user interface”** is missing `action` and only includes browser actions and Page actions.
- Would be nice to clearly identify which APIs are MV2 like on [API reference](https://developer.chrome.com/docs/extensions/reference/)

**To Reproduce**
1. Go to 'https://developer.chrome.com/docs/extensions/mv3/devguide/ '

**Expected behavior**
To see `action` in the “Customize extension user interface” table.

**Additional context**
Discovered bug when working on #983

tannerdolby:
Fixes #1006

- Add `action` to the "Customize extension user interface" table on https://developer.chrome.com/docs/extensions/mv3/devguide/
- Todo: Would be nice to clearly identify which APIs are MV2 like on API reference

The wording I used for the `action` description might need to be tweaked but hopefully this is a good start. As for identifying which APIs are MV2 etc, that can be added in this PR as well. Would it be preferred to keep the same styling as on [API reference](https://developer.chrome.com/docs/extensions/reference/), e.g.`<div class="reference-requirements"><span>MV3+</span></div>` representing the availability?

cc @AmySteam

### DIFF
--- a/site/en/docs/extensions/mv3/devguide/index.md
+++ b/site/en/docs/extensions/mv3/devguide/index.md
@@ -16,6 +16,10 @@ functionality.
       <th colspan="2"><strong>Customize extension user interface</strong></th>
     </tr>
     <tr>
+      <td><a href="/docs/extensions/reference/action">Action</a></td>
+      <td>Control the display of an extension's icon in the toolbar.</td>
+    </tr>
+    <tr>
       <td><a href="/docs/extensions/reference/browserAction">Browser&nbsp;Actions</a></td>
       <td>Add an icon, tooltip, badge, and popup to the toolbar.</td>
     </tr>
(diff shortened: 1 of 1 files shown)


AmySteam:
@tannerdolby Do you think we should alphabetize this list? Happy to see progress on this issue! 🙂

tannerdolby:
@AmySteam Yeah, I think having the list alphabetized would be a good idea!

samthor:
This is specifically a MV3 page; I'd like to in fact only include MV3 things here, so remove `browserAction`.

We could specifically note this at the top. I'm happy for this to be submitted for now though, but @tannerdolby you're welcome to submit a cleanup PR for the above issue generally.

samthor (approved):

tannerdolby:
Ok will do, and thanks for the review @samthor !

I removed `browserAction` in #1448 as indeed its MV2 on an MV3 only page. That PR also addresses adding a note that the stuff on this page is all MV3.

samthor (approved):

PR #1450, opened 2021-09-17
Add 'Label' shortcode to render a filename for code snippets

### LINKED ISSUE #915: Provide a way to set a filename for a code block
### Is your feature request related to a problem? Please describe.
At the moment there isn't a good way for documentation editors to tag a code sample with a filename. 

### Describe the solution you'd like
I'd like to see a shortcode that we could use to tag a code block with a filename. I've included a rough sketch of how this might work below.

The source markdown + nunjucks
````njk
{% codeFile name="manifest.json" %}
```json
{ ... }
```
{% endCodeFile %}
````

The rendered output (code formatting on the code sample excluded)
```html
<figure>
<figcaption><strong>manifest.json</strong></figcaption>
<pre>{ ... }</pre>
```

### Describe alternatives you've considered
I've experimented with a couple of approaches, including using a bolded paragraph above the code sample and adding a [special comment](https://developer.chrome.com/docs/extensions/reference/runtime/#example-content-msg) at the top of the code block to denote the filename. Neither reads well or feels good in the final output.

### Additional context
@robdodson noted that the [label component](https://web.dev/handbook/web-dev-components/#labels) on web.dev fills this role.

tannerdolby:
Fixes #915

- Defines `Label` function in the `./site/_shortcodes/` directory (Maybe we don't include the parent `<div class="type--caption">` and just use `<p class="code-label">` like web.dev does?)
- Registers `Label` as a paired shortcode (for consistency since web.dev does) in the eleventy config file
- Todo: Add styles to ensure margins look good between filename and code snippet e.g. `.code-label + pre { margin-top: get-size(200) }` where get-size(200) will be tinkered with until it looks good.

[Comments in original issue](https://github.com/GoogleChrome/developer.chrome.com/issues/915#issuecomment-868885621) by @robdodson

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -24,6 +24,7 @@ const {IFrame} = require('./site/_shortcodes/IFrame');
 const {Glitch} = require('./site/_shortcodes/Glitch');
 const {Hreflang} = require('./site/_shortcodes/Hreflang');
 const {Img} = require('./site/_shortcodes/Img');
+const {Label} = require('./site/_shortcodes/Label');
 const {Video} = require('./site/_shortcodes/Video');
 const {YouTube} = require('./site/_shortcodes/YouTube');
 const {Columns, Column} = require('./site/_shortcodes/Columns');
@@ -128,6 +129,7 @@ module.exports = eleventyConfig => {
   eleventyConfig.addPairedShortcode('Compare', Compare);
   eleventyConfig.addPairedShortcode('CompareCaption', CompareCaption);
   eleventyConfig.addPairedShortcode('Aside', Aside);
+  eleventyConfig.addPairedShortcode('Label', Label);
   eleventyConfig.addShortcode('LanguageList', LanguageList);
 
   // Add transforms
--- a/site/_shortcodes/Label.js
+++ b/site/_shortcodes/Label.js
@@ -1,0 +1,8 @@
+/** Renders a filename associated with a code snippet.
+ * @param {string} filename A filename representing the code snippet
+ */
+function Label(filename) {
+  return `<div class="type--caption"><p class="code-label">${filename}</p></div>`;
+}
+
+module.exports = {Label};


tannerdolby:
Some good candidates to make use of the `{% Label %}` paired shortcode would definitely be code snippets like:
https://developer.chrome.com/docs/extensions/reference/runtime/#example-content-msg

Screen Shot 2021-09-17 at 4 22 43 PM

then we could remove those comments inside the snippet that currently represent the filename.

samthor:
Post a screenshot of what this looks like in practice?

In web.dev, we actually run this through the `md` filter because people want to use non-filenames at times.

tannerdolby:
Template Usage:

Use the `Label` paired shortcode to generate the HTML representing a filename for an associated code snippet.

```md
{% Label 'content.js' %}{% endLabel %}

```js
{ // Block used to avoid setting global variables
  const img = document.createElement('img');
  img.src = chrome.runtime.getURL('logo.png');
  document.body.append(img);
}
```

which would generate:

```
<div class="type--caption"><p class="code-label">content.js</p></div>
...
```

Screenshot (completely unstyled other than reducing the top margin on the code snippet for the filename):

Screen Shot 2021-09-20 at 11 16 09 PM

The styling could be updated to match the same sort of styling found on web.dev - https://web.dev/handbook/web-dev-components/#labels

samthor:
I'm a bit confused about how Markdown is being run here but we should match web.dev, which accepts the thing being lablled in between the start and end shortcode.

tannerdolby:
> I'm a bit confused about how Markdown is being run here but we should match web.dev,

Shortcodes can be run inside a Markdown file in 11ty (with configuration set in frontmatter) so in the example above I'm just using the `Label` paired shortcode inside the `runtime/index.md` template.

tannerdolby:
> which accepts the thing being lablled in between the start and end shortcode.

That was a typo on my part, its late and thats the syntax for a 'regular' shortcode `{% foo 'bar' %}` where paired shortcodes use this syntax with the content or template variables in between: `{% foo %}bar{% endfoo %}

tannerdolby:
In conclusion, the generated filename HTML still needs styling and to be tested a bit. 

It might not be feasible to add `templateEngineOverride` to every markdown file if we want to use the `{% Label %}` paired shortcode in, but for now that's the only way forward in order to execute a shortcode inside a Markdown file.

tannerdolby:
I plan to do a bit of "test driving" of this sometime this week/weekend. Then I can provide some screenshots/examples to better illustrate how this would work in practice.

jeffposnick:
Hello @tannerdolby—just wanted to check to see if this was still something you wanted to actively pursue. I don't have the full context from the original discussion, as there's been some turnover on the team since it was originally proposed.

tannerdolby:
Hello @jeffposnick! Yeah, I'm still interested in pursuing this PR. Ok, I will pickup from where I left off and try providing you with a small summary to fill in the gaps from original discussion.

tannerdolby:
A good page to see the `{% Label %}` paired shortcode in action: http://localhost:8080/docs/extensions/reference/runtime/

**Before:**
Screen Shot 2022-03-10 at 9 11 10 PM

**After:**
Screen Shot 2022-03-10 at 9 10 40 PM

Notes:
Instead of having to add `templateEngineOverride: njk, md` to every `.md` template for using the `{% Label %}` shortcode, we can instead specify `templateFormats: ["md", "liquid", "njk"]` in the .eleventy.js config file to achieve the same behavior of using `{% Label %}` in Markdown files.

For the styling, I just kept it similar to what web.dev is doing by making `--flow-space: 0.5rem` to ensure that code snippets with labels (i.e. code snippets that are adjacent with `.type--caption`'s) have appropriate top margins.

jeffposnick (approved):
Given that this mostly matches what's implemented in web.dev, as per Rob's original feedback, this seems fine to me.

I'll leave things open for a few days in case @devnook or @rachelandrew have any feedback before merging.

tannerdolby:
@jeffposnick Ok, sounds good. Thanks for the review.

AmySteam:
@tannerdolby Thanks for putting this together! 👏  I'm going to be using this a lot as I audit the extension docs.

@jeffposnick Thanks for following up on this, it will improve the extension code samples significantly! 👍 

@rachelandrew @devnook I'm thinking of merging this PR, any thoughts before I do so? Thanks! :)

tannerdolby:
@AmySteam No problem! I'm glad to see that this shortcode will be useful for the extension docs! 👍

I recently had a [feature request](https://github.com/GoogleChrome/web.dev/pull/7455) in the web.dev repo that was idle for quite awhile like this one and it broke the build when merged as the `$_presubmit` label wasn't added to catch any existing errors before merging. Not sure if that same workflow applies here, but I thought it might be of some value to mention.

rachelandrew:
@tannerdolby the tests all run here by default I (we don't have the presubmit step here). However, I'd like @devnook to give an ok to this before merging it.

AmySteam:
@devnook Thank you for merging this! :)

PR #1479, opened 2021-09-21
Updated doc to mention https requirement

mikecann:
I discovered this requirement the hard way, I think this document is misleading in that it isnt mentioned.


mikecann:
@googlebot I signed it!

samthor:
Thanks, this is a useful improvement.

@andreban I'm going to submit this but perhaps you want to fix up the writing style later.

PR #1480, opened 2021-09-21
allow more varied release data from Chrome

samthor:
We now just assume the existence of 'stable'. If stable === beta, then pretend beta is stable+1.

Also is more permissive on the home page, if releases are missing don't render them.


Catclaws12:
- []()

PR #1454, opened 2021-09-18
Fix broken links on '/mv3/tut_oauth/' page

tannerdolby:
Fixes #964 

Changes proposed in this pull request:

- Changes 'https://developers.google.com/people/v1/read-people' to 'https://developers.google.com/people/api/rest/v1/contactGroups/get'
- Todo: Find correct URLs to sample files within `GoogleChrome/samples` repo - download links for sample source files (should go to samples repo)

Broken sample file links (need to find matching file for these in `GoogleChrome/samples`):
- https://developer.chrome.com/docs/extensions/mv3/tut_oauth/examples/tutorials/oauth_starter/oauth.js
- https://developer.chrome.com/docs/extensions/mv3/tut_oauth/examples/tutorials/oauth_starter/index.html

I had a look at the [`samples`](https://github.com/GoogleChrome/samples) repository and couldn't find the files by the names used in their original links, I'm maybe thinking the filenames changed or maybe their location changed within `samples` (if they even exist)? @samthor

PR #1485, opened 2021-09-22
Update manifest and page to be MV3 on '/mv3/tut_oauth/'

tannerdolby:
Related to #1454

- Update manifest fields from MV2 to MV3 in code snippet
- Use 'background service worker' instead of 'background script' and `action` instead of `browserAction`


samthor:
This is great. Could you also remove the links to any of the downloadable files, since they don't work right now?

Even comment them out, and we can get back to them later cc @dotproto

tannerdolby:
Yeah sure thing, I will remove the links to all of the downloadable files since they aren't working right now until we find/add the correct source in `GoogleChrome/samples` repo.

samthor (approved):

PR #1496, opened 2021-09-23
MV2 support/sunset timeline and its introduction page

awfuchs:
Adds a blog post announcing the MV2 support timeline and transition to MV3, as well as a page with more detailed description of the dates and their impact.


dotproto (approved):
LGTM :shipit:

PR #1508, opened 2021-09-25
Fix a key name

gingerhot:

- The key `extensions` should be `extension_ids`.


samthor:
Thanks for this PR but I've made a similar and extended change in #1522.

PR #1520, opened 2021-09-29
Updated FLEDGE flag information

samdutton:


site/en/docs/privacy-sandbox/fledge/index.md
```diff
@@ -6,17 +6,17 @@ subhead: >
...
 date: 2021-05-18
-updated: 2021-09-20
+updated: 2021-09-29
 authors:
   - samdutton
 ---
 
 {% Aside %}
-FLEDGE can be tested in Chrome from versions 91.0.4472.49 or 92.0.4493.0 with the following flags enabled:
+FLEDGE can be tested in current versions of Chrome and Chrome Canary with the following flags enabled:
```
mihajlija, line 15: What's the reasoning for this change? Having the version numbers is more specific and future proof than this phrasing.

samdutton: I realised that 'in' implied _only_ those versions, rather than _from_ those versions, and since these versions are now well in the past, I think 'current versions' is OK. 

If FLEDGE can no longer be tested (or something else make the information here obsolete) we'll need to change this anyway.

Hope that makes sense!

mihajlija (approved):

PR #1539, opened 2021-10-03
Fix typo in commands API usage on '/extensions/reference/commands/'

### LINKED ISSUE #1537: Incorrect block quote about commands.getAll() on /reference/commands/
**Describe the bug**

This block quote on https://developer.chrome.com/docs/extensions/reference/commands/ is incorrect:

> _execute_action, _execute_browser_action, and _execute_page_action will not appear in the list of commands returned by command.getAll().

**Expected behavior**

I don't know about the new `_execute_action`, but `_execute_browser_action` is definitely returned in the list of commands.

Also note that it says `command.getAll()` rather than `commands.getAll()`.

**Screenshots**

Here's the console output for `commands.getAll()`:

image

 - OS: Windows 10
 - Browser Chrome
 - Version 92

tannerdolby:
Fixes #1537

- Changes `command.getAll()` to `commands.getAll()` to correct typo in commands API usage
- Todo: Fix language in blockquote to remove `_execute_browser_action` since it does appear in the list returned by `commands.getAll()`

### DIFF
--- a/site/en/docs/extensions/reference/commands/index.md
+++ b/site/en/docs/extensions/reference/commands/index.md
@@ -274,7 +274,7 @@ experience by anticipating this possibility and checking for collisions at insta
 {% Aside %}
 
 `_execute_action`, `_execute_browser_action`, and `_execute_page_action` will not appear in the list
-of commands returned by `command.getAll()`.
+of commands returned by `commands.getAll()`.
 
 {% endAside %}
 


tannerdolby:
@samthor Should I remove `_execute_browser_action` from that blockquote in a separate cleanup PR or is it ok? I left that as a todo here since I wasn't entirely sure if it needed to be changed.

fwextensions:
FWIW, someone should check that _execute_page_action is also returned. I don't have an extension that uses it, but I'd assume it's returned. 

Maybe _execute_action is different, as I think it's new in manifest V3 (though I'd hope not, as my extension relies on knowing the modifier keys used to open the pop up). 

Anyway, thanks for fixing the commands typo.

tannerdolby:
> FWIW, someone should check that _execute_page_action is also returned. I don't have an extension that uses it, but I'd assume it's returned.

cc @AmySteam are you familiar with what `commands.getAll()` returns to clarify this?

PR #1540, opened 2021-10-03
Adds table of contents to blog posts

### LINKED ISSUE #1521: Add table of contents to blog posts
The ToC appears on docs ([example](https://developer.chrome.com/docs/devtools/overview/)) but not blog posts ([example](https://developer.chrome.com/blog/new-in-devtools-95/)). We should add it to blog posts since those can sometimes be quite long.

tannerdolby:
Fixes #1521 

- Adds ToC to the base blog posts template `blog-post.njk` (Similar to how `doc-post.njk` generates a ToC)

### DIFF
--- a/site/_includes/layouts/blog-post.njk
+++ b/site/_includes/layouts/blog-post.njk
@@ -44,4 +44,6 @@
     {% include 'partials/tags.njk' %}
     {% include 'partials/updated.njk' %}
   </article>
+  
+  {% include 'partials/toc-side.njk' %} 
 {% endblock %}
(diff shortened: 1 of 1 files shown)


samthor:
If you look at doc-post you'll see that we actually render the ToC twice, one for responsive and one for not. Can you mirror that?

tannerdolby:
Ah, yes after looking back at `doc-post.njk` I see how I missed the second ToC rendering. I will update that now

samthor (approved):

PR #1545, opened 2021-10-05
implement "up-to-date" build design

samthor:
This is a major PR changing the way we pull in external data to the site. It mostly implements the internal named design doc.

- We have a subfolder named external/ which itself contains data/ (dynamically fetched), build/ (run to build data/) and fallback/ (provides consistent data for CI).
- The Cloud Build task will be run every ~hour, and populates a Cloud Storage bucket with external data on success
- Normal users of the site, as part of `npm run dev`, will have this folder synchronized to external/data/
- If you want to instead test external data or change its build, you can run `npm run build-external`; this will also prevent sync until you later run `npm run sync-external` to clear a boolean

The idea is that external data breakages, while important (and will alert the team, because we alert on Cloud Build failures of any type), shouldn't stop regular deploy for our authors. They'll always get a "last known good" during local dev, or as part of our regular site deploy.

### DIFF
--- a/external/build-external.js
+++ b/external/build-external.js
@@ -1,0 +1,70 @@
+/**
+ * @fileoverview Builds all external sources. This clears the data/ folder here
+ * and allows all scripts in build/ to write new files there.
+ *
+ * This is intended for use by Cloud Build, or by site devs doing local work.
+ */
+
+// nb. This is important; we load dotenv here and pass environment down.
+// It's safe to call this inside the build scripts too, but we run them with
+// cwd=data/, so it can't find the env.
+require('dotenv').config();
+
+const fs = require('fs');
+const path = require('path');
+const glob = require('glob');
+const childProcess = require('child_process');
+const crypto = require('crypto');
+
+async function run() {
+  let errors = 0;
+
+  const scripts = glob.sync('build/*.js', {cwd: __dirname});
+  scripts.sort(); // run in alphabetical order
+
+  const dataTarget = path.join(__dirname, 'data');
+  fs.rmSync(dataTarget, {recursive: true});
+  fs.mkdirSync(dataTarget, {recursive: true});
+
+  /** @type {childProcess.CommonExecOptions} */
+  const options = {cwd: dataTarget};
+
+  for (const script of scripts) {
+    const r = path.join(__dirname, script);
+    console.info('> Running', r);
+    try {
+      childProcess.execFileSync('node', [r], options);
+    } catch (e) {
+      // We don't log the error here, as we're already getting STDERR piped above.
+      console.warn(`! Failed to execute "${script}" (${e.status})`);
+      ++errors;
+    }
+  }
+
+  // Determine the hash for everything in data/.
+  const h = crypto.createHash('sha256');
+  const allFiles = glob.sync('data/**/*', {cwd: __dirname});
+  if (!allFiles.length) {
+    throw new Error('no files generated, cowardly refusing to hash');
+  }
+
+  // Sort allFiles, in case glob.sync is inconsistent.
+  allFiles.sort();
+
+  for (const f of allFiles) {
+    const p = path.join(__dirname, f);
+    const bytes = fs.readFileSync(p);
+    h.update(bytes);
+  }
+  const digest = h.digest('hex');
+  console.info(`@ Generated digest=${digest} for ${allFiles.length} files`);
+  fs.writeFileSync(path.join(__dirname, 'data/.hash'), digest);
+
+  // If there were any errors, return with a non-zero status code anyway.
+  if (errors) {
+    // eslint-disable-next-line no-process-exit
+    process.exit(1);
+  }
+}
+
+run();
--- a/external/build-external.js
+++ b/external/build-external.js
@@ -27,7 +27,7 @@ async function run() {
   fs.mkdirSync(dataTarget, {recursive: true});
 
   /** @type {childProcess.CommonExecOptions} */
-  const options = {cwd: dataTarget};
+  const options = {cwd: dataTarget, stdio: 'inherit'};
 
   for (const script of scripts) {
     const r = path.join(__dirname, script);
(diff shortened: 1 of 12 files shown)


external/build-external.js
```diff
@@ -0,0 +1,76 @@
+/**
+ * @fileoverview Builds all external sources. This clears the data/ folder here
+ * and allows all scripts in build/ to write new files there.
+ *
+ * This is intended for use by Cloud Build, or by site devs doing local work.
+ */
+
+// nb. This is important; we load dotenv here and pass environment down.
```
devnook, line 8: This is too cryptic for me to comprehend, Sam. Can you expand the comment?

external/maybe-sync-external.js
```diff
@@ -0,0 +1,41 @@
+/**
+ * @fileoverview
```
devnook, line 2: Please fill in, esp explain when "maybe-sync" means "don't sync"

external/maybe-sync-external.js
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+const fs = require('fs');
+const path = require('path');
+const childProcess = require('child_process');
+
+// Synchronize remote data at most once every 12 hours.
+const syncThresholdMs = 12 * 60 * 60 * 1000;
```
devnook, line 10: I wonder if it wouldn't be possible to use eleventy-cached-data plugin here instead of custom cashing.

external/build-external.js
```diff
@@ -0,0 +1,76 @@
...
+      childProcess.execFileSync('node', [r], options);
+    } catch (e) {
+      // We don't log the error here, as we're already getting STDERR piped above.
+      console.warn(`! Failed to execute "${script}" (${e.status})`);
+      ++errors;
+    }
+  }
+
+  // Determine the hash for everything in data/.
+  const h = crypto.createHash('sha256');
```
MichaelSolati, line 45: Would prefer a different name so I know what `h` is later in the code. Same for `r` above. I think the only one letter variables I like are for incrementors (`i`) or `e` for error. Otherwise a more descriptive variable name would be appreciated.

external/build-external.js
```diff
@@ -0,0 +1,76 @@
...
+// cwd=data/, so it can't find the env.
+require('dotenv').config();
+
+const fs = require('fs');
+const path = require('path');
+const glob = require('glob');
+const childProcess = require('child_process');
+const crypto = require('crypto');
+
+async function run() {
```
MichaelSolati, line 19: I don't think I see any await, couldn't this just be script.... Like, nothing in the run function, i.e. no run function?

external/build-external.js
MichaelSolati: Is there a reason why individual build scripts can't import `.env`, like if we ever want to run one file?

external/maybe-sync-external.js
MichaelSolati: +1

external/maybe-sync-external.js
MichaelSolati: I think this is to determine weather to write data to the storage bucket.

site/_data/tweets.js
```diff
@@ -134,19 +127,11 @@ const formatEntities = tweet => {
...
-        if (process.env.NODE_ENV === 'production') {
-          return undefined;
-        } else {
-          return require('./tweets-sample.json');
-        }
-      });
+  // We have to cast this to unknown first as Typescript gets very literal about the underlying type
+  // of the JSON.
+  let tweets = /** @type {TwitterTweet[]} */ (
+    /** @type {unknown} */ (require('../../external/data/tweets.json'))
```
MichaelSolati, line 133: Maybe we should use the cache tool here to point to the external tweets if there isn't a local one.

external/build-external.js
samthor: I've removed this behavior completely and instead made the child processes run with a cwd of the project's root, so `dotenv` will just work. I think it's probably clearer.

external/build-external.js
samthor: You're right, but this matches the other files, and it makes it easier if it ever needs to do async work. What do you think?

site/_data/tweets.js
samthor: I have mixed feelings on this but I'm happy to be disagreed with. I think there's a big advantage in _not_ having to have conditionals here—the code always just pulls _something_ in.

But I've added an eslint rule to allow this to be missing and that's awkward too.

WDYT?

samthor:
I've updated the README and made the linter run.

- In CI build mode, we aggressively use the fallback files _first_, in case e.g. Twitter's API key isn't available.

- In CI sync mode, we aggressively use the fallback files _last_, so that Percy and such use the same data every time.

external/maybe-sync-external.js
```diff
@@ -0,0 +1,50 @@
...
+  let mtimeMs = 0;
+  try {
+    const stat = fs.statSync(path.join(__dirname, 'data'));
+    mtimeMs = stat.mtimeMs;
+  } catch (e) {
+    // The folder probably doesn't exist.
+  }
+  const since = +new Date() - mtimeMs;
+  if (since < syncThresholdMs) {
+    // Don't log at all, and don't synchronize.
```
devnook, line 35: Should we consol.log a warning on this occasion?

external/maybe-sync-external.js
```diff
@@ -0,0 +1,50 @@
...
+    return;
+  }
+
+  const out = childProcess.spawnSync('npm run sync-external', {
+    shell: true,
+    stdio: 'inherit',
+  });
+  if (out.status) {
+    // TODO(samthor): If we're offline and have valid data, this could perhaps allow builds to
```
devnook, line 44: I'd say we log a warning that the sync failed and the build is using old data, and use data from /data if present but older than 12h. If no files in /data, then fail.

devnook (approved):
LGTM with small comments left

PR #1556, opened 2021-10-07
always use reference layout

samthor:
Fixes a race condition which caused some pages to often have no layout.

Pages should have a layout. If they don't, throw an error.

### DIFF
--- a/site/en/docs/extensions/reference/reference.11tydata.js
+++ b/site/en/docs/extensions/reference/reference.11tydata.js
@@ -29,11 +29,13 @@ function namespaceForData(data) {
   }
 
   if (api in chromeApiNamespaces) {
+    console.warn('found API :party:', api);
     return chromeApiNamespaces[api];
   }
 
   // This can be called several times by Eleventy. The first time it's called it's unlikely that
   // the namespace data is available yet, so we can't warn here if it's missing.
+  console.warn('failed to find API', api);
   return undefined;
 }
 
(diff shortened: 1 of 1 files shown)


dotproto (commented):
It looks like the previous implementation could end up returning a data.layout value of undefined, which I assume caused the missing layout bug we're trying to fix. Instead, we will now noisily error (throw on line 72) if such a condition is encountered.

Do we understand why data.layout would sometimes be undefined?
Can layout it be computed multiple times as the note in namespaceForData implies?
Does is the Error thrown on line 72 recoverable, or does this fail the build?
If the answer to 3 is "fail the build", do we retry it or does that deploy job fail and log an error?

samthor:
> * Do we understand why `data.layout` would sometimes be `undefined`?

I actually think sometimes it was the blank string, which won out over `undefined`.

> * Can `layout` it be computed multiple times as the note in `namespaceForData` implies?

Yes. 11ty, as far as I can tell, will reevaluate `undefined` variables if others change. Basically it means—we're not sure yet! Call me later! Resolving multiple variables against each other is hard!

> * Does is the Error thrown on line 72 recoverable, or does this fail the build?

Fails the build.

> * If the answer to 3 is "fail the build", do we retry it or does that deploy job fail and log an error?

No, it'll fail. It should fail; pages in this folder should all be trying to display a specific `api:`, which is codified in the front-matter. If they're not then we should change this rule. I've added a comment saying this.

dotproto (approved):
Thanks for the clarifications. Failing the build if we produce unstyled or empty pages is definitely an improvement over publishing them, so LGTM.

PR #1563, opened 2021-10-08
Fix document error.

ReedSun:
- There are two duplicate 'action' properties in  [Manifest file format](https://developer.chrome.com/docs/extensions/mv3/manifest/), I deleted the one in optional properties.
- Fix  some link errors in [chrome.i18n](https://developer.chrome.com/docs/extensions/reference/i18n/).


samthor:
Thank you.

PR #1624, opened 2021-10-26
Add android:exported=true to TWA integration guide

devbridie:
See [Android doc on Behavior Changes in Android 12](https://developer.android.com/about/versions/12/behavior-changes-12#exported). A value for `android:exported` is required when targeting Android 12; true is required for the `LAUNCHER` category.

### DIFF
--- a/site/en/docs/android/trusted-web-activity/integration-guide/index.md
+++ b/site/en/docs/android/trusted-web-activity/integration-guide/index.md
@@ -110,6 +110,7 @@ Add the Trusted Web Activity by inserting an `activity` tag into the `applicatio
 
     <application
         android:allowBackup="true"
+        android:exported="true"
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
         android:roundIcon="@mipmap/ic_launcher_round"


AndreasBoehm (approved):

site/en/docs/android/trusted-web-activity/integration-guide/index.md
```diff
@@ -110,6 +110,7 @@ Add the Trusted Web Activity by inserting an `activity` tag into the `applicatio
 
     <application
         android:allowBackup="true"
+        android:exported="true"
```
andreban, line 113: Thanks for sending the PR.

My understanding from the docs is that `android:exported="true"` should be added to the `activity` tag, not the `application` tag (as in the example [here](https://github.com/GoogleChromeLabs/bubblewrap/blob/main/packages/core/template_project/app/src/main/AndroidManifest.xml#L104)). Would you mind moving it to the `activity` tag?

andreban (changes requested):

rachelandrew:
Closing as OP did not return to address comments.

PR #1653, opened 2021-10-31
render workbox types

samthor:
Renders SW-only Workbox types at latest. (Workbox also wants to render their old types; this is a start. We're also missing things like version numbers.)

This re-uses the code we use to parse Chrome types, and has a bit of refactoring/moving code to support that.

There's some minor issues with rendering which are not worth fixing now.

Preview will be under https://deploy-preview-1653--developer-chrome-com.netlify.app/docs/workbox/reference/.


jeffposnick:
This is really encouraging to see! Thanks, @samthor!

Something that stood out on initial investigation was that the cross-class links we have set up that worked with JSDoc don't render correctly here. E.g. `registerRoute()` on [this page](https://deploy-preview-1653--developer-chrome-com.netlify.app/docs/workbox/reference/workbox-routing#method-registerRoute):

Screen Shot 2021-11-01 at 6 59 29 PM

We can adjust how those links are written in the source code if needed; let us know what format would be easiest to ingest.

samthor:
Yeah, the `module:` syntax is unexpected, but easily parseable: we generate
top-level "module" objects that are rendered, and e.g., one of them _is
called_ "workbox-routing".

I can try to fix this today before we check it in. There's also a weird way
y'all do links that I can try to match too.

One thing I've noticed is that you often document objects _where you use
them_, and not on the underlying type. E.g., annotations like ***@***.***
{number} [paramName.foo] blah`, but this actually points to an interface
called something like "RouteFoo" that is somewhere else.
This we'll probably never support - we see the reference to that other type
and just say, go over there to understand it.

jeffposnick (approved):
The output meets the bar for an initial checkin, with future tweaking to update the source comments, as well as in the parsing, to come.

PR #1664, opened 2021-11-02
don't render comment again for top-level namespace

samthor:
🤦

### DIFF
--- a/site/_includes/macros/reference.njk
+++ b/site/_includes/macros/reference.njk
@@ -27,7 +27,6 @@
         </td>
         <td class="reference-collapse type">
           {{ macros.renderAvailability(namespace.root, 'pad-bottom-300') }}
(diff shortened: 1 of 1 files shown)


dotproto (approved):
This seems straight forward.

Quick clarifying question, though; is the duplicate content coming from `{{ macros.renderCommentText(namespace.root) }}` and `{{ namespace.root.comment.shortText  | md | safe }}` or something else?

samthor:
That's right. So every node in TypeDoc has a 'shortText' comment and a 'text' comment (if there's more than one line).

For better or worse, `renderCommentText` takes those and adds things like relative links (e.g. with `@​link`). But since we're not a peer of other APIs, those links instead can go "up" a folder, which isn't correct.

For that reason I figure, just do the basic rendering and don't try to fix those.

Issue #1655, opened 2021-11-01
Site builds every hour, only needs to on external data or git change


samthor:
We currently build and deploy every hour, even if the git hash or the hash or the generated external data changes.

The git hash is obviously part of the repo, and the external hash file is at `gs://external-dcc-data/.hash` ([public URL here](https://storage.googleapis.com/external-dcc-data/.hash)).

web.dev takes an approach to run a script called "version-check" which _kicks off_ a build if these things have changed.

PR #1666, opened 2021-11-02
add version-check task which can kick off build

samthor:
Next step of #1655.

This is the final code change, the rest is config in our Pantheon project.


dotproto (approved):
:shipit:

PR #1681, opened 2021-11-04
Fix external data changing every time

samthor:
Because we used a temp folder to parse with TypeDoc, it got included in the output, and therefore the hash. Clear the filename for now.

We've been "checking" that the site is up-to-date, but having it need an update every time because of this.

### DIFF
--- a/external/build/lib/dts-parse.js
+++ b/external/build/lib/dts-parse.js
@@ -327,6 +327,9 @@ class Transform {
     if (!this.filter(node, parent, namespace)) {
       return null;
     }
+    for (const source of node.sources ?? []) {
+      source.fileName = ''; // HACK
+    }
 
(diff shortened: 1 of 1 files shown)


external/build/lib/dts-parse.js
```diff
@@ -327,6 +327,9 @@ class Transform {
     if (!this.filter(node, parent, namespace)) {
       return null;
     }
+    for (const source of node.sources ?? []) {
+      source.fileName = ''; // HACK
```
devnook, line 331: Please add a cause for this hack, ie that we want to avoid duplicate builds

devnook (approved):

external/build/lib/dts-parse.js
samthor: This was actually my 1st attempt that was superceded by the code below. I've removed this and updated that comment even further.

PR #1746, opened 2021-11-18
Add Auto Dark opt-out for the Chrome logo 

andreban:
- Adds an auto dark per-element opt-out for the Chrome logo.
- See https://crbug.com/1271390

### DIFF
--- a/site/_scss/blocks/_top-nav.scss
+++ b/site/_scss/blocks/_top-nav.scss
@@ -16,10 +16,11 @@ top-nav {
     }
   }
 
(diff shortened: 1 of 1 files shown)


chybie (approved):

jeffposnick (approved):
Feel free to merge this if it's still relevant.

andreban:
A lot is changing in the implementation on the Chrome side to fix this, so hopefully this won't be necessary. I think we can drop this and re-evaluate in the future, when those improvements land.

PR #1773, opened 2021-11-24
Document script.executeScript() promise behavior

rdcronin:
If the result of code injected with scripting.executeScript() is a
promise, Chrome will wait for the promise to resovle and return the
resulting value. Add documentation about this to the chrome.scripting
API reference page and the "What's New" page.


dotproto (approved):
Looks great. Thanks for adding this, @rdcronin!

PR #1776, opened 2021-11-24
Don't auto-include jpmedley in extensions PRs

dotproto:
The workflow we're currently using for the Chrome Web Store CUJ project has us including @jpmedley only after other reviewers have finished providing technical feedback. While works as expected most of the time, there are some scenarios where this does not work as expected. Specifically, scenarios where a PR is created but we are not yet ready for @jpmedley to be tagged in. It seems that PullApprove will not allow us to remove reviews that are not prefixed with a tilde (`~`) from a PR. As a result, every time we try to remove @jpmedley before the PR is ready, he is automatically re-added.

In order to prevent @jpmedley from being tagged into a review too early, this PR reworks how reviewers are added to Extensions/CWS/Apps doc changes. Rather than automatically add all possible reviewerse, we now only automatically add the `required` number of reviewers (1) in the order specified in this config. This effectively means that @dotproto will be added to all extensions PRs, but @jpmedley will only be automatically added when @dotproto creates a PR.

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -13,14 +13,22 @@ pullapprove_conditions:
 
 groups:
   # Extensions
-  # Handles any changes related to extensions, web store, and chrome apps.
-  # By default Simeon and Andy will be tagged in as reviewers, but anyone
-  # on the dcc-extensions, dcc-content, or dcc-approvers team can approve PRs.
+  #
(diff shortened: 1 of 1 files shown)


MichaelSolati (approved):

jpmedley:
👏

PR #1788, opened 2021-11-29
Update description to match behavior.

solomonkinard:


jpmedley:
@dotproto Is this change correct?

dotproto (approved):
@jpmedley yep, this is technically accurate. @solomonkinard and I discussed confusion around this language this last week. I believe @solomonkinard also created some additional Chromium tests to validate expected behavior and submitted this PR to close the loop.

jpmedley (approved):

PR #1811, opened 2021-12-04
Fix a typo in CSS-in-JS support in DevTools post

nvh95:
Fix a typo in https://developer.chrome.com/blog/css-in-js/


chybie (approved):
LGTM thanks!

PR #1830, opened 2021-12-08
[devtools-blog] add full a11y tree blog post

johanbay:


site/en/blog/full-a11y-tree/index.md
```diff
@@ -0,0 +1,167 @@
...
+tags:
+  - devtools-engineering
+  - devtools
+---
+
+{% include 'partials/devtools/en/banner.md' %}
+
+Chrome DevTools is launching a full accessibility tree making it easier for developers to get an overview of the whole tree. In this post find out how this tree is created and how to use it in your work.
+
+{% Img src="image/gGNVLl64MBQvhaYBXAAuRiOs3r92/uuDsB0fUMzarbKmRi5Uz.gif", alt="animated screenshot of new accessibility tree", width="800", height="519" %}
```
chybie, line 20: Suggest to upload video instead as it's smaller and better in performance. It will look like gif.

```
{% Video src="video/some-link.mp4", autoplay="true", muted="false", loop="true",  class="screenshot" %}
```

site/en/blog/full-a11y-tree/index.md
```diff
@@ -0,0 +1,167 @@
...
+
+{% Img src="image/gGNVLl64MBQvhaYBXAAuRiOs3r92/fpzVAuTVolMbVjiLMt0A.png", alt="screenshot of new accessibility tree for large page", width="800", height="488" %}
+
+
+### Live
+
+The new tree view is live and dynamically updates if the accessibility tree changes in the renderer. It hooks into the same mechanics that would notify assistive technology of changes to the tree, and uses that to emit events to the DevTools frontend with updated nodes.
+In practice, the CDP backend listens for updates to the tree, keeps track of which nodes have been requested before, and sends events to the DevTools frontend if any of these nodes changes.
+
+{% Img src="image/gGNVLl64MBQvhaYBXAAuRiOs3r92/ygHxJsgWCYOjZStrdGAj.gif", alt="animated screenshot of new tree that live updates", width="800", height="519" %}
```
chybie, line 115: Same gos to this one.

chybie (approved):
LGTM with comments

rachelandrew (commented):
A few fixes to links and alt text, please see the notes about changing gifs to video. Otherwise this is almost there and we can get it ready to go out on Monday.

I'm also wondering if we should change the folder name (which becomes the URL slug) to `full-accessibility-tree`, in the case of someone typing the URL out for some reason it may not be obvious those are 1s if you don't know.

site/en/blog/full-a11y-tree/index.md
johanbay:
```suggestion
{% Video src="video/gGNVLl64MBQvhaYBXAAuRiOs3r92/XQNBCFrPG8YqNSOloDE8.mp4", autoplay="true", muted="false", loop="true", class="screenshot" %}
```

site/en/blog/full-a11y-tree/index.md
johanbay:
```suggestion
{% Video src="video/gGNVLl64MBQvhaYBXAAuRiOs3r92/vLXJU0FMF41pqKqVJ61D.mp4", autoplay="true", muted="false", loop="true", class="screenshot" %}
```

devnook (approved):

rachelandrew (approved):

site/en/blog/full-accessibility-tree/index.md
```diff
@@ -0,0 +1,167 @@
...
+
+{% Img src="image/gGNVLl64MBQvhaYBXAAuRiOs3r92/DqQ4L3DRo6HTg8RIJnwO.png", alt="Diagram showing how we prune the tree.", width="800", height="450" %}
+
+On the frontend, we construct the full tree including ignored nodes and only prune them just before rendering the nodes. We do this for two reasons:
+
+1. **It makes it much simpler to handle node updates from the backend**, since we have the same tree structure on both endpoints. For example, if the node B is removed in the example, we would receive an update for node X (since its children have changed), but if we had pruned that node we would struggle to figure out what to update.
+2. **It ensures that all DOM nodes have a corresponding accessibility node.** When the tree is toggled, we select the node corresponding to the node currently selected in the DOM tree. So for the previous example, if the user toggles the tree while the DOM node corresponding to X is selected, we inject X between the nodes A and B and select X in the tree. This allows the user to inspect the accessibility node for all DOM nodes and help determine why the node is ignored.
```
rachelandrew, line 147:
```suggestion
- **It makes it much simpler to handle node updates from the backend**, since we have the same tree structure on both endpoints. For example, if the node B is removed in the example, we would receive an update for node X (since its children have changed), but if we had pruned that node we would struggle to figure out what to update.
- **It ensures that all DOM nodes have a corresponding accessibility node.** When the tree is toggled, we select the node corresponding to the node currently selected in the DOM tree. So for the previous example, if the user toggles the tree while the DOM node corresponding to X is selected, we inject X between the nodes A and B and select X in the tree. This allows the user to inspect the accessibility node for all DOM nodes and help determine why the node is ignored.
```

PR #1892, opened 2022-01-01
Add link to the permissions warnings source code

avm99963:
As mentioned in the doc, the permissions table "is updated on a
best-effort basis". This CL adds a link to the piece of code where the
warnings are actually defined so developers have another way to see the
most up to date warnings.

This CL also fixes a broken link in the same doc.


dotproto (approved):
Thanks for the PR!

PR #1901, opened 2022-01-05
Ports URL markdown linter rules from web.dev

malchata:
web.dev has [a number of markdown linter rules](https://github.com/GoogleChrome/web.dev/blob/main/tools/linting/bad-urls.js) that the tech writers team finds useful for eliminating repetitive review tasks. This PR ports those rules from web.dev, which include:

- Ensuring that `developer.mozilla.org` is linked to rather than `wiki.developer.mozilla.org`.
- Warning about locales in MDN links.
- Ensuring that same-site links are relative rather than absolute.


MichaelSolati (approved):
Only speaking for the JS code, this looks good. I'm assuming all of the MD changes are adjustments based on the linter, but I'd rather leave that for the tech writing team to review. I am happy with the JS though.

malchata:
Yup, the MD adjustments are just to satisfy the added linter warnings. I did my best to make sure my search/replace patterns were as limited as possible, but if we need to fix something down the line, we can. In my experience, these changes didn't seem to break anything on web.dev. Mergin'!

PR #1903, opened 2022-01-05
feat(workbox): add workbox-build and workbox-window

MichaelSolati:
Fixes #SOME_ISSUE_NUMBER

- Pull types location from package.json
- Add new workbox packages to build
- Ignore event typing rules for workbox

### DIFF
--- a/external/build/lib/dts-parse.js
+++ b/external/build/lib/dts-parse.js
@@ -348,7 +348,11 @@ class Transform {
     }
 
     if (node.type?.type === 'reference') {
-      if (chromeEventRefTypes.includes(node.type.name)) {
+      // Only upgrade node is not a workbox type
+      if (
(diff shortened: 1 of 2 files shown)


external/build/lib/dts-parse.js
```diff
@@ -348,7 +348,11 @@ class Transform {
     }
 
     if (node.type?.type === 'reference') {
-      if (chromeEventRefTypes.includes(node.type.name)) {
+      // Only upgrade node is not a workbox type
```
philipwalton, line 351: I think think there's a mistake in this comment? s/is/if/

Also it might be worth adding a bit more context as to why the exception for workbox is made here. If someone else came to this code base it wouldn't be clear.

philipwalton (approved):
LGTM. Just one minor comment to help make things clearer.

PR #1904, opened 2022-01-05
Add documentation for "match_origin_as_fallback" in content scripts

rdcronin:
Manifest-specified content script definitions now support the new
property `match_origin_as_fallback`.  This allows them to inject into
URLs with schemes that they wouldn't otherwise match, such as `data:`,
`blob:`, `filesystem:`, and `about:`.

[Preview page](https://deploy-preview-1904--developer-chrome-com.netlify.app/docs/extensions/mv3/content_scripts/)


dotproto (approved):
Thanks for writing this up, @rdcronin.

 I mentioned in conversation that I'd like to tweak the intro to make it more obvious to the reader that this refers to statically declared content scripts, but my efforts to suggest a rewrite haven't gone well. At this point I think it's best to take the PR as is rather than continue to block it for want of a relatively minor improvement.

PR #1921, opened 2022-01-07
feat(workbox): move embedded types to the top level

MichaelSolati:

- If types definition of a workbox library is not located in the root of the package then copy the content of the directory it is in into the top level.


jeffposnick:
LGTM!

PR #1930, opened 2022-01-10
feat(types): get defaut values from `@default` tag

MichaelSolati:

- Add `switch` case to `external/build/lib/dts-parse.js` to store the value of `@default` to `node._feautre`.
- Render `node._feature.default` when rendering single icon type.


jeffposnick (approved):
LGTM!

PR #1942, opened 2022-01-12
fix(types): if node is a getter use getter type for rendering, fixes #1938

### LINKED ISSUE #1938: Getters not returning a type
Getters aren't having their type returned, such as this one => https://developer.chrome.com/docs/workbox/reference/workbox-background-sync/#type-Queue-name

MichaelSolati:
Fixes #1938

- if node is a getter use getter type for rendering
-
-


jeffposnick (approved):
Thanks!

PR #1951, opened 2022-01-13
[Workbox] Add a note to the DOMContentLoaded note

jeffposnick:
This addresses a visibility issue that @snugug ran into with the current docs set.

(I'm just making the change to the d.c.c. version, as I don't want to update the d.g.c. version moving forward.)


Snugug:
ty!

malchata (approved):
Good catch on this. Thanks, @jeffposnick and @Snugug!

PR #1974, opened 2022-01-18
More serious descriptions for Workbox pages

jeffposnick:
R: @malchata
CC: @tropicadri 

This removes some placeholder text, in advance of releasing the new Workbox docs.


malchata (approved):
In my estimation, this new copy is appropriately serious and I approve.

PR #2011, opened 2022-01-24
Add a caveat about using BroadcastUpdatePlugin with workbox-precaching

jeffposnick:
This addresses some developer confusion about using BroadcastUpdatePlugin with `workbox-precaching`.

C.f. https://github.com/GoogleChrome/workbox/issues/2931#issuecomment-1020481215


malchata (approved):
LGTM

PR #2043, opened 2022-01-28
Add keeping a consistent id partial to OAuth tutorial

AmySteam:

- Add a partial for keeping a consistent extension id partial to OAuth2 tutorial.

NOTE: The short IDs match the original anchors, so it shouldn't be a problem for inbound links or bookmarks

Staged links


dotproto (approved):
Nice! Hopefully this makes it easier for users to find consistent instructions while also lowering the maintenance cost ⬆️ 5️⃣

PR #2097, opened 2022-02-05
Fix shorthand format for CSS nested layers

drzamich:

- Fix shorthand format of CSS nested layers on this page: https://developer.chrome.com/blog/cascade-layers/#nested-layers
- Based on https://developer.mozilla.org/en-US/docs/Web/CSS/@layer#nesting_layers


drzamich:
Tagging @una :)

rachelandrew:
@drzamich thanks for this, however you will need to sign the CLA. Details are in the GitHub thread.

PR #2121, opened 2022-02-12
Update index.md

shu6h4m:
Fixes #SOME_ISSUE_NUMBER

-
-
-


dotproto:
I appreciate the clarification you tried to add here. I'm rejecting this PR primarily because I don't think it adds sufficient clarification to be useful. It also has a couple of minor issues that, due to the age of the PR, I don't anticipate that we will get through the iteration cycles.

If you'd still like to try to land this change, please open another PR for it.

PR #2143, opened 2022-02-18
[devtools] Add devTools-tip video for CSS Grid

chybie:


OrKoN (approved):

jeffposnick:
Sorry to detail this PR, but I'm curious whether there's been a discussion in place around tweaking our `{% Video %}` embed tag to prevent what might be a negative user experience.

Right now, with `autoplay="true"` set, visiting https://deploy-preview-2143--developer-chrome-com.netlify.app/docs/devtools/recorder/ will download the 17.2mb movie unconditionally, and it will start playing without user interaction. Also, note that you have `muted="false"` is set, but we [force `muted`](https://github.com/GoogleChrome/webdev-infra/blob/877a8b2f488fb440dd37dd7e46e4aa3016731667/shortcodes/Video.js#L93) whenever `autoplay` is set, so there may be wasted bytes being transferred for an audio track that is never audible. (Unless the browser is smart enough not to download the audio track when `muted` is set?) In any case, it's good that it is `muted`, because I think many folks would be surprised to hear audio immediately after visiting a docs page.

I am new to d.c.c. and don't know the history behind the `{% Video %}` tag, so I'm going to move this discussion over to https://github.com/GoogleChrome/webdev-infra/issues/24 to make it more formal. I don't want to stand in the way of this PR if it's important to get it merged, but I wanted to at least bring some visibility to the file size.

chybie:
Close this PR as it is a mistake. It's an stale branch. The video has already been removed months ago and replaced with a YouTube video. See https://developer.chrome.com/docs/devtools/recorder/.

PR #2163, opened 2022-02-23
DevTools suggestion: update `console.table`

gabrielsroka:
for `console.table(array [, columns])`, it seems that `columns` can be a string, or an array of strings.


gabrielsroka:
CLA is done

sofiayem (approved):

PR #2166, opened 2022-02-23
Fix typo 'default_icons' -> 'default_icon'

vnoctem:
Fix typo: `default_icons` should be `default_icon`
Screen Shot 2022-02-22 at 9 48 32 PM


jeffposnick (approved):
Based on https://developer.chrome.com/docs/extensions/reference/action/#icon and the later inline example in the same doc, I feel comfortable that this change is correct and can be merged.

PR #2183, opened 2022-02-25
Enable workbox-webpack-plugin TSDoc generation

### LINKED ISSUE #1936: Enable `workbox-webpack-plugin` type generation
Presently the `workbox-webpack-plugin` library is not having it's types generated as it doesn't have TS types. As soon as it does we should enable them here.

See GoogleChrome/workbox#2882

jeffposnick:
Fixes #1936

Workbox v6.5.0 includes TypeScript definitions for `workbox-webpack-plugin`, so we can generate the TSDocs now.

### DIFF
--- a/external/build/lib/dts-parse.js
+++ b/external/build/lib/dts-parse.js
@@ -807,6 +807,7 @@ module.exports = async function parse({silent, sources, mode}) {
   /** @type {typedoc.TypeScript.CompilerOptions} */
   const typescriptOptions = {
     declaration: true,
+    esModuleInterop: true,
   };
 
   /** @type {Partial<typedoc.TypeDocOptions>} */
--- a/external/build/workbox-types.js
+++ b/external/build/workbox-types.js
@@ -25,7 +25,7 @@ const workboxPackages = [
   'workbox-routing',
   'workbox-strategies',
   'workbox-streams',
-  // 'workbox-webpack-plugin',
+  'workbox-webpack-plugin',
   'workbox-window',
 ];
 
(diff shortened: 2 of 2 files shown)


external/build/lib/dts-parse.js
```diff
@@ -807,6 +807,7 @@ module.exports = async function parse({silent, sources, mode}) {
   /** @type {typedoc.TypeScript.CompilerOptions} */
   const typescriptOptions = {
     declaration: true,
+    esModuleInterop: true,
```
jeffposnick, line 810: The `webpack` dependency that `workbox-webpack-plugin` imports is still CommonJS, so this flag is needed.

external/build/workbox-types.js
```diff
@@ -54,6 +54,12 @@ async function fetchAndPrepare(packages, targetDir) {
 async function run() {
   const t = tmp.dirSync();
   try {
+    // webpack is a peerDependency of workbox-webpack-plugin, and needs to be
+    // manually installed.
+    childProcess.execFileSync('npm', ['install', 'webpack'], {
+      cwd: t.name,
+      stdio: 'inherit',
+    });
```
malchata, line 62: Is there a reason why this can't be specified in `package.json` under the `peerDependencies` key?

malchata (approved):
Minor question, otherwise LGTM!

external/build/workbox-types.js
jeffposnick: It is, in the `package.json` for `workbox-webpack-plugin`: https://github.com/GoogleChrome/workbox/blob/b47500cd8a20719f4f2d7f163139408234a70f4a/packages/workbox-webpack-plugin/package.json#L30-L32

`peerDependencies` don't get installed by default when you install an `npm` package—at least not in the version of `npm` that we're using. `npm` versions 1 and 2 did, but versions 3 through 6 (which is currently being used) stopped doing that. Apparently it will start happening again in `npm` version 7. See See https://nodejs.org/es/blog/npm/peer-dependencies/#using-peer-dependencies for more info.

Issue #281, opened 2021-01-21
Extensions docs: MV3 content scripts page needs more


awfuchs:
The [content scripts](https://developer.chrome.com/docs/extensions/mv3/content_scripts/) page was initially adapted for MV3, and is suitable for release, but needs additional work. There is some TODO in the file, bracketed by an {% if false %}; also the following summary was provided in the review of [PR #268](https://github.com/GoogleChrome/developer.chrome.com/pull/268):

> Good enough shape for now, but we should definitely come back to this. We still need to…
> 
> - flesh out the intro of the "Inject with dynamic declarations" section to better match the style of the sibling sections
> - flesh out the examples in the "Inject with dynamic declarations" section
> - link to the appropriate Scripting API documentation when it exists
> - link to an example extension to show off content script injection and usage patterns
>

PR #2232, opened 2022-03-04
Fix broken links to Match Patterns page

gileswells:
References #281 (_doesn't fully resolve it so I changed "Fixes" to not auto-close that issue_)

- Update the links to the Match Patterns page to use the url references in the footer of the doc
  - Looks like the footer links were auto-generated but didn't get automatically linked up in a migration (I'm guessing)

CLA was signed today under Giles Wells


sofiayem:
The original links are correct because they are inside an HTML injection. Markdown-style links don't render inside HTML tags.

PR #2252, opened 2022-03-09
WIP: Add draft of experimental docs section

### LINKED ISSUE #1956: Create section under /docs for Experimental API documentation
Currently experimental API documentation is often put into blog posts, which are a mixture of API announcement, request for feedback from developers, tutorial, and documentation. In some cases these articles are linked to from Chrome Status as the documentation for a feature. For example:

- https://web.dev/local-fonts/
- https://web.dev/storage-foundation/

The ultimate aim for an API it to become part of a stable browser, at that point it is possible to document it over on MDN. However there is a period between launch and MDN docs suitability where things have no docs at all.

I'm proposing to document these experimental APIs on developer.chrome.com under /docs/experimental. The information there should follow similar content guidelines to those used on MDN, as the aim will be to build up documentation on our side and then move these docs to MDN on shipping into stable. That said, some docs are better than nothing, so for a new API if all we have is some brief overview information and a code example, that's fine and we can add to that as the trial progresses.

This will achieve a few things:

- It will give developers one place to find our experimental APIs.
- It will make it easier to see what we have in the list, and remember to make changes as APIs progress through the experimental status.
- It will give us a head start on MDN docs.

Note that I'm not intending to prevent announcement or feedback gathering blog posts to be written, these are all useful too and can be linked from the docs. However, I'd like to extract the documentation part into one place.

Technically, I don't think we need anything different to any of the other documentation sets. If that's the case I can build this out as I start to move existing experimental things over. Raising this issue for awareness and any thoughts that anyone else has.

rachelandrew:
Fixes #1956 

An initial draft to demonstrate this. I've also pulled in existing content about origin trials which I think should live here rather than buried in the blog.

Ultimately this should have sections for origin trials, developer trials, features in Canary, meaning that blog posts can contain announcements and docs can be documentation and kept up to date as the trial progresses. I'm seeing these as pretty much like an MDN API Overview page, as ultimately we would move this to MDN once the feature hits stable. 

To do:

- [x] Collect launch set of docs for ongoing trials from the blog.
- [x] Add sections to main page for different trials.
- [x] Create icon for experimental docs.
- [x] Link from main docs page.


site/en/docs/handbook/web-platform/index.md
```diff
@@ -0,0 +1,72 @@
...
+`interfaceName`
+: Short description.
+
+## Use cases 
+
+Change the heading above to make sense for your content, 
+however here you can add a short explanation, 
+and the core use cases for the API.
+
+## Examples
```
tomayac, line 49:
```suggestion
## Use cases 

Change the heading above to make sense for your content, 
however here you can add a short explanation, 
and the core use cases for the API.

## Current status

Add information about status, and how developers can try out the API. 

## Interfaces

Describe each interface that is part of the API 
using a description list format.

`interfaceName`
: Short description.

## Examples
```

tomayac: Should we move **Use Cases** right after the **Outline**, so developers can early on decide if this is something they need?

site/en/docs/web-platform/index.md
```diff
@@ -0,0 +1,7 @@
+---
+layout: 'layouts/project-landing.njk'
+title: 'Web Platform'
+description: 'Documentation for Web Platform APIs.'
+---
+
+This section contains documentation for Web Platform APIs. Some are currently in an experimental status in Chrome. For example, origin and developer trials, or features only found in Canary. Others have shipped in stable Chrome, however are unimplemented or currently under development in other engines.
```
tomayac, line 7:
```suggestion
This section contains documentation for Web Platform APIs. Some are currently in an experimental status in Chromium. For example, origin and developer trials, or features only found in Canary. Others have shipped in stable Chromium, however are unimplemented or currently under development in other engines.
```

tomayac: Not sure about this one. We might suggest that not all Chromium-derivates enable an API if it's supported in the engine. Example: Brave chose to disable the File System Access API despite the engine supporting it.

site/en/docs/handbook/web-platform/index.md
rachelandrew: I'd like to have the current status above it, because it's not helpful for developers to find something they need then realise they can't use it. So maybe reorder to status, then use cases, then interfaces?

site/en/docs/web-platform/index.md
rachelandrew: Yeah, I was thinking that this is really Chrome, or at least we can be sure they are available in Chrome.

With shipped things we should be clear about status, especially if we know that some Chromium-based browsers haven't shipped. So I think I'll leave this as Chrome based on that?

tomayac: Let's leave it at Chrome. That's one advantage of writing on DCC :-)

site/en/docs/handbook/web-platform/index.md
tomayac: Likewise sounds good to me. Deleting the suggestion then.

rachelandrew:
> Should we move Use Cases right after the Outline, so developers can early on decide if this is something they need?

I thought I'd replied to this but apparently not! I want to make sure the status is prioritized. As one of the things that annoys people is when they think something will solve their problem then realize it's only a trial. I've moved Use Cases up under Status though.

> Not sure about this one. We might suggest that not all Chromium-derivates enable an API if it's supported in the engine. Example: Brave chose to disable the File System Access API despite the engine supporting it.

Thinking about it I'm not sure we need say anything. So I've removed that.

devnook (approved):

PR #2302, opened 2022-03-15
Re-add chromeOsOnly metadata to generated Chrome types

### LINKED ISSUE #2298: chromeOsOnly warning not implemented since Oct 2021
**Describe the bug**
The "This API works only on Chrome OS" warning is not implemented since commit fd6a63977ac8768ca331197785b0bb6445c1129f (2021-10-28)

**To Reproduce**
1. Visit https://developer.chrome.com/docs/extensions/reference/certificateProvider/
2. Notice that there is no "This API works only on Chrome OS" warning.
3. Compare to pre-commit Wayback Machine snapshot: https://web.archive.org/web/20211015050419/https://developer.chrome.com/docs/extensions/reference/certificateProvider/
4. See the warning before that commit.

**Expected behavior**
The warning should continue to display for relevant extension APIs

**Screenshots**
Current website
Pre-commit archive snapshot

**Additional context**
Code that mentions it's a TODO: https://github.com/GoogleChrome/developer.chrome.com/blob/main/site/_includes/layouts/namespace-reference.njk#L54-L61

### LINKED ISSUE #2542: Chrome OS specific permissions are not listed as such
https://developer.chrome.com/docs/extensions/reference/loginState/
currently shows it as "Chrome 78+" while the master-branch of chromium still lists it as "chromeos".

Also see:
https://github.com/GoogleChrome/developer.chrome.com/issues/607
https://source.chromium.org/chromium/chromium/src/+/master:chrome/common/extensions/api/_permission_features.json;l=495;drc=ba4cc4353e4048c46d51ffbbf461de1fa7fa0e79

jeffposnick:
Fixes #2298, fixes #2542

This adds `chromeOsOnly: true` metadata to a given Chrome feature if, when parsing its TypeScript docs, there is only one `@chrome-platform` tag, and it's set to `@chrome-platform chromeos`.

You can see an example of such as feature at, e.g., lines 6720-6726 of https://unpkg.com/browse/chrome-types@0.1.94/_all.d.ts:

```js
/**
 * Use this API to expose certificates to the platform which can use these certificates for TLS authentications.
 *
 * @since Chrome 46
 * @chrome-permission certificateProvider
 * @chrome-platform chromeos
 */
```

There are a handful of other features, beyond `chrome.certificateProvider`, that are flagged as well.

The rendering looks like:

Screen Shot 2022-03-15 at 10 47 33 AM

### DIFF
--- a/external/build/lib/dts-parse.js
+++ b/external/build/lib/dts-parse.js
@@ -708,6 +708,9 @@ class Transform {
       channel: 'stable',
     };
 
+    // See https://github.com/GoogleChrome/developer.chrome.com/issues/2298
+    let chromeOsOnly = undefined;
+
     tags.forEach(({tag, text}) => {
       text = text.trim(); // some show up with extra \n
 
@@ -749,8 +752,23 @@ class Transform {
         case 'chrome-disallow-service-workers':
           out.disallowServiceWorkers = true;
           break;
+        case 'chrome-platform':
+          // If chromeos is the platform, and chromeOsOnly is undefined because
+          // we haven't seen any other platforms, this might be chromeOsOnly.
+          if (text === 'chromeos' && chromeOsOnly === undefined) {
+            chromeOsOnly = true;
+          } else {
+            // The first time we see a platform that's not chromeos, we know the
+            // feature isn't chromeOsOnly.
+            chromeOsOnly = false;
+          }
+          break;
       }
     });
+
+    if (chromeOsOnly === true) {
+      out.chromeOsOnly = true;
+    }
 
     return out;
   }
--- a/site/_includes/layouts/namespace-reference.njk
+++ b/site/_includes/layouts/namespace-reference.njk
@@ -51,7 +51,6 @@
           </div>
         {% endif %}
 
-        {# TODO: unsupported #}
         {% if f.chromeOsOnly %}
           {# There are some historic cases of multi-platform support in Platform Apps, but
               the only restriction these days is CrOS vs everywhere. #}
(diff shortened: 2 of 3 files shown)


ANONYMOUS25K:
RE:MUHAMMAD SHAHRUL FAHMI BIN MOHD ZAIDI<danialfahmim@gmail.com>

ANONYMOUS25K:
RE:MUHAMMAD SHAHRUL FAHMI BIN MOHD ZAIDI<danialfahmim@gmail.com>

pyBlob:
I neither see the "Chrome OS only" badge, nor the related notice box in the linked developer preview.

The logic that extracts the chromeOsOnly flag, can be written more clearly in two steps. This looks more closely like the other cases that handle multiple values.
1a. Initialize platforms = []
1b. When encountering a `chrome-platform` tag, push it into the platforms array
2. chromeOsOnly = set_equals(platforms, ["chromeos"])

This test is not enough for some APIs, like the `chrome.loginState`. They actually have two platforms, none of which is chrome:
```
* @chrome-platform chromeos
* @chrome-platform lacros
```

jeffposnick:
@pyBlob, the Netlify preview uses JSON data for Chrome features that's [synced from Google Cloud Storage](https://github.com/GoogleChrome/developer.chrome.com/blob/main/external/sync-external.js).

The change in this PR will affect how this JSON data is generated, but you need to run a local dev server based on the updated JSON to see it in a preview. That's why I included a screenshot.

I do not know the best UI treatment for a feature that's tagged with `@chrome-platform lacros` as well as `@chrome-platform chromeos`. My initial reaction is that reserving the `chromeOsOnly` metadata for features that legitimately only have `@chrome-platform chromeos` has an element of correctness to it.

devnook (approved):

PR #2311, opened 2022-03-16
ContentMigrationAuthors#2

simgobb:
Added all the missing authors of the articles newly labelled as DCC in the ContentMigration Sheet

@rachelandrew

### DIFF
--- a/site/_data/i18n/authors.yaml
+++ b/site/_data/i18n/authors.yaml
@@ -1,3 +1,193 @@
+mustaqahmed:
+  title:
+    en: 'Mustaq Ahmed'
+  description:
+    en: 'Mustaq is a contributor to WebFundamentals'
+shampson:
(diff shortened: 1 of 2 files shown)


site/_data/i18n/authors.yaml
```diff
@@ -1,3 +1,193 @@
+mustaqahmed:
+  title:
+    en: 'Mustaq Ahmed'
+  description:
+    en: 'Mustaq is a contributor to WebFundamentals'
```
rachelandrew, line 5: can we change these to a contributor to Chrome Developers

rachelandrew (changes requested):
As noted inline can we update the description to mention Chrome Developers not Web Fundamentals.

rachelandrew (approved):

PR #2391, opened 2022-03-23
Adding tags to DevTools docs

sofiayem:
New tags are the following:
- Collections from old table of contents: css, javascript, performance, storage
- New CUJ-based collections: get-started, prototype-fixes, find-issues, emulate, test

### DIFF
--- a/site/_data/i18n/tags.yml
+++ b/site/_data/i18n/tags.yml
@@ -31,7 +31,7 @@ javascript:
   en: JavaScript
 
 get-started:
-  en: Get started with DevTools
+  en: Get started
 
 prototype-fixes:
   en: Prototype fixes


site/_data/i18n/tags.yml
```diff
@@ -23,6 +23,34 @@ autoplay:
...
 
+# DevTools-related tags
+css:
+  en: CSS
+
+javascript:
+  en: JavaScript
+
+get-started:
+  en: Get started with DevTools
```
heyawhite, line 34: It would be great to make this a bit more generic -- I wouldn't be surprised if other blog posts end up using "Get started" that aren't related to DevTools

heyawhite (approved):
One request for a change, but otherwise LGTM

site/en/docs/devtools/css/print-preview/index.md
```diff
@@ -7,6 +7,9 @@ date: 2018-12-14
 #updated: YYYY-MM-DD
 description:
   "Open the Rendering tab and select Emulate CSS media > print."
+tags:
+  - emulate
+  - test
```
jecfish, line 12: maybe css as well?

site/en/docs/devtools/recorder/index.md
```diff
@@ -6,6 +6,8 @@ authors:
 date: 2021-11-02
 updated: 2021-11-03
 description: "Record, replay and measure user flows with the Recorder panel."
+tags:
+  - test
 ---
```
jecfish, line 11: "performance" as well, because there are options to run performance analysis.

site/_data/i18n/tags.yml
```diff
@@ -23,6 +23,34 @@ autoplay:
...
+  en: CSS
+
+javascript:
+  en: JavaScript
+
+get-started:
+  en: Get started
+
+prototype-fixes:
+  en: Prototype fixes
```
jecfish, line 37: this is not a common term we use as web developer

sofiayem: It was suggested by the PM. I want to see how this tags does in terms of usage statistics.

PR #2412, opened 2022-03-24
Redirect i18n URLs that would be valid without the locale prefix

### LINKED ISSUE #2398: [High priority] Non-translated articles return 404
When you go to a localized page such as https://developer.chrome.com/ja/docs/privacy-sandbox/
The links work as expected as long as the destination is also localized. However, when they are *not* localized, d.c.c returns 404 - Awe snap!. For example: https://developer.chrome.com/ja/docs/privacy-sandbox/chips/

Redirecting them to the original English page is preferable.

jeffposnick:
R: @devnook 
CC: @agektmr 
Fixes #2398

This shares a similar goal to https://github.com/GoogleChrome/web.dev/pull/7283, but that PR was never merged, and the d.c.c. serving infrastructure differs from web.dev, so I wanted to write something fresh and focused.

The behavior implemented in this PR is:

- If any navigation would return a `404` today, check whether replacing the first item in the path with the default locale (`en`) would lead to a valid HTML document.
  - If so, instead of returning a `404`, return a `302` redirecting to the non-localized version of that URL.
  - If not, continue returning a `404` as before.

One approach that I considered, but ultimately decided against, was returning the HTML content from the default locale as a `200` response directly from the localized URL, instead of redirecting. That seemed riskier from the point of view of indexing, so I went with a redirect instead.


jeffposnick:
I realize that we have some tests for server functionality, so I'll add a new test to this PR to cover the `getNonLocalizedURL()` logic.

jeffposnick:
I've added a few tests for the `getNonLocalizedURL()` logic to this PR.

devnook (approved):

jeffposnick:
Let's merge on Monday, as I would prefer there not to be any surprises with our server code over the weekend 😄

devnook:
I noticed lhci is unhappy, because robots-txt changed location. Let's resolve it before merging.

jeffposnick:
Are we sure that LHCI error is legitimate? I don't see a https://developer.chrome.com/robots.txt on the production website, either. Let's chat about it on Monday.

server/not-found.js
```diff
@@ -21,10 +21,75 @@ const express = require('express');
...
+  // See https://github.com/GoogleChrome/developer.chrome.com/issues/2398
+  const nonLocalizedURL = getNonLocalizedURL(req.path);
+  if (nonLocalizedURL !== null) {
+    // Use a 302 redirect here, since there might be a localized version of this
+    // URL at some point in the future.
+    return doRedirect(res, nonLocalizedURL, 302);
+  }
+
+  // Otherwise, we have a 404, not a redirect.
   res.status(404);
```
devnook, line 93: You need to exempt here a robots.txt file, or, alternatively, provide a valid robots.txt at the root of the project /dist

devnook (changes requested):
With the new server logic we return a "404" page for /robots.txt, which is a "malformed" robots file, while previously it was a real native 404.

Simple solution would be to add a robots.txt same as in web.dev

server/not-found.js
jeffposnick: Done, in https://github.com/GoogleChrome/developer.chrome.com/pull/2517

devnook (approved):

Naveed-Iftikhar (commented):
404 and 403 open

Issue #2439, opened 2022-03-28
Get a better handle on 11ty memory usage


jeffposnick:
Our production build's 11ty step has been failing recently with out of memory errors. We have a similar experience on web.dev.

Short-term, we can address this by setting `NODE_OPTIONS="--max_old_space_size=4096"` (or higher), [similar to what we do on web.dev](https://github.com/GoogleChrome/web.dev/blob/9a197fd63a3c87249eb5145736ba35f291b58ccd/.cloudbuild/deploy.yaml#L23).

Long-term, we should get a handle (memory joke!) on what exactly it is about our 11ty build is causing such excessive memory usage, to the point where the default heap size is ineffective.

PR #2440, opened 2022-03-28
Increase node heap size in deploy builds

jeffposnick:
R: @devnook 

This is related to #2439, but does not address the long-term issue.

Our web.dev production build uses [8gb of heap size](https://github.com/GoogleChrome/web.dev/blob/9a197fd63a3c87249eb5145736ba35f291b58ccd/.cloudbuild/deploy.yaml#L23), but we also have it configured to [use `E2_HIGHCPU_32`](https://github.com/GoogleChrome/web.dev/blob/9a197fd63a3c87249eb5145736ba35f291b58ccd/.cloudbuild/deploy.yaml#L35).

I'm guessing we can't safely configure more than 4gb of explicit heap size if our current configuration for d.c.c. is `E2_HIGHCPU_8` (i.e. and 8gb machine). I'm not sure what the implication would be if we switched over to `E2_HIGHCPU_32` in this PR, and the `# yolo` comment for the current VM configuration doesn't give me a huge amount of confidence...

### DIFF
--- a/.cloudbuild/deploy.yaml
+++ b/.cloudbuild/deploy.yaml
@@ -11,6 +11,8 @@ steps:
   - name: node
     entrypoint: npm
     args: ['run', 'production']
+    env:
+      - 'NODE_OPTIONS="--max_old_space_size=4096"' # https://github.com/GoogleChrome/developer.chrome.com/issues/2439
 
   - name: node
     entrypoint: npm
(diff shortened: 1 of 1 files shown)


.cloudbuild/deploy.yaml
```diff
@@ -19,18 +21,18 @@ steps:
   - name: 'gcr.io/cloud-builders/gcloud'
     entrypoint: 'bash'
     args:
-    - '-c'
-    - |
-      # This snippet lets us pass additional args via the Cloud Build console
-      # in the _EXTRA_GCLOUD_ARGS var.
-      # nb. We don't have to specify --project; it's part of the environment.
-      gcloud app deploy ${_EXTRA_GCLOUD_ARGS}
+      - '-c'
```
jeffposnick, line 24: My IDE's code formatter fixed a few things that were non-standard about the current YAML. I can revert them if needed, but it seemed harmless.

devnook (approved):

PR #2449, opened 2022-03-29
Generalize featured card on homepage

devnook:


jeffposnick:
This is what it looks like on localhost when I add in a placeholder `/100/` file:

Screen Shot 2022-03-29 at 7 55 44 AM

As long as that look as intended to you, I'm fine with merging.

jeffposnick (approved):

PR #2451, opened 2022-03-29
Dino Image upload line:no 924

yogesh-google:

- Image upload for Chrome dino game card.


jeffposnick:
As an FYI, this is what it looks like on localhost:

Screen Shot 2022-03-29 at 9 02 04 AM

jeffposnick (approved):

PR #2452, opened 2022-03-29
Replace heart icons with like-icon

devnook:
Replaces for now 2 first icons with like-icon


jeffposnick (commented):
The functionality looks 👍 from doing some local testing.

One thing that I wanted to confirm is that `<like-icon>` has different `data-ga-` attributes than what's currently checked in to the HTML. Are we expecting specific Google Analytics data to be sent?

PR #2455, opened 2022-03-29
chrome-100 release

devnook:


jeffposnick (changes requested):
I think only the `banner.yml` changes need to be in this PR, right?

site/_data/banner.yml
```diff
@@ -0,0 +1,6 @@
+type: info
+text: "Take a stroll down memory lane and celebrate #100CoolWebMoments since Chrome's first release."
+actions:
+  - text: Discover
+    href: /100
+  - text: Dismiss
```
jeffposnick, line 6: Can you add a newline here?

jeffposnick (approved):

PR #2479, opened 2022-03-30
[Chrome 100] Fixes wording in "PWAs on desktop OS" item

### LINKED ISSUE #2478: [/100 page] 2019 card description fix
In the 2019 card on developer.chrome.com/100, please update the description to say "desktop OSes" instead of "desktop browsers"
image

jeffposnick:
Fixes #2478

Instead of the suggestion "desktop OSes", I went with "desktop operating systems", because a) I don't think "OSes" is the correct pluralization, and b) there is room in the card for the unabbreviated text.

Screen Shot 2022-03-30 at 9 37 23 AM


harleenkbatra08:
LGTM

PR #2483, opened 2022-03-31
Content updates and css fixes

rutuja-google:

- Content updates and css fixes.
- Updated incorrect values of item for like-icon tag at various places.

image

### DIFF
--- a/site/_includes/partials/chrome-100-card-section.njk
+++ b/site/_includes/partials/chrome-100-card-section.njk
@@ -188,27 +188,10 @@
             <p class="card-description">One of the gamechangers for web layouts.</p>
             <div class="display-flex gap-top-200 logos">
               {% Img src="image/x1Los57vDga6OEMNi1dIJwZ0qvp2/PYajnvHnl9nnQpdWDfzA.svg", alt="Promise tag", class="tags gap-top-100", width="120", height="16" %}
-              <like-icon item="100-card-61"></like-icon>
+              <like-icon item="100-card-7"></like-icon>
             </div>
           </div>
         </div>
-        <div class="card gap-bottom-300">
-          <div class="card-info card-top">
-            <div class="arrow-wrapper">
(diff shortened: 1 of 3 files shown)


site/_includes/partials/chrome-100-card-section.njk
```diff
@@ -188,27 +188,10 @@
             <p class="card-description">One of the gamechangers for web layouts.</p>
             <div class="display-flex gap-top-200 logos">
               {% Img src="image/x1Los57vDga6OEMNi1dIJwZ0qvp2/PYajnvHnl9nnQpdWDfzA.svg", alt="Promise tag", class="tags gap-top-100", width="120", height="16" %}
-              <like-icon item="100-card-61"></like-icon>
+              <like-icon item="100-card-7"></like-icon>
```
devnook, line 191: If you change the id of the like icon, you change the asociated likes count. Can you please explainn in the PR description why this change is needed?

devnook (changes requested):

site/_includes/partials/chrome-100-card-section.njk
rutuja-google: If you check on live it looks like the item values for card mentioned in PR description is 100-card-61 whereas its value should be 100-card-7 as per sequence and placement of cards.

We have received few cards placement changes from Harleen and we have modified their item values as well.

If you are fine to keep the current item values then we can revert back the changes. Please suggest.

devnook: The ID is used to associate number of likes with a given card, and this is already running 2 days in the production. It is essential that the IDs (`item` value on `<like-icon>`) stays the same for a given card, doesn't matter where the card is placed in the HTML. Also, if you add a new card, please ensure you give it a new, unique ID (`item` value).

harleenkbatra08: We should keep the card id consistent so we don't lose the associated likes count. I don't mind if the number is different based on their placement. Please keep the card id same as before!

rutuja-google: Reverted the item values. Preview build is in process.

devnook (approved):

PR #2488, opened 2022-03-31
[DO NOT MERGE] Use 11ty 1.0.1 canary

jeffposnick:
This updates our 11ty installation to use the 1.0.1 canary (and some related plugins), which has some performance improvements that will help us out. This matches what we're using on web.dev.

I did a spot check of the local build output and ran things through https://github.com/11ty/eleventy-upgrade-help, and did not encounter any warnings for features we were using. We also had a smooth upgrade when we performed a similar upgrade for web.dev, so I do not anticipate any issues.


jeffposnick:
Locally, for `npm run production`, I'm seeing:

```
[11ty] Copied 726 files / Wrote 3383 files in 339.62 seconds (100.4ms each, v1.0.1-canary.3)
```

```
Copied 726 files / Wrote 3384 files in 353.57 seconds (104.5ms each, v0.11.1)
```

So... not actually a huge speedup locally either. Definitely not what we were seeing with web.dev.

I'm going to still pursue an upgrade, because we can't stick on an outdated version of 11ty forever, but let's hold off on merging this PR for the time being.

jeffposnick:
I'm going to create a fresh PR for this, as resolving the package conflicts is going to be a bit annoying otherwise.

PR #2555, opened 2022-04-08
Switch to a Rust-based HTML & inline CSS/JS minifier

jeffposnick:
Switching away from `html-minifier` and `csso` in favor of `@minify-html/js` leads to production builds that are around 33% faster on my M1 Macbook.

I'm not sure what the corresponding speedup will be in the production deployment, but if web.dev was any indication, it should be meaningfully faster.

That being said, we have run into a number of issues with this HTML minifier when used on web.dev because it treats invalid HTML differently than `html-minifier` did. We need to keep an eye out for any reports of issues with HTML rendering on the production site following this deployment.


devnook (changes requested):
testing approvals

devnook (approved):

site/_includes/layouts/base.njk
```diff
@@ -12,8 +12,8 @@
       <link rel="alternate" type="application/rss+xml" title="{{ site.title }}" href="{{ site.url }}/feeds/blog.xml">
     {% endblock %}
 
-    <link rel="preload" as="font" crossorigin href="/fonts/google-sans-v2003/regular/latin.woff2">
-    <link rel="preload" as="font" crossorigin href="/fonts/google-sans-v2003/medium/latin.woff2">
+    <link rel="preload" as="font" crossorigin="anonymous" href="/fonts/google-sans-v2003/regular/latin.woff2">
```
jeffposnick, line 15: See https://github.com/GoogleChrome/web.dev/pull/7790

devnook (approved):

devnook (approved):

jeffposnick:
I'm going to do a one-off run of an HTML analyzer on the minifier output before we merge this, just to get a gauge for how broken things might be.

jeffposnick:
Actually, this is using an older version of the HTML minifier. See https://github.com/GoogleChrome/web.dev/issues/8381

matthiasrohmer:
Since the API of @minify-html/node changed quite a bit since this implementation, closing this and tracking further in #4048.

Issue #2556, opened 2022-04-08
Re-evaluate our PurgeCSS approach


jeffposnick:
I've found that running [PurgeCSS](https://purgecss.com/) against each HTML document we generate contributes to about 25% of our total production build time on my M1 Macbook:

https://github.com/GoogleChrome/developer.chrome.com/blob/e55c39d577eeb468ebf25721f49616afb4614be6/site/_transforms/purify-css.js#L37-L63

At the same time, using PurgeCSS serves an important purpose, as there's approximately 200kb of total CSS in our `main.css` file, and most of that is unused on any given page.

I'm new to the project, so I am probably missing some important context about why this approach was adopted. If we're interested in optimizing our production build speed, I think we need to revisit some alternatives that don't rely on PurgeCSS, like:

- Generating a more targeted subset of styles for each logical section of the site (`/blog/`, `/docs/`, etc.) and just using that subset, without PurgeCSS, on each of those pages. If we went with that approach, we should also consider loading this CSS from a versioned URL in a `<link>` instead of inlining it in each page's `<style>`, and take advantage of cacheability.

- Creating a single, site-wide CSS file that also gets loaded via a `<link>` instead of inlined, but which is significantly smaller than 200kb. I.e., find all the styles that are not currently being used anywhere on the site, and just remove them from our sources completely.

jeffposnick:
@devnook and I discussed this a bit, and we wanted to see whether the size of the input CSS (`dist/css/main.css`) had a substantial effect on the build time, or whether just using `purgeCSS` at all was always going to be slow, given the number of documents we have. I also wanted to evaluate the performance impact of using `csso` optimization, even though we can switch to a Rust-based alternative in #2555.

Here's a table showing the build time for `NODE_ENV=production npm run eleventy` given various combinations of purgeCSS and csso being enabled/disabled, both using our full 200kb CSS file as input as well as passing in a truncated set of CSS roughly 100kb in size.

(The big caveat is that I ran this on my M1 Macbook Pro, and in most scenarios, only ran a single build instead of averaging multiple builds.)

|purgeCSS?|csso?|Size in kilobytes of main.css|Build time in ms per file|
|---------|-----|-----------------------------|-------------------------|
|Yes      |Yes  |200                          |72                       |
|Yes      |No   |200                          |61                       |
|No       |Yes  |100                          |53                       |
|Yes      |Yes  |100                          |47                       |
|Yes      |No   |100                          |47                       |
[...]

PR #2620, opened 2022-04-18
async worker pool for CSS purge + minification

jeffposnick:
This will hopefully offer some improvements for #2556, by using the same worker pool approach we already use for our DOM transforms when implementing our CSS purging.

I would still like to find a way to move off of per-file CSS purging in general, so I want to keep #2556 open, but given that this PR should improve things with minimal changes, I wanted to implement it first.

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -38,13 +38,13 @@ const {LanguageList} = require('./site/_shortcodes/LanguageList');
 
 // Transforms
 const {domTransformer} = require('./site/_transforms/dom-transformer-pool');
-const {purifyCss} = require('./site/_transforms/purify-css');
+const {purifyCss} = require('./site/_transforms/purify-css-pool');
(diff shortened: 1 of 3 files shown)


.eleventy.js
```diff
@@ -88,17 +88,19 @@ module.exports = eleventyConfig => {
...
-      .getFilteredByGlob(`./site/${locale}/blog/*/*.md`)
-      .filter(filterOutDrafts)
-      .reverse();
-    // If we're running inside of Percy then just show the first six blog posts.
-    if (process.env.PERCY_BRANCH) {
-      blogCollection = blogCollection.slice(blogCollection.length - 6);
-    }
-    return blogCollection;
-  }));
+  locales.forEach(locale =>
```
jeffposnick, line 91: This is due to my IDE's auto-formatting, but it's probably a good idea to include it just to keep everything consistent.

devnook (approved):

PR #2622, opened 2022-04-18
Update deploy Cloud Build task

jeffposnick:
This brings our d.c.c. `deploy` task in line with how we have web.dev's [`deploy` configured](https://github.com/GoogleChrome/web.dev/blob/main/.cloudbuild/deploy.yaml) configured.

- Explicitly request `node:16.14.2`. Right now, the generic `node` alias resolves to a Docker container using node v17.9.0, which I don't think anyone on the team uses during regular development.
- Request `E2_HIGHCPU_32` instead of `E2_HIGHCPU_8`. That beefier VM is what we use in web.dev already.
- Bumps up `max_old_space_size` from 4gb to 8gb, since we have the extra RAM and under the impression that fewer garbage collections will lead to a faster build.

### DIFF
--- a/.cloudbuild/deploy.yaml
+++ b/.cloudbuild/deploy.yaml
@@ -1,20 +1,23 @@
+# This deploys the site. Roughly equivalent to
+# https://github.com/GoogleChrome/web.dev/blob/main/.cloudbuild/deploy.yaml
+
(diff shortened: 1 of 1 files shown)


devnook (approved):

GO2D77:
thanks

PR #2631, opened 2022-04-19
DevTools > 101 > Elements > Styles > cascade layers

sofiayem:
Documented https://developer.chrome.com/blog/new-in-devtools-101/#layer

Also https://developer.chrome.com/blog/new-in-devtools-101/#hwb

### DIFF
--- a/site/en/docs/devtools/css/reference/index.md
+++ b/site/en/docs/devtools/css/reference/index.md
@@ -124,48 +124,6 @@ If your browser supports the `lab()` function, the element is green, otherwise i
 **Note**: At the time of writing, only Safari [supports the CIELAB color space](https://caniuse.com/?search=lab).
 {% endAside %}
 
-### View cascade layers {: #cascade-layers}
-
-[Cascade layers](/blog/cascade-layers/) enable more explicit control of your CSS files to prevent style-specificity conflicts. This is useful for large codebases, design systems, and when managing third-party styles in applications.
-
-To view cascade layers, [inspect](/docs/devtools/open/#elements) the element below and open **Elements** > **Styles**.
-
-<div class="cascade-box"></div>
-<style>
-    .cascade-box{
-    width: 250px;
-    height: 30px;
-    text-align: center;}
-    .cascade-box::after { 
-        content: "My styles are layered!";
-      }
-    /* Define the specificity */
-@layer base, component, page;
-@layer page {
-  .cascade-box {
-    background: palegreen;
-  }
-}
-@layer base {
-  .cascade-box {
-    background: rebeccapurple;
-  }
-}
-@layer component {
-  .cascade-box {
-    background: hotpink;
-  }
-}
-</style>
-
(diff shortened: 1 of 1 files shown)


site/en/docs/devtools/css/reference/index.md
```diff
@@ -121,6 +124,48 @@ If your browser supports the `lab()` function, the element is green, otherwise i
 **Note**: At the time of writing, only Safari [supports the CIELAB color space](https://caniuse.com/?search=lab).
 {% endAside %}
 
+### View cascade layers {: #cascade-layers}
```
jecfish, line 127: Mov eit to lower part of the document as this is not as important and common feature than others.

site/en/docs/devtools/css/reference/index.md
```diff
@@ -494,3 +542,7 @@ To open the **Shadow Editor**:
 [26]: https://material.io/design/color/the-color-system.html#color-usage-and-palettes
 [27]: #select
 [28]: /docs/devtools/shortcuts#styles
+[29]: https://drafts.csswg.org/css-color/#rgb-functions
+[30]: https://drafts.csswg.org/css-color/#the-hsl-notation
+[31]: https://drafts.csswg.org/css-color/#the-hsl-notation
```
jecfish, line 547: is this link for hwb too?

jecfish (changes requested):

site/en/docs/devtools/css/reference/index.md
```diff
@@ -121,6 +124,48 @@ If your browser supports the `lab()` function, the element is green, otherwise i
 **Note**: At the time of writing, only Safari [supports the CIELAB color space](https://caniuse.com/?search=lab).
 {% endAside %}
 
+### View cascade layers {: #cascade-layers}
+
+[Cascade layers](/blog/cascade-layers/) enable more explicit control of your CSS files to prevent style-specificity conflicts. This is useful for large codebases, design systems, and when managing third-party styles in applications.
+
+To view cascade layers, [inspect](/docs/devtools/open/#elements) the element below and open **Elements** > **Styles**.
+
+<div class="cascade-box"></div>
```
jecfish, line 133: @devnook  @jeffposnick at the moment, we embeded demo in d.c.c like this sometimes. Do we have other recommend way to do it? (How about the web.dev code component?)

In the old WebFu, there's an iframe shortcode which we can used to separate the demo code.

jecfish (approved):

site/en/docs/devtools/css/reference/index.md
jeffposnick: d.c.c. supports `Glitch`, `Codepen`, and a generic `IFrame` shortcodes for embedded content. Though for this sort of example—where you're asking folks to inspect an on-page element—it might be a better user experience to keep the HTML + CSS inline. I could imagine the Codepen/Glitch hosted examples adding in a bunch of extra DOM elements and styles that might distract from what you want the user to focus on.

Issue #2635, opened 2022-04-19
Update our deployment gcloud usage


jeffposnick:
Following some recent PRs, our [11ty build time](https://github.com/GoogleChrome/developer.chrome.com/blob/6f7146d7263705c234b8ece91f62239c35732c6b/.cloudbuild/deploy.yaml#L14-L18) is faster than it's previously been. However, something that currently contributes quite a bit of variability into our overall deployment time is the step in the process that runs `gcloud`:

https://github.com/GoogleChrome/developer.chrome.com/blob/6f7146d7263705c234b8ece91f62239c35732c6b/.cloudbuild/deploy.yaml#L24-L32

The amount of time that this step takes can be from anywhere between 5 and 12 minutes, and the actual duration seems to depend on how many files `gcloud` thinks it needs to upload to Google Cloud Storage to make available for App Engine.

We should do a couple of things:

- Figure out how the set of files that need to be uploaded to Google Cloud Storage is calculated, and whether there's anything we can do to prevent files from being re-uploaded if they haven't changed. For some recent builds, I've seen `Uploading 240 files to Google Cloud Storage` (which ends up taking only around 5 minutes to complete) and `Uploading 1054 files to Google Cloud Storage` (which takes almost 10 minutes to complete).

- Update our `gcloud` command to use a specific image from [`gcr.io/google.com/cloudsdktool/google-cloud-cli`](https://github.com/GoogleCloudPlatform/cloud-sdk-docker), instead of `gcr.io/cloud-builders/gcloud`, which appears to be [less supported](https://github.com/GoogleCloudPlatform/cloud-builders/tree/master/gcloud#gcloud).

devnook:
with the recent updates to staging environment, this is now obsolete.

matthiasrohmer:
Even once we serve from bucket there is still some room for improvement, though we need to evaluate more carefully as for on-demand renders (see /_preview/ route) we will need all of the 11ty config files. We can fore sure skip linting and tooling though.

matthiasrohmer:
Won't fix, due to the pending move to the new CMS.

PR #2637, opened 2022-04-19
[DO NOT MERGE] Exclude uneeded files from App Engine deploys

jeffposnick:
This addresses part of #2635, by restricting the files that we upload to Google Cloud Storage for use in App Engine following our Cloud Build.

(Let's hold off on merging this until next week, but I'm opening this PR now so that I don't forget.)


jeffposnick:
This PR won't work as-is because

https://github.com/GoogleChrome/developer.chrome.com/blob/6f7146d7263705c234b8ece91f62239c35732c6b/server/not-found.js#L25

needs to load data from `site`. I added that line of code a few weeks ago, but it's definitely not worth having to include all of `site` in the App Engine deployment, so I will just switch that over to hardcode `'en'`.

PR #2662, opened 2022-04-21
Update retrying-requests-when-back-online

dartess:
The original version looks weird.
Either we talk about the plugin as a whole entity, and wrap it in "code", or as a plugin responsible for something, and write it in plain text (plugin for background sync maybe?).


malchata (approved):
LGTM

PR #2666, opened 2022-04-21
Update managing-fallback-responses

dartess:
How `ExtendableEvent` doesn't have `request` field in typescript types, so code examples don't work in typescript.


malchata:
@jeffposnick, I think you're the right person to review this one.

jeffposnick (approved):

PR #2722, opened 2022-04-28
Update percy

devnook:
This PR updates our PERCY workflow
+ Updates percy to 1.1. and uses percy snapshot tool instead of percy-agent
+ Defines new list of snapshots to be taken
+ starts server only after eleventy finished initial run. Use --incremental for next eleventy run on :watch command. This prevents confusion that the server started but localhost:8080 returns 404
+ Add translation directories to eleventyignore by default. Can be prevented by the use of ELEVENTY_INCLUDE_TRANSLATED flag


package.json
```diff
@@ -8,11 +8,13 @@
   },
   "scripts": {
     "algolia": "node ./algolia.js",
+    "pre-build": "npm-run-all clean ignore gulp rollup maybe-sync-external",
```
jeffposnick, line 11: I found using the `pre-` prefix here a little confusing, because `pre` (without the `-`) and `post` have a special meaning in `npm` scripts: https://docs.npmjs.com/cli/v8/using-npm/scripts#pre--post-scripts

Given that you're not using the special `pre` behavior, could you name this something different, so that others aren't confused? Like `setup` or something similar?

package.json
```diff
@@ -8,11 +8,13 @@
...
     "clean": "rimraf dist",
     "cloud-secrets": "node ./cloud-secrets.js",
-    "dev": "npm-run-all clean ignore gulp rollup maybe-sync-external --parallel --race gulp:watch rollup:watch eleventy:watch start",
+    "dev": "npm-run-all pre-build eleventy --parallel start watch",
+    "watch": "npm-run-all --parallel --race gulp:watch rollup:watch eleventy:watch",
```
jeffposnick, line 15: Is [`--race`](https://github.com/mysticatea/npm-run-all/blob/master/docs/npm-run-all.md#:~:text=npm%20run%20bar%27.-,%2Dr%2C%20%2D%2Drace,-%2D%20%2D%20%2D%20%2D%20%2D%20%2D%20%2D%20Set%20the%20flag) needed here? When will any of those `:watch` scripts return a `0` exit status?

package.json
jeffposnick, line 11: I think `maybe-sync-external` needs to be moved before `gulp` in this list. Right now, the `lhci` GitHub action is failing because `gulp` is run and there's no `external/data/.hash` present: https://github.com/GoogleChrome/developer.chrome.com/runs/6210054958?check_suite_focus=true#step:6:34

jeffposnick (changes requested):

jeffposnick (approved):

PR #2729, opened 2022-04-29
Add FAQ for OT opt-out

cucuandreea:
Fixes #SOME_ISSUE_NUMBER

- Add FAQ for region specific opt-out for OT

### DIFF
--- a/site/en/docs/privacy-sandbox/faq/index.md
+++ b/site/en/docs/privacy-sandbox/faq/index.md
@@ -141,6 +141,15 @@ For more detailed information, see
 
 Chrome on iOS and iPadOS does not support Chrome origin trials.
 
+### Can a site allow testing but opt-out of the OTs in specific geographic regions?
+
+A site can set a Permission Policy to control when an API can be accessed. The way 
+the Permission Policy works is that a site adds a little snippet of code to their 
+pages that provides instructions to the browser. When the page loads, the browser 
+reads the Permission Policy instructions and will allow or block features (or APIs)
+as outlined in the Permission Policy. So, if a site wanted to restrict an API in a 
+specific region, they could set a policy for all pages requested from that region.
+
 ## Trust Tokens
 
 ### How can I ask a question about this feature?


site/en/docs/privacy-sandbox/faq/index.md
```diff
@@ -141,6 +141,15 @@ For more detailed information, see
...
 
+### Can a site allow testing but opt-out of the OTs in specific geographic regions?
+
+A site can set a Permission Policy to control when an API can be accessed. The way 
+the Permission Policy works is that a site adds a little snippet of code to their 
+pages that provides instructions to the browser. When the page loads, the browser 
+reads the Permission Policy instructions and will allow or block features (or APIs)
+as outlined in the Permission Policy. So, if a site wanted to restrict an API in a 
+specific region, they could set a policy for all pages requested from that region.
+
```
heyawhite, line 152:
```suggestion
A site can set a [Permissions Policy](/docs/privacy-sandbox/permissions-policy/)
to control when an API can be accessed.

With a Permissions Policy, a site adds a little snippet of code to their 
pages that provides instructions to the browser. When the page loads, the browser 
reads the Permission Policy instructions and will allow or block features (or APIs)
as outlined in the Permissions Policy. If a site wants to restrict an API in a 
specific region, the developer could set a policy for all pages requested from that region.

```

heyawhite: Seems right to me. @kevinkiklee can you confirm?

kevinkiklee (changes requested):

site/en/docs/privacy-sandbox/faq/index.md
```diff
@@ -141,6 +141,17 @@ For more detailed information, see
 
 Chrome on iOS and iPadOS does not support Chrome origin trials.
 
+### Can a site allow testing but opt-out of the OTs in specific geographic regions?
```
heyawhite, line 144:
```suggestion
### Can a site allow testing but opt-out of the origin trials in specific geographic regions?
```

site/en/docs/privacy-sandbox/faq/index.md
```diff
@@ -141,6 +141,17 @@ For more detailed information, see
...
+
+A site can set a [Permissions Policy](/docs/privacy-sandbox/permissions-policy/)
+to control when an API can be accessed.
+
+With a Permissions Policy, a site adds a little snippet of code to their 
+pages that provides instructions to the browser. When the page loads, the browser 
+reads the Permission Policy instructions and will allow or block features (or APIs)
+as outlined in the Permissions Policy. If a site wants to restrict an API in a 
+specific region, the developer could set a policy for all pages requested from that region.
```
heyawhite, line 153:
```suggestion
In short, no, you cannot opt-out of an origin trial for specific regions.

However, a geographic-specific origin can set a
[Permissions Policy](/docs/privacy-sandbox/permissions-policy/)
to control what features are usable. For example, `us.example.com` and
`uk.example.com` are geographic-specific origins which can be controlled.
This does not mean that a region has opted-out of the origin trial.

With a Permissions Policy, a site adds a little snippet of code to their 
pages that provides instructions to the browser. When the page loads, the browser 
reads the Permission Policy instructions and will allow or block features (or APIs)
as outlined in the Permissions Policy. If a site wants to restrict an API in a 
specific region, the developer could set a policy for all pages requested from that region.

{% Aside 'warning' %}
Users may choose to visit an origin from a region that's different
from where they are. In other words, a user in the United States may be
able to visit `uk.example.com`. Those users would see features and functions
for the United States site that were blocked for the United Kingdom site.
{% endAside %}
```

heyawhite: @kevinkiklee does this accurately capture your feedback?

cucuandreea: @heyawhite @kevinkiklee 
Can we maybe just mention that by default OT will be active on all pages containing the token, but depending on their setup, site owners might be able to implement some logic to decide which pages will include the token?

```suggestion
In short, no, you cannot opt-out of an origin trial for specific regions.

By default, OT will be active on pages which contain the token, included 
via HTTP Headers (server side) or Meta Tags (front end). Site owners might 
build some logic on when the token should be included and geolocation may 
be a parameter they could use depending on their setup.

Alternatively, a geographic-specific origin can set a
[Permissions Policy](/docs/privacy-sandbox/permissions-policy/)
to control what features are usable. For example, `us.example.com` and
`uk.example.com` are geographic-specific origins which can be controlled.
This does not mean that a region has opted-out of the origin trial.
With a Permissions Policy, a site adds a little snippet of code to their 
pages that provides instructions to the browser. When the page loads, the browser 
reads the Permission Policy instructions and will allow or block features (or APIs)
as outlined in the Permissions Policy. If a site wants to restrict an API in a 
specific region, the developer could set a policy for all pages requested from that region.
{% Aside 'warning' %}
Users may choose to visit an origin from a region that's different
from where they are. In other words, a user in the United States may be
able to visit `uk.example.com`. Those users would see features and functions
for the United States site that were blocked for the United Kingdom site.
{% endAside %}
```

site/en/docs/privacy-sandbox/faq/index.md
```diff
@@ -141,6 +141,29 @@ For more detailed information, see
 
 Chrome on iOS and iPadOS does not support Chrome origin trials.
 
+### Can a site participate in origin trials but opt-out of using a feature in specific geographic regions?
+
+In short, no, you cannot opt-out of an origin trial for specific regions.
+
+However, a geographic-specific origin can set a
```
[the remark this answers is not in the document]
heyawhite: @cucuandreea rewrote your suggestion. Overall, sounds good! I intend to add and publish, with your final approval.

@kevinkiklee FYI

heyawhite (approved):
Once the final suggestion on tokens is added, this is ready to go.

site/en/docs/privacy-sandbox/faq/index.md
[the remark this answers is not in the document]
heyawhite: The logic doesn't check if the token is on the page, but wraps around to prevent sharing the token in certain geolocations?

As far as what `geolocation` is, thats a question for @cucuandreea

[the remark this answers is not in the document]
cucuandreea: Since OT is available on the pages that include the token I was trying to suggest the site owners could probably implement their own logic to decide which pages include it. The geolocation is referring to the browser's feature, yes, but there are other options as well like the IP.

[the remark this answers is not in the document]
kevinkiklee: Got it.  That makes sense.  Hmm, not sure if `geolocation` is the best recommendation because that pops up a user prompt, and not sure how the users would feel about seeing the permissions prompt pop up in some news site.  I think using IP address as an example here would a better approach. 

Let me know if the following paragraph is clearer: "If it's possible to determine the user's location in some way, then the developer can choose to include or not include the OT token header or meta tag based on that information. One way would be using the IP address to determine the location, but beware that IP addresses can be spoofed."  _(Also please clean up the grammar if required)_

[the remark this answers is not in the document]
heyawhite: Works for me. Will add / clarify / publish

PR #2730, opened 2022-04-29
[Fugu Showcase article] Make iframe embed linkable

tomayac:

- Make iframe embed linkable.

### DIFF
--- a/site/en/blog/fugu-showcase/index.md
+++ b/site/en/blog/fugu-showcase/index.md
@@ -14,6 +14,8 @@ tags:
   - capabilities
 ---
 
+## Motivation
+
 The Capabilities Project (code name [Project Fugu](/blog/fugu-status/)) is a cross-company effort with
 the objective of making it possible for web apps to do anything platform-specific apps can. We
 enable amazing web applications like [Photoshop](https://web.dev/ps-on-the-web/) by exposing the capabilities of
@@ -33,6 +35,8 @@ browsers, you can share any of the listed apps via the **Share app** feature. As
 Easter egg, the Project Fugu API Showcase is of course
 <a href="https://tomayac.github.io/fugu-showcase/data/#tomayac.github.io!fugu-showcase!data" target="showcase">contained
 in the Project Fugu API Showcase</a>. Happy browsing!
+
+## Showcase
 
 <div class="showcase-container" style="height: 100%; width: 100%">
   <noscript>
--- a/site/en/blog/fugu-showcase/index.md
+++ b/site/en/blog/fugu-showcase/index.md
@@ -2,12 +2,12 @@
 layout: 'layouts/blog-post.njk'
 title: Project Fugu API showcase
 subhead: >
-  The Project Fugu API Showcase is a collection of apps that make use of APIs that were conceived
-  in the context of Project Fugu.
+  The Project Fugu API Showcase is a collection of apps that make use of APIs that were conceived in
+  the context of Project Fugu.
 authors:
   - thomassteiner
 date: 2022-04-25
-updated: 2022-05-03
+updated: 2022-05-05
 hero: image/8WbTDNrhLsU0El80frMBGE4eMCD3/8FZcBmFowbDKWxpkOytx.jpg
 alt: Blowfish swarm swimming in the ocean.
 tags:
@@ -16,11 +16,11 @@ tags:
 
 ## Motivation
 
-The Capabilities Project (code name [Project Fugu](/blog/fugu-status/)) is a cross-company effort with
-the objective of making it possible for web apps to do anything platform-specific apps can. We
-enable amazing web applications like [Photoshop](https://web.dev/ps-on-the-web/) by exposing the capabilities of
-the underlying operating system platforms to the web platform, while maintaining user security,
-privacy, trust, and other core tenets of the web.
+The Capabilities Project (code name [Project Fugu](/blog/fugu-status/)) is a cross-company effort
+with the objective of making it possible for web apps to do anything platform-specific apps can. We
+enable amazing web applications like [Photoshop](https://web.dev/ps-on-the-web/) by exposing the
+capabilities of the underlying operating system platforms to the web platform, while maintaining
+user security, privacy, trust, and other core tenets of the web.
 
 But what are examples of some of the apps that make use of these capabilities? The Project Fugu API
 Showcase embedded below is sourced by community submissions and contains a filterable list of apps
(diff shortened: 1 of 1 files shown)


tomayac:
This lets people link to an individual API ([example API link](https://deploy-preview-2730--developer-chrome-com.netlify.app/blog/fugu-showcase?api=async-clipboard-images)) or to and individual app ([example app link](https://deploy-preview-2730--developer-chrome-com.netlify.app/blog/fugu-showcase?api=async-clipboard-images#editor.construct.net)).

tomayac:
(Note that the iframe will be migrated to the `GoogleChromeLabs` repo. Just awaiting the internal Ariane launch ticket to resolve.)

site/en/blog/fugu-showcase/index.md
```diff
@@ -34,6 +36,59 @@ Easter egg, the Project Fugu API Showcase is of course
...
+      src="https://tomayac.github.io/fugu-showcase/data/"
+      allow="web-share; clipboard-write; clipboard"
+    ></iframe>
+  </noscript>
 </div>
+
+<script>
+  const SHOWCASE_URL = 'https://tomayac.github.io/fugu-showcase/data/';
+  
+  const iframe = document.createElement('iframe');
```
jeffposnick, line 56: What's the advantage of creating the `iframe` imperatively instead of just using the version created via HTML? Are you worried about a timing issue around when the `load` event listener is added?

If so, you could consider:

- Using the `onload` attribute in HTML to invoke a function that's defined before the HTML.
- Just unconditionally calling `iframe.contentWindow.postMessage()` and relying on the fact that `postMessage()` will be [queued](https://www.w3.org/TR/2021/NOTE-workers-20210128/#:~:text=workers%20has%20its-,port%20message%20queue,-implicitly%20enabled%20when) until the document being posted to has fully loaded.

It would be cleaner to avoid defining the `iframe`'s in both HTML (for `<noscript>`) and JavaScript.

jeffposnick (changes requested):
I understand the limitations that led to this blog post getting progressively more complicated, to the point where there's extensive `postMessage()` communication going on.

I just want to make @devnook is also cool with this level of complexity being moved into inline JavaScript on a blog post page before we move ahead.

site/en/blog/fugu-showcase/index.md
```diff
@@ -34,6 +36,59 @@ Easter egg, the Project Fugu API Showcase is of course
...
+  iframe.src = SHOWCASE_URL;
+
+  window.addEventListener('message', (event) => {
+    if (event.origin !== new URL(SHOWCASE_URL).origin) {
+      return;
+    }
+    const url = new URL(window.location);
+    if ('search' in event.data) {
+      if (event.data.search) {
+        const [key, value] = event.data.search.split('=');
```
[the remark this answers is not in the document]
tomayac: I think the overhead in this case wouldn't be worth it. It literally just expects `foo=bar`.

site/en/blog/fugu-showcase/index.md
tomayac: I used `<iframe onload>`, but the only way I found to make the `Window.postMessage()` call to work was via `frames[0]`, which sounds very brittle should we add more frames. The intuitive `this.contentWindow.postMessage()`  didn't work. So I came up with the present `<noscript>` and scripted approach, which works universally.

jeffposnick: Apologies, but I am missing some nuance here. Which `Window.postMessage()` call are you talking about? How does `frames[0]` factor in?

site/en/blog/fugu-showcase/index.md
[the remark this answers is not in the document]
jeffposnick: `URLSearchParams`'s `get()` will decode URL entities, while `split()` obviously won't.

You're already using `URLSearchParams`'s `set()`, which will encode URL entities. If you don't pass in something that's been previously decoded to `set()`, you can end up with double-encoding.

E.g. you'd see a difference with `key=value+with+spaces`.

site/en/blog/fugu-showcase/index.md
tomayac: I managed to get it working with the `onload` attribute. PTAL.

site/en/blog/fugu-showcase/index.md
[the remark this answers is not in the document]
tomayac: Using `URLSearchParams` now.

jeffposnick (approved):
Thanks for those changes, Tom.

I'm fine with this implementation, but it would be good to make sure that the added in-page complexity is cool with @rachelandrew and/or @devnook before we merge.

site/en/blog/fugu-showcase/index.md
```diff
@@ -34,6 +36,40 @@ Easter egg, the Project Fugu API Showcase is of course
...
+    title="Fugu showcase"
+    name="showcase"
+    style="min-height: 1000px;width: 100%;border: solid 1px var(--color-hairline);"
+    src="https://tomayac.github.io/fugu-showcase/data/"
+    allow="web-share; clipboard-write; clipboard"
+    onload="this.contentWindow.postMessage({hash:location.hash.substr(1),search:location.search.substr(1)},'*');"
+  ></iframe>
 </div>
+
+<script>
```
devnook, line 52: I'm rather sure this will violate CSP policy

site/en/blog/fugu-showcase/index.md
```diff
@@ -34,6 +36,40 @@ Easter egg, the Project Fugu API Showcase is of course
...
+  const SHOWCASE_URL = 'https://tomayac.github.io/fugu-showcase/data/';
```
devnook, line 53: I am strongly against embedding scripts like this in an .md file. Let's keep it for now for the sake of upcoming event needs but please let's reimplement it in a more sane way asap afterwards (as discussed offline). 

@tomayac Please add a TODO: Please remove in favor of a standalone solution (one with full js capabilities).

devnook (approved):
I'm approving, but please see comments.

site/en/blog/fugu-showcase/index.md
tomayac, line 52:
```suggestion
<!-- TODO: @tomayac
  Remove this inline script once a stand-alone has been implemented.
--> 
<script>
```

site/en/blog/fugu-showcase/index.md
tomayac: Done

site/en/blog/fugu-showcase/index.md
tomayac: Perfect then, resolving.

site/en/blog/fugu-showcase/index.md
```diff
@@ -2,23 +2,25 @@
...
 subhead: >
-  The Project Fugu API Showcase is a collection of apps that make use of APIs that were conceived
-  in the context of Project Fugu.
+  The Project Fugu API Showcase is a collection of apps that make use of APIs that were conceived in
+  the context of Project Fugu.
 authors:
   - thomassteiner
 date: 2022-04-25
-# updated: 2022-04-21
+updated: 2022-05-05
```
tomayac, line 10:
```suggestion
updated: 2022-05-09
```

PR #2753, opened 2022-05-03
Add a mini TOC to the SXG preamble.

twifkak:
In addition to providing links (as already covered by the TOC), it provides
short summaries of the salient points of each section.

@malchata WDYT? My goal is to summarize the actionable steps to help the reader decide which section to read for more details.


malchata (approved):
Adding a mini-ToC seems like a nice way to break down what the user is about to learn. LGTM!

PR #2762, opened 2022-05-04
DevTools 98 > Application > Back/forward cache

sofiayem:
Documented https://developer.chrome.com/blog/new-in-devtools-98/#bfcache

Also, shuffled article ordering in the section to more resemble the UI tab order.


jecfish (approved):

jecfish:
Approve content wise. I am thinking about the title.

We name all our panel article to start with [panel name:] prefix. Should we do the same for all the Application tab and other pane (e.g. Styles pane, Properties pane, etc) too?

PR #2790, opened 2022-05-09
[Fugu Showcase] Update self link

tomayac:

- Update self link.

### DIFF
--- a/site/en/blog/fugu-showcase/index.md
+++ b/site/en/blog/fugu-showcase/index.md
@@ -33,7 +33,7 @@ You can launch each app by clicking the app's name, the app's preview image, or 
 link. For many apps, you can also see the source code by clicking **Source code**. On supporting
 browsers, you can share any of the listed apps via the **Share app** feature. As a special inception
 Easter egg, the Project Fugu API Showcase is of course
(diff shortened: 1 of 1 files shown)


tomayac:
> If I'm at https://deploy-preview-2790--developer-chrome-com.netlify.app/blog/fugu-showcase/ and click on the "contained in the Project Fugu API Showcase" link, the URL updates to include the hash, but nothing scrolls into view or changes in the `<iframe>` embed.
> 
> If I navigate directly to https://deploy-preview-2790--developer-chrome-com.netlify.app/blog/fugu-showcase/#tomayac.github.io!fugu-showcase!data then the showcase does scroll into view when the page loads.
> 
> Is there a race condition or some missing code that prevents the same-page hash transition from being picked up and triggering the scroll-into-view?

The iframe only checks for the hash of the embedding page `onload`. Since we discussed with @devnook that we want to move away from the iframe and get this properly supported as a first-party feature on dcc, there's no need to add even more complexity and communicate to the iframe when the hash changes, which could technically be done. The quick and dirty fix in https://github.com/GoogleChrome/developer.chrome.com/pull/2790/commits/628bb36a2804f2de86ef7fe71d5b3d8ec6c9905c is to open the link in a new window.

tomayac:
Thanks, @jeffposnick, for detecting the problem.

PR #2820, opened 2022-05-12
Add article content type

devnook:
This change aims at introducing new content type for dcc, called Article. It will be used for most of the articles that now get published under blog posts (long form, maintained posts). We will redirect the blog post type to represent short form, time-aware content (no maintenance obligation).

Articles create a separate collection for each locale: articles-en, articles-de etc (see .eleventy.js).

RSS Feeds:
Blog gets it's own feed as before, and Articles are included in the /feeds/all.xml feed.

Algolia and search:
Articles will be listed as a separate results list than Blog and Docs. It appears now after Blog, but once we got more content published as Article we might swap the order.

 As a next step, we might rename a bunch of "blog-" named parials and templates, to make them more universal (e.g. so that they can be used also for articles without confusion). I'd like to do it in a separate PR though.


jeffposnick (approved):
No blocking comments 👍 

One meta-comment is that there are a couple of places in which an article references `layouts/blog-post.njk` and `layouts/blog-landing.njk`. We should think about two scenarios:

- If blogs and articles will always have the same layouts/landing pages, then we should probably rename those templates something like `blog-article-landing.njk` to make it clear that they're being shared.
- If we expect that we will use a different layout at some point in the future, then we should probably make copies of those `blog-*` templates and name them `article-*`. It's fine if they are copies of the same content right now, as they might diverge in the future.

I don't know what the right approach is for now, so we can decide and then move forward with that either in this PR, or a future PR.

PR #2384, opened 2022-03-22
Add MV3 samples to Design the User Interface

AmySteam:
Fixes #2382 

Changes proposed in this pull request:

- Restructure content
- Add links to MV3 examples
- Add link to action demo example

TODO:
- [ ] Add links to MV3 example repos when PRs are merged

https://github.com/GoogleChrome/chrome-extensions-samples/pull/678
https://github.com/GoogleChrome/chrome-extensions-samples/pull/679
https://github.com/GoogleChrome/chrome-extensions-samples/pull/680

PR #2825, opened 2022-05-12
[crx-docs] Fix columns and link APIs in Design UI

AmySteam:

- Fix how columns titles. Currently they are not rendering correctly
- Change format for link APIs. Current format doesn't look very good. Trying another second approach (See [convo here](https://github.com/GoogleChrome/developer.chrome.com/pull/2384#discussion_r851563894)).

Staged links:
[Design UI](https://deploy-preview-2825--developer-chrome-com.netlify.app/docs/extensions/mv3/user_interface/)

### DIFF
--- a/site/en/docs/extensions/mv3/user_interface/index.md
+++ b/site/en/docs/extensions/mv3/user_interface/index.md
@@ -485,7 +485,7 @@ To use the [Notifications API][api-notif], you must declare the `"notifications"
 the manifest.
 
 ```json/5
-//manifest.json
+// manifest.json
 { 
   "name": "Drink Water Event Popup",
 ...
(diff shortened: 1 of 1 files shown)


dotproto (approved):
Thanks for cleaning this up, @AmySteam! I have two minor cleanup suggestions; feel free to merge when those are integrated.

I'm mostly thinking out loud here, but most of the changes in this PR are converting from one naming convention for referencing APIs to another. At the moment our docs use a couple (or a few?) different conventions, so I appreciate the desire to establish some rigor. The only hesitation I have with this PR (but not a significant enough one to block it) is that we don't currently have an established preference for which conventions to use and when to use them.

I'd like to try to sort this out with the product team, tech writers, and other relevant stakeholders. To that end, I'm going to start putting together some doc for discussion purposes. I'll follow up with you about this offline.

rachelandrew:
@dotproto just looking at this, I think the convention using in this PR is more consistent with how we generally refer to APIs in other documentation. So "Action API", not "`action` API".

Ping me on the doc for cleaning up these conventions, as it would be good to align with other product areas.

PR #2829, opened 2022-05-13
Add opt-out information moved from /docs/privacy-sandbox/fledge

samdutton:
It makes more sense to have technical detail in this article.

**Content is moved, not new. Shouldn't need editing or checking.**


heyawhite (approved):

samdutton:
Thanks Alexandra!

PR #2569, opened 2022-04-11
Updates to origin trials articles

samdutton:

PR #2865, opened 2022-05-19
Move Origin Trial documentation

rachelandrew:
As part of the debloggification of documentation moving the OT docs into the new docs/web-platform section and redirecting.

- After doing my test move some updates had been made to the originals in https://github.com/GoogleChrome/developer.chrome.com/pull/2569 so merged those in.
- Fixed links.
- Removed docs from blog
- Move the jp translations
- Added redirects.


rachelandrew:
@samdutton The only changes I've made to the content is to update some links, plus there was a mention that we provide docs on web.dev, which isn't true any more. Future docs (and old ones once I've finished moving them) will be in this docs section. However let me know if you spot any issues.

samdutton (approved):
LGTM  — thank you!

PR #2871, opened 2022-05-20
Updated `exclude`'s href to `exclude`

avi12:
Fixes the link to `@exclude`
image


AmySteam:
@avi12 Thanks for this! :) 👍 

@dotproto LGTM! The link is correct, we can merge this

sofiayem (approved):

PR #2887, opened 2022-05-24
Migrate Fugu Showcase from iframe to proper inline

devnook:
Migrate the Fugu Showcase to be a proper page.


tomayac:
@devnook This starts to take shape, here's the [deploy preview](https://deploy-preview-2887--developer-chrome-com.netlify.app/fugu-showcase#) (there are some issues since the deploy doesn't run `npm i`, so for example the IDs are missing). It looks fine on `localhost`. Also the images aren't appearing i the deploy preview.

localhost_8080_fugu-showcase_

I'd appreciate some help with the cards. If I manually force the following CSS, it works so that there are two side-by-side.

```css
@media (min-width: 592px)
.blog-grid {
  gap: 2rem;
  /* 18.625rem is the magic number */
  grid-template-columns: repeat(auto-fit, 18.625rem);
 ```

Screen Shot 2022-06-22 at 16 36 27

tomayac:
@devnook We have talked about uploading the image files to the CDN. I'm no longer sure this is the best solution. The reason is that upon each rerun the images will have changed slightly, so we end up with a lot of out-of-use images. It may be better to have them checked in directly and overwrite them.

tomayac:
@devnook and @matthiasrohmer: If we were to host the app screenshots on [`GoogleChromeLabs/fugu-showcase/tree/main/data`](https://github.com/GoogleChromeLabs/fugu-showcase/tree/main/data) (so they don't pollute the `GoogleChrome/developer.chrome.com` repo) and then reference the images from the showcase article at `developer.chrome.com/fugu-showcase/`, would this be a viable solution to unlock this PR?

`` 



(CC: @rachelandrew)

tomayac:
Closed as completed via #4362

PR #2898, opened 2022-05-25
Update to 11ty v1.0.1

jeffposnick:
This is a clean PR that includes the changes from https://github.com/GoogleChrome/developer.chrome.com/pull/2488, but with 11ty v1.0.1 instead of the canary builds.


jeffposnick:
There are a few visual diffs in Percy because the updated `@11ty/eleventy-plugin-syntaxhighlight` dependency will add syntax highlighting to the property keys in a JavaScript object, which is a change that we'd want anyway.

devnook (approved):

Issue #2798, opened 2022-05-10
broken link


Thalimenel:
**Describe the bug**
Broken link to Match Patterns in [create](https://developer.chrome.com/docs/extensions/reference/contextMenus/#method-create)
it leads to: https://developer.chrome.com/docs/extensions/match_patterns
it should lead to: https://developer.chrome.com/docs/extensions/mv3/match_patterns/

**To Reproduce**
1. Go to https://developer.chrome.com/docs/extensions/reference/contextMenus/#method-create
2. Under "createProperties"
3. See the "documentUrlPatterns" property
4. Click on the "[Match Patterns](https://developer.chrome.com/docs/extensions/match_patterns)" link. It will lead to a 404 page
5. See error

**Expected behavior**
Should go to https://developer.chrome.com/docs/extensions/mv3/match_patterns/

AmySteam:
@Thalimenel Thanks for letting us know! 👍 I just made a PR that should fix this broken link :)

Thalimenel:
👍 :smiley:

PR #2928, opened 2022-05-30
[crx-doc] Add new group redirect to extension docs

### LINKED ISSUE #2924: The link of Manifest File is missing in the documentation
⚠️ This is for issues with the https://developer.chrome.com site, not Chromium itself.
If you want to file a bug with Chromium (the open-source project behind Google Chrome and other browsers), please use https://crbug.com.

**Describe the bug**
On this [page](https://developer.chrome.com/docs/extensions/reference/runtime/#method-getManifest) I can see a link to the _manifest file_, but that link doesn't exist.

**The screenshot of link
image

Thanks,
Farzad

AmySteam:
Fixes #2924 #2798

- Add another redirect group to fix broken links. The following are a few examples:

[/docs/extensions/messaging](https://developer.chrome.com/docs/extensions/messaging) should redirect
to: [/docs/extensions/mv3/messaging](https://developer.chrome.com/docs/extensions/mv3/messaging)

[/docs/extensions/manifest](https://developer.chrome.com/docs/extensions/manifest) should redirect to
[/docs/extensions/mv3/manifest](https://developer.chrome.com/docs/extensions/mv3/manifest)


AmySteam:
@dotproto I can't get this redirect or any of the existing redirects to work in the Netlify Deploy Preview link.

I can demo this working code when we meet this week.

redirects.yaml
```diff
@@ -354,6 +354,15 @@ redirects:
     - /docs/apps/manifest/...
   else: /docs/extensions/...
 
+- from: /docs/extensions/...
+  try:
+    - /docs/extensions/reference/...
+    - /docs/extensions/mv3/manifest/...
+    - /docs/extensions/mv3/...
```
dotproto, line 361: I think we should also include MV2 docs in the fallback check. There are some pages that aren't in the MV3 doc set but are in the MV2 and apps one. In those cases, it would be best to load the "extensions" version of the page before falling back to the app docs.

```suggestion
    - /docs/extensions/mv3/...
    - /docs/extensions/mv2/manifest/...
    - /docs/extensions/mv2/...
```

dotproto (approved):
LGTM once the suggestion is merged.

Just thinking out loud here, but this PR makes me wish we had a tool that would crawl the docs and create a diffable list of the links on each page. 🤔

PR #2957, opened 2022-06-04
Update article on Inert

ebidel:
The default is false. Also not observing that nested attributes take precedence since this is a boolean attribute.


jpmedley:
@twerske Does this look good to you?

jpmedley:
It will take all of one minute. The submitter previously worked for us, so it's probably good.

rachelandrew (approved):
I've checked the spec and this change is correct. https://html.spec.whatwg.org/multipage/interaction.html#inert

rachelandrew:
thanks @ebidel

PR #2961, opened 2022-06-05
Correct browser.tabs permission requirements

RedHotRojo:
Developers MUST use the host permission. The tabs permission is optional. Source: personal experience (trust me bro); I was able to access tab.url, tab.pendingUrl, tab.title, and tab.favIconUrl without the tabs permission. Using just the tabs permission, the console complained that I didn't have host permissions. With just the host permission and no tabs permission, I can access tab.url, tab.pendingUrl, tab.title, and tab.favIconUrl.

(I did not create an issue for this pull request; the issue is described above)

- Tell developers that they must use host permissions instead of tabs permission
- Change the example manifest to match above

### DIFF
--- a/site/en/docs/extensions/reference/tabs/index.md
+++ b/site/en/docs/extensions/reference/tabs/index.md
@@ -8,15 +8,15 @@ extra_permissions_html:
 
 You can use most `chrome.tabs` methods and events without declaring any permissions in the
 extension's [manifest][manifest] file. However, if you require access to the `url`, `pendingUrl`,
-`title`, or `favIconUrl` properties of [`tabs.Tab`][tab], you must declare the `"tabs"` permission
+`title`, or `favIconUrl` properties of [`tabs.Tab`][tab], you must declare the `"host"` permission
 in the manifest, as shown below:
 
 ```json
 {
   "name": "My extension",
   ...
-  "permissions": [
-    "tabs"
+  "host_permissions": [
+    "http://www.blogger.com/"
   ],
   ...
 }


AmySteam:
@RedHotRojo Thanks for this PR! I'm curious about what you are running into

> Using just the tabs permission, the console complained that I didn't have host permissions.

Were you trying to use `tabs.executeScript()` on an MV2 extension? or is this an MV3 extension?

I'm curious... could you share **your code and maybe a screenshot** of the console error? 🙂

I'm currently working on a PR for the **Tabs API**, which includes a detailed explanation of the permissions needed. And you are right that if you are already declaring broad host permissions, you do not "need" the tabs permission to access the tab.url, tab.pendingUrl, tab.title, and tab.favIconUrl.

RedHotRojo:
@AmySteam Yes, I was using `executeScript`, so the extension probably required the host permission there. And I am using MV3.

I created a simple 3-line service worker that logs every tab's url, tapendingUrl, title, and favIconUrl:
```
chrome.tabs.onUpdated.addListener((tabId, changeInfo, tab) => {
  console.log(tab.url, tab.pendingUrl, tab.title, tab.favIconUrl);
});
```
This code spit out `undefined` for each value when there were no permissions declared in the manifest:
```
undefined undefined undefined undefined
```
But, with **either** the `tabs` permission or the respective tab's URL host permission, the URL, title, and favIconUrl were logged with values. However, pendingUrl remained at undefined: 
```
https://www.google.com/ undefined Google https://www.google.com/favicon.ico
```
Essentially, you could inform developers to continue using host permissions if they already have them declared or use the tabs permission instead. It would also help if you told them what exactly the Chrome Web Store will alert them if either one is used. Anyway, you can close this PR if you are creating your own.

site/en/docs/extensions/reference/tabs/index.md
```diff
@@ -8,15 +8,15 @@ extra_permissions_html:
...
+`title`, or `favIconUrl` properties of [`tabs.Tab`][tab], you must declare the `"host"` permission
 in the manifest, as shown below:
 
 ```json
 {
   "name": "My extension",
   ...
-  "permissions": [
-    "tabs"
+  "host_permissions": [
```
dotproto, line 18: Please revert this change. We want to encourage developers to use the least permissions possible in their extensions. In this case, we are only demonstrating how to access the properties that tabs exposes. It would be better to update the intro to this code sample in order to better set the reader’s expectations than to change the code sample itself.

dotproto (changes requested):
The behavior @RedHotRojo described in their last comment is correct: in order to access the ‘url’, ‘pendingUrl’, ‘title’, and ‘favIconUrl’ of a tab the extension must either have the tabs permission or host permissions for that tab. As @AmySteam seemed to suggested, I suspect that when “the console complained that I didn't have host permissions” the error was in reference to an ‘executeScript()’ call. This method _does_ require host permissions.

I agree that we could do with an update to better explain the difference between host permissions, the ‘tabs’ permissions, and content scripts match patterns, but the PR as is doesn’t accurately represent the extensions platform.

RedHotRojo:
@dotproto Yes, a better explanation of each permission would work better. Developers can then make their choice of what permissions they want to use. Since @AmySteam is working on a PR, I will close this one and allow her to use this thread to better streamline all of our ideas into one commit.

PR #2969, opened 2022-06-06
Fix code style: remove unreachable break statements

theghostbel:
Fixes code style: removed unreachable "break" statements - fast-return makes them obsolete.


malchata:
@jeffposnick, would you mind verifying this? I think this is a fine change, but wanted your perspective.

jeffposnick (approved):

malchata (approved):

PR #2986, opened 2022-06-09
DevTools 98 > Mention endpoints in Reporting API

sofiayem:
Documented https://developer.chrome.com/blog/new-in-devtools-98/#reporting-api


site/en/docs/devtools/javascript/background-services/index.md
```diff
@@ -170,6 +170,8 @@ To view the reports generated by a page:
 
 1. Click on a report to see its body.
 
+Additionally, the **Reporting API** gives you an overview of all the endpoints configured in the `Reporting-Endpoints` header.
```
jecfish, line 173: Probably it should be a pointer too. Or, we should expand the last point "Click on a report to see its body" to explain the 3 sections of the report

jecfish (changes requested):

jecfish:
1. I think we should change the url of this document. It should not be under `Javascript`. Probably `application/` ? Also maybe reorganize and redirect other application panel url as well. Can be another PR.
2. Just realize the screenshot of all the other section in this docs is obsolete. There are more background services undocumented. 🙈

sofiayem:
> 1. I think we should change the url of this document. It should not be under `Javascript`. Probably `application/` ? Also maybe reorganize and redirect other application panel url as well. Can be another PR.
> 2. Just realize the screenshot of all the other section in this docs is obsolete. There are more background services undocumented. 🙈

1. Yeah. On my TODO list.
1. That too.

:)

PR #3028, opened 2022-06-15
Fix a missing "i" in hero image URL

jeffposnick:
This resolves an issue that was breaking our build.


jeffposnick:
Admin-merging in the interest of fixing the build.

Issue #1755, opened 2021-11-18
Improve hero image layout?


maudnals:
**Is your feature request related to a problem? Please describe.**
The hero image doesn't look aligned with the text.
See screenshot ⤵️
![image](https://user-images.githubusercontent.com/9762897/142498322-e34414ff-8d70-4eca-b450-55f07ff47238.png)

**Describe the solution you'd like**
The hero image is aligned with the text.
See screenshot ⤵️
![image](https://user-images.githubusercontent.com/9762897/142498635-ad2811b2-31dd-4237-a3fa-7c497b33ea00.png)

PR #3080, opened 2022-06-23
Remove .toc-container width to fix article body centering

### LINKED ISSUE #2747: Hero image is strangely off-balance with large viewports (>=`1440px`)
Note how the hero image is strangely position too far on the right. This happens when the screen hits the `1440px` media query.

Screen Shot 2022-05-03 at 11 53 20
Screen Shot 2022-05-03 at 11 53 30

jeffposnick:
Fixes #2747 by removing the explicit width on the `.toc-container` element. This should shift the overall article body to the right, which then makes everything look properly centered. (The issue was not with the hero image being off-center, it was with the article body.)

CC: @tomayac and @maudnals to check the pages they reported issues with and confirm that the new layout matches what they'd expect.

### DIFF
--- a/site/_scss/blocks/_toc.scss
+++ b/site/_scss/blocks/_toc.scss
@@ -1,6 +1,5 @@
 .toc-container {
   max-width: 100%;
-  width: px-to-rem(320px);
 }
 
 .toc > summary {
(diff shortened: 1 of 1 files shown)


tomayac (approved):
LGTM. This makes the pages I negatively noticed this issue on look way more balanced.

maudnals:
The page I reported the issue on in #1755 isn't live anymore, but I'm confident the issue I filed was the same as what @tomayac observed on #2747.

Thank you Jeff for fixing this!

PR #3111, opened 2022-06-29
Show that "persistent" value is set to true by default

solomonkinard:
Clarity is the key here. It should be completely obvious what the default value is. This is especially important because so many other undefined values make a lot of sense as false by default.


dotproto (approved):
Thanks, @solomonkinard!

PR #3150, opened 2022-07-07
Update description for Fenced Frames cross-site data access

kevinkiklee:
The existing documentation implies that Fenced Frames have general access to unpartitioned data, which is confusing.  A better explanation has been drafted to show how cross-site data works with fenced frames.


heyawhite (approved):
LGTM % suggestion

PR #3197, opened 2022-07-12
Update dependencies to pick up yaml-redirect bugfix

dotproto:
Version 2.0.4 of `yaml-redirect` allows the server to redirect requests to a given anchor on a page.


jpmedley:
@devnook I know Simeon's not here any more, but this is something we can use. 

cc: @rachelandrew

matthiasrohmer:
We updated to 2.0.4 in the meantime.

PR #3214, opened 2022-07-13
Fetch streaming request article move & update

jakearchibald:
Fixes https://github.com/GoogleChrome/web.dev/issues/8288.
Previous PR https://github.com/GoogleChrome/web.dev/pull/8317.
Preview https://deploy-preview-3214--developer-chrome-com.netlify.app/docs/web-platform/fetch-streaming-requests/

@rachelandrew I'm not sure which index to add this to. It's shipping in 105, so should it go in `i18n.docs.web-platform.shipped`?

Also, should I be referring to Chrome or Chromium in this post? It will work in other Chromium browsers, so I guess Chromium is more appropriate?


rachelandrew:
@jakearchibald maybe we should add a beta section? I think if we say things are shipped from a dev pov they might expect that to be in stable.

As for Chromium vs. Chrome, Chromium likely works best especially on this site. On web.dev I'd rather talk about browsers than engines as that maps to the compat data etc. but I think here we can be more accurate. I need to clarify this as part of the terminology stuff I've been discussing with people.

site/en/docs/web-platform/fetch-streaming-requests/index.md
```diff
@@ -0,0 +1,269 @@
...
+description: >
+  Chrome now supports upload streaming as of version 95, which means
+  you can start a request before you have the whole body available.
+date: 2020-07-22
+updated: 2022-07-13
+tags:
+  - capabilities
+---
+
+From Chrome 105, you can start a request before you have the whole body available by using the streams API.
```
rachelandrew, line 15:
```suggestion
From Chrome 105, you can start a request before you have the whole body available by using the [Streams API](https://developer.mozilla.org/docs/Web/API/Streams_API).
```

rachelandrew (changes requested):
Is this really docs? This feels more like an article about a new feature for this API.

How much of this isn't documented on MDN for example? I wrote a lot of the Streams documentation. If it's a case of we're just missing this capability we could go and add it to the docs. I'm thinking this probably wants moving to the articles section here, and if it's not a lot of work to add the new capability to MDN let's do that.

Is this detailed in BCD?

jakearchibald:
> @jakearchibald maybe we should add a beta section? I think if we say things are shipped from a dev pov they might expect that to be in stable.

That works for me!

> Is this really docs? This feels more like an article about a new feature for this API.

I'm not really clear on the difference. Happy to put it wherever you think. But yeah, it's an article about a new feature in `fetch()`.

> How much of this isn't documented on MDN for example? I wrote a lot of the Streams documentation. If it's a case of we're just missing this capability we could go and add it to the docs. I'm thinking this probably wants moving to the articles section here, and if it's not a lot of work to add the new capability to MDN let's do that.
> 
> Is this detailed in BCD?

There are significant missing parts in MDN or BCD. Happy to add them.

What's the right way round to do it? MDN and BCD both reference each other, so I guess I'd need to create PRs for both, then PRs for both again to add the cross-references?

jakearchibald:
I've moved it to articles. Let me know if I haven't done it correctly.

PR #3221, opened 2022-07-14
[JA] Translate Progress in the Privacy Sandbox 2022-06 into Japanese

ilyaspiridonov:
Translate Progress in the Privacy Sandbox 2022-06 into Japanese


ilyaspiridonov:
@agektmr FYI:
I also wanted to report a small issue, which is specific to Japanese: 
- The English text extensively uses bold text formatting (**)
- In Japanese those asterisks will leak into the rendered document: 

Screenshot 2022-07-14 at 10 33 14

- Reason: there are no spaces around the bold markup (**)
- I'm correcting this by adding extra spaces, but this may look un-natural in Japanese (as I understand, spaces should not be used there). 

Screenshot 2022-07-14 at 10 54 49


Please confirm whether those extra spaces are OK. 
My understanding is this is the way the static site generator currently renders md->htlml.

agektmr:
@ilyaspiridonov adding extra spaces for these cases sounds good to me. Please do the same for `__` as well.

agektmr (changes requested):

agektmr (approved):
LGTM

PR #3258, opened 2022-07-19
[CRX] Refactor permissions warning

AmySteam:
## Problem
APIs that don't have permission warnings display an empty "Has warning" metadata:
Screen Shot 2022-07-19 at 11 51 15

### APIs with no permission warning

Do not display a message. 

Screen Shot 2022-07-19 at 15 08 00

### APIs with one permission warning

Display message in new line

Screen Shot 2022-07-19 at 15 07 38

### APIs with multiple permissions with or without warnings

Display message in new line

Screen Shot 2022-07-19 at 15 26 23

## Staged Links

- [Alarms API](https://deploy-preview-3258--developer-chrome-com.netlify.app/docs/extensions/reference/alarms)
- [Bookmarks API](https://deploy-preview-3258--developer-chrome-com.netlify.app/docs/extensions/reference/bookmarks)
- [DeclarativeNetRequest API](https://deploy-preview-3258--developer-chrome-com.netlify.app/docs/extensions/reference/declarativeNetRequest)

### DIFF
--- a/site/_includes/layouts/namespace-reference.njk
+++ b/site/_includes/layouts/namespace-reference.njk
@@ -79,19 +79,7 @@
                     <code>{{ permission }}</code><br />
                   {% endfor %}
                   {{ extra_permissions_html | safe }}
-                </div>
-              </li>
-              <li>
-                <div class="code-sections__label">Has warning?</div>
-                <div>    
                   {{ has_warning | safe }}
-                </div>
-              </li>
-            {% else %}
-            <li>
-                <div class="code-sections__label">Permission</div>
-                <div>
-                {{ special_permissions_html | safe }}
                 </div>
               </li>
             {% endif %}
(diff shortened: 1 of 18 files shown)


site/_includes/layouts/namespace-reference.njk
```diff
@@ -79,21 +79,9 @@
                     <code>{{ permission }}</code><br />
                   {% endfor %}
                   {{ extra_permissions_html | safe }}
-                </div>
-              </li>
-              <li>
-                <div class="code-sections__label">Has warning?</div>
-                <div>    
                   {{ has_warning | safe }}
```
dotproto, line 82: The current implementation of `extra_permissions_html` and `has_warning` implicitly requires that page authors manually handle line breaks between these sections. It would be better to treat each block of HTML as as completely independent in order to minimize the chance of rendering bugs. In this case, we should conditionally insert a `<br>` after `extra_permissions_html` if `has_warning` is not empty.

dotproto (approved):
I have one concern regarding the interaction of `extra_permissions_html` and `has_warning` in the `namespace-reference.njk` template, but we don't need to block merging this change on that.

PR #3295, opened 2022-07-25
Allow custom <video> aspect ratio

jakearchibald:
This is part of https://github.com/GoogleChrome/webdev-infra/issues/28, to allow custom aspect ratio on `<video>` and reduce CLS.


jakearchibald:
I'm not sure how to interpret the lighthouse failure here.

site/_scss/_global.scss
```diff
@@ -23,5 +23,9 @@ main:focus {
 }
 
 video {
+  // These custom properties are set by the Video shortcode if both the width
+  // and height attributes are set.
+  aspect-ratio: var(--vid-width) / var(--vid-height);
```
bramus, line 28: What if these attributes are not set? Should be add fallback values or simply allow this value to become invalid?

jakearchibald: The fallback would be the default object size of 300x150 right?

jakearchibald: Falling back to invalid, which is a no-op, is probably more compatible with existing content on the site.

jakearchibald:
I haven't tested this btw, so it'll need an eyeball to check that it works with existing video content.

PR #3306, opened 2022-07-26
[JA] Update  translation for Proposal Lifecycle

ilyaspiridonov:
Updates Japanese translation for the Proposal Lifecycle article


agektmr (approved):
LGTM

PR #3330, opened 2022-07-28
Correct mention of web standard in background-fetch

rachelandrew:
Reported to me by @cwilso, the Background Fetch article incorrectly refers to the spec as a web standard.

I've updated the article, added a correction note at the bottom, I also fixed some links that now redirect due to the Web Fundamentals migration and removed outdated origin trial info while I was in there. 

Preview: https://deploy-preview-3330--developer-chrome-com.netlify.app/blog/background-fetch

### DIFF
--- a/site/en/blog/background-fetch/index.md
+++ b/site/en/blog/background-fetch/index.md
@@ -21,7 +21,7 @@ date: 2018-12-03
 
 # Optional
 # Include an updated date when you update your post
-updated: 2020-12-17
+updated: 2022-07-28
 
 # Optional
 # How to add a new tag
@@ -31,7 +31,7 @@ updated: 2020-12-17
 
 ---
 
-In 2015 we [introduced Background Sync](https://developers.google.com/web/updates/2015/12/background-sync) which allows the
+In 2015 we [introduced Background Sync](/blog/background-sync/) which allows the
 service worker to defer work until the user has connectivity. This means the user could type a
 message, hit send, and leave the site knowing that the message will be sent either now, or when they
 have connectivity.
@@ -42,8 +42,9 @@ too long the browser will kill the service worker, otherwise it's a risk to the 
 battery.
 
 So, what if you need to download something that might take a long time, like a movie, podcasts, or
-levels of a game. That's what Background Fetch is for. Background Fetch is a [web
-standard](https://wicg.github.io/background-fetch/) available by default since Chrome 74.
+levels of a game. That's what [Background Fetch](https://developer.mozilla.org/docs/Web/API/Background_Fetch_API) is for. 
+
+Background Fetch is available by default since Chrome 74.
 
 Here's a quick two minute demo showing the traditional state of things, vs using Background Fetch:
 
@@ -51,11 +52,7 @@ Here's a quick two minute demo showing the traditional state of things, vs using
 
 
 [Try the demo yourself](https://bgfetch-http203.glitch.me/) and [browse the
-code](https://glitch.com/edit/#!/bgfetch-http203?path=public/client.js). It requires Chrome 71, and
-the *Experimental Web Platform features* flag to be enabled.
-
-This is also being run as an Origin Trial. If you're interested in testing this API with real users
-without a flag, [see below](#origin-trial).
+code](https://glitch.com/edit/#!/bgfetch-http203?path=public/client.js). 
 
 ## How it works
 
@@ -92,7 +89,7 @@ if ('BackgroundFetchManager' in self) {
 
 ### Starting a background fetch
 
-The main API hangs off a [service worker](https://developers.google.com/web/fundamentals/primers/service-workers/) registration,
+The main API hangs off a [service worker](/docs/workbox/service-worker-overview/) registration,
 so make sure you've registered a service worker first. Then:
 
 ```js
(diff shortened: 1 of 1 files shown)


site/en/blog/background-fetch/index.md
```diff
@@ -469,26 +466,10 @@ addEventListener('backgroundfetchclick', (event) => {
...
-implementation bugs to [the BackgroundFetch Blink
-component](https://bugs.chromium.org/p/chromium/issues/entry?template=Defect+report+from+developer&components=Blink%3EBackgroundFetch).
-
 ## Additional resources
 
 * [Explainer](https://github.com/WICG/background-fetch/blob/main/README.md)
 * [Specification](https://wicg.github.io/background-fetch/)
 * [chromestatus.com entry](https://www.chromestatus.com/features/5712608971718656)
 
+_Correction: A previous version of this article incorrectly referred to Background Fetch as being a "web standard". The API is not currently on the standards track, the [specification](https://wicg.github.io/background-fetch/) can be found in WICG as a Draft Community Group Report._
```
jakearchibald, line 475: Do we usually put changelogs in the article like this?

rachelandrew: I am suggesting that we do in the guidelines I'm creating, where there is inaccuracy fixed (in common with most editorial guidelines, which generally discourage stealth edits).

jakearchibald: Huh, so I'm currently updating articles for a couple of APIs that have undergone significant changes. Doing inline changelogs will double the length of the article and make it pretty hard to read 😢

rachelandrew: Just put a note at the end like I did? If people want to see all the details they can poke around GitHub.

rachelandrew: The aim is to alert people that major things have changed - as people link to our articles and then we make them essentially say something different. It's one of the issues of people using blog posts as docs, as the expectations are wrong, but here we are.

jakearchibald: For comparison, I'm updating an article that Sam wrote, which is now incorrect because the API has since changed. I think it would be wrong for me to claim "a previous version of this article incorrectly claimed…", and fairer to say "A previous version of this article referred to an older version of the API" (although I'm not sure how useful that is in general)

petele (approved):

cwilso:
Thanks @rachelandrew !

site/en/blog/background-fetch/index.md
rachelandrew: Yeah, I think your suggestion is fine in that case. I've definitely been in the position of quoting from an article that has then been changed, making my post look weird :D so a note like that stops someone who referenced us looking like they were talking nonsense. It's also generally seen as good practice and I'm trying to make our editorial policies a bit more robust.

jakearchibald:
It doesn't seem particularly Googly that this change was merged when I'm clearly not happy with it.

This change pretty strongly suggests that, as the author, I was at fault. But as far as I'm aware, the term "web standard" didn't have a strong definition at the time that would have put me at fault. Does it even have a strong definition now?

I don't have a problem changing references like this now we have strict rules around the term. My problem is purely with calling things out as author error, when it was actually a change of strategy after the fact.

This is why I drew the comparison to API changes. If an article needs updating due to an API change, we wouldn't call out the author for making an error. At best we'd point out that the article changed because definitions changed.

rachelandrew:
Apologies I thought it had been approved, I realise now it was by Pete, who I didn't realise was reviewing. That was my error, rather than an attempt to not be Googly, to be fair I also thought your issue was with a theoretical example about API changes rather than this one.

@cwilso can probably confirm the history of the term web standard, personally I've never considered things in WICG or in incubation a web standard or on the standards track, but maybe that's just me.

jakearchibald:
Sorry, I should have been more straightforward.

Is there a current definition for how "web standard" should be used?

rachelandrew:
Yep, internally, I'll ping you the links in chat so we don't get random requests for access :)

jakearchibald:
I'm still interested in whether this was an error on my part back when the article was written (as the published change to the article now states). cc @cwilso

But in general, I don't like notes like this being part of the article. I think as a team we should have a culture that celebrates making fixes and improvements, not one that amplifies and highlights errors and mistakes. As an author, is definitely discourages me from contributing to platforms like this, and it isn't something I'd recommend to new authors, who perhaps aren't as confident.

I realise that other publications publish changes like this, but those publications don't generally have a fully open and transparent line-by-line changelog, like we already have via GitHub.

Knowing the article was once wrong, that the author once fucked up, doesn't seem beneficial to the first time reader. For someone looking for article history, that's already all there.

Maybe we should do more to link to that changelog? The "Updated on Thursday…" text could link to the GitHub history for the article.

jakearchibald:
@cwilso ping on the above question

cwilso:
Gah, sorry, missed the notification.

We certainly did not have a policy in place at that time, or even a clear glossary to follow.  It was still incorrect to refer to something that is not on the standards track as a "Web standard", even at the point in time that the article was originally written.

I do not think it is important (or even necessary) to call attention to this change (i.e. the note), and agree that knowing the article was once wrong is not particularly beneficial, and certainly agree that there's little point in amplifying and highlighting this mistake.  It seems to me to be an editorial decision about how we communicate changes and updates in general, though, and that's not my bailiwick; sometimes it's probably quite important to communicate why a change was made.

PR #3339, opened 2022-07-29
Update shared element transition article for origin trial

jakearchibald:
This is a significant change, so I'll also get engineers to review.

Preview: https://deploy-preview-3339--developer-chrome-com.netlify.app/blog/shared-element-transitions-for-spas/

### DIFF
--- a/site/_scss/_global.scss
+++ b/site/_scss/_global.scss
@@ -23,6 +23,5 @@ main:focus {
 }
 
 video {
-  height: auto;
   width: 100%;
 }
--- a/site/_scss/_global.scss
+++ b/site/_scss/_global.scss
@@ -23,6 +23,6 @@ main:focus {
 }
 
 video {
-  width: 100%;
   height: auto;
+  width: 100%;
 }
--- a/site/_scss/_global.scss
+++ b/site/_scss/_global.scss
@@ -24,4 +24,5 @@ main:focus {
 
 video {
   width: 100%;
+  height: auto;
 }
--- a/site/en/blog/shared-element-transitions-for-spas/index.md
+++ b/site/en/blog/shared-element-transitions-for-spas/index.md
@@ -4,193 +4,152 @@ title: Smooth and simple page transitions with the shared element transition API
 authors:
   - jakearchibald
 description: >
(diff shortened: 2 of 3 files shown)


jakearchibald:
Should this article live somewhere else?

Is there a better style to use for the API reference?

jakearchibald:
Pretty sure the use of 'please' in this article is acceptable.

jh3y:
Looks great to me. Some great in-depth stuff there 👏

rachelandrew:
@jakearchibald sorry for slow replies, are there any docs for this anywhere else? If not it would be great to have stuff in docs though get you might not have time right now. I have been going through the stuff to move from web.dev, but I probably need to trawl through this site as well and see what would be better as part of the web platform docs section.

jakearchibald:
This does feel like more of a doc than a blog to me. Would it make sense to just move it and redirect, or does the style of the article need to change?

jakearchibald:
Sorry, to answer your question, nah, this isn't documented elsewhere.

rachelandrew:
Ultimately I'd like them to be in more of a documentation format, as my hope is that eventually we'd move the canonical docs to MDN (once the thing is in stable and MDN want them). So we'd have blog posts doing genuine updates, which we'd leave as the paper trail and history but our docs really be seen as a staging area for MDN docs.

That said, right now I'll put anything in any format there :D however it might be worth then doing a quick blog post announcing the changes and pointing to the docs as that will show up in people's feed readers at the top (which updates to old posts don't).

jakearchibald:
In MDN terms, I think this is more in the "Using shared element transitions" format, than the API reference docs.

rachelandrew:
@jakearchibald let's publish this where it is and then figure it out, I think there are probably a bunch of things that could do with some actual docs being created (by pulling out the relevant bits from articles and blog posts). Once I've moved the ones off web.dev I'll start going through and making a list.

rachelandrew (approved):

jakearchibald:
Thank you! I'm still planning to write a little 'cookbook' of transitions, but I'll chat to you about where that should live when I'm ready to write it.

PR #3354, opened 2022-08-02
Improve Getting Started article to fix some bugs and conventions #3353

### LINKED ISSUE #3353: Improve Getting Started article to fix some bugs and standards
Found some issues while following the step by step getting started article for developing my first chrome browser extension, following the below article.
https://developer.chrome.com/docs/extensions/mv3/getstarted/

**Fixes:**
- script `popup.js` should be executed inside `window.onload`, otherwise the extension throws a runtime exception for `changeColor` element null or undefined
- script `options.js` should be executed inside `window.onload`, otherwise the extension throws a runtime exception for `page` element null or undefined

**Improvements:**
- script files `popup.js` and `options.js` should be part of HTML `<head>` section
- minor change to swap line 380 and 381, for code to look a bit more readable

aziz-shafique:
Fixes #3353
CLA Name: Abdul Aziz Shafique

**Fixes:**
- script `popup.js` should be executed inside `window.onload`, otherwise the extension throws a runtime exception for `changeColor` element null or undefined
- script `options.js` should be executed inside `window.onload`, otherwise the extension throws a runtime exception for `page` element null or undefined

**Improvements:**
- script files `popup.js` and `options.js` should be part of HTML `<head>` section
- minor change to swap line 380 and 381, for code to look a bit more readable

### DIFF
--- a/site/en/docs/extensions/mv3/getstarted/index.md
+++ b/site/en/docs/extensions/mv3/getstarted/index.md
@@ -254,12 +254,14 @@ The last step for the popup UI is adding color to the button. Create and add a f
 `popup.js` with the following code to the extension's directory.
 
 ```js
-// Initialize button with user's preferred color
-let changeColor = document.getElementById("changeColor");
+window.onload = () => {
+    // Initialize button with user's preferred color
+    let changeColor = document.getElementById("changeColor");
 
-chrome.storage.sync.get("color", ({ color }) => {
-  changeColor.style.backgroundColor = color;
-});
+    chrome.storage.sync.get("color", ({ color }) => {
+        changeColor.style.backgroundColor = color;
+    });
+}
 ```
 
 This code grabs the button from `popup.html` and requests the color value from storage. It then
@@ -271,10 +273,10 @@ applies the color as the background of the button. Include a script tag to `popu
 <html>
   <head>
     <link rel="stylesheet" href="button.css">
+    <script src="popup.js"></script>
   </head>
   <body>
     <button id="changeColor"></button>
-    <script src="popup.js"></script>
   </body>
 </html>
 ```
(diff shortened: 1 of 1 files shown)


dotproto (changes requested):
I appreciate the intent behind this PR, but the current set of proposed changes are unnecessary, to the best of my knowledge do not follow current web development best practices, and do not improve the performance of page loads. Since you took the effort to create this PR, I'll do my best to explain my concerns.

First, placing the script tag in the document's head as suggested in this PR will block HTML parsing and page rendering until the script can be downloaded and executed. By loading the script at the end of the body, we are able to get the UI on screen as fast as possible and add interactions to the page only moments later. This strategy isn't perfect – it's technically possible for the user to see and click a button before JS is downloaded and executed – but it is reasonable and appropriate here.

Second, placing the script tag at the end of the document's body avoids the need to use page lifecycle event handlers (onload as in this PR) to register our event button event listeners. The "runtime exception for X element null or undefined" you mentioned only occurs if the script tag is moved a moved above other content in the HTML. In other words, moving the script tag to the document's head is what introduced the bug that you called out in the PR description.

Third, the `load` event occurs later in the page's life cycle than necessary, which delays the time to ready. It would be better to initialize event handlers for the page using the `DOMContentLoaded`. This event event fires after HTML has been parsed, but does not wait for “stylesheets, images, and subframes to finish loading” ([source](https://developer.mozilla.org/en-US/docs/Web/API/Window/DOMContentLoaded_event)). The `load` event only fires after all of these operations have completed.

Fourth, the pattern of using `element.event = eventHandler` has some notable drawbacks. The two main ones that stick out to me are the fact that not all events are exposed this way (I believe DOMContentLoaded is one such event) and that only a single event handler for a given element and event can be registered this way. It is generally better practices to use `element.addEventListener('event', eventHandler)`.

All that said, your PR did make me realize that we can improve on our current script loading strategy and I'd welcome . There are two specific changes from the main branch of the repo that would improve the quality of the current example:

1. Setting the `defer` attribute on script tags (see [this web.dev article](https://web.dev/efficiently-load-third-party-javascript/#defer) for details)
2. Moving script tags into the &lt;head>

A deferred script will be downloaded as early as possible but will not execute until right before DOMContentLoaded. This means that by moving the script tag to the document's head, we can prompt the browser to begin downloading the file as soon as possible while also avoiding the need to leverage page life cycle events to bind our event listeners for statically declared HTML.

dotproto:
Closing this PR as there have been no additional changes since I shared feedback about a month ago.

PR #3356, opened 2022-08-02
Fix local build

sofiayem:
Local build errors out:
```
Problem writing Eleventy templates: (more in DEBUG output)
> Having trouble rendering njk (and markdown) template ./site/en/blog/shared-element-transitions-for-spas/index.md

`TemplateContentRenderError` was thrown
> "plain" is not a valid Prism.js language for eleventy-plugin-syntaxhighlight

`Error` was thrown:
    Error: "plain" is not a valid Prism.js language for eleventy-plugin-syntaxhighlight
```

From [11ty docs](https://www.11ty.dev/docs/plugins/syntaxhighlight/#show-changes-using-diff-syntax):

```
Alternatively, you can use diff without another language name to enable plaintext line highlighting.
```

### DIFF
--- a/site/en/blog/shared-element-transitions-for-spas/index.md
+++ b/site/en/blog/shared-element-transitions-for-spas/index.md
@@ -113,7 +113,7 @@ Once that's complete, the callback passed to `.start()` is called. That's where 
 
 Once the state is captured, the API constructs a pseudo-element tree like this:
 
-```plain
+```diff
 ::page-transition
 └─ ::page-transition-container(root)
    └─ ::page-transition-image-wrapper(root)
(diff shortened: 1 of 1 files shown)


sofiayem:
JFYI @jakearchibald

jakearchibald:
Oh wow, I wonder why this didn't error for me when I was working on the article.

The code isn't a diff, so I'm not sure it's the right solution, but if it appears to work, I'm fine with it.

The 'proper' way to do this is to omit the language all together, but when I did that the linter complained.

sofiayem:
Yeah, I'm not entirely sure either but diff is what 11ty docs recommend.

rachelandrew (approved):

jakearchibald:
I think the 11ty docs are recommend `diff` for diff highlighting, that doesn't have a sub-language highlighting.

It works here, but it would break if any of the lines started `+` or `-`, as it would highlight those lines.

PR #3383, opened 2022-08-04
Update tags for deprecations and removals.

jpmedley:
After years of using them, I've found little difference in content between items tagged deprecation and items tagged removal. This PR combines them into a single tag.

### DIFF
--- a/redirects.yaml
+++ b/redirects.yaml
@@ -446,6 +446,14 @@ redirects:
 - from: /insider/...
   to: /tags/insider/
 
+# Tag Redirects
+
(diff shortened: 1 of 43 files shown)


jpmedley:
I don't see much value in editing old documents for new standards.

rachelandrew:
@jpmedley the linter is objecting as you need to add the tag in the i18n file.

jpmedley:
I pushed a change

rachelandrew (approved):

jpmedley:
It looks like I can't merge this without a review from someone in @devnook 's group.

PR #3413, opened 2022-08-09
Fix formatting of PNA article

agektmr:
Fix formatting of PNA article

### DIFF
--- a/site/en/blog/private-network-access-preflight/index.md
+++ b/site/en/blog/private-network-access-preflight/index.md
@@ -50,21 +50,25 @@ the change and adjust accordingly.
 
 1. In Chrome 104:
     * Chrome experiments by sending preflight requests ahead of private network
-      subresource
-      requests.
+      subresource requests.
     * Preflight failures only display warnings in DevTools, without otherwise
       affecting the private network requests.
-    * Chrome gathers compatibility data and reaches out to the largest
-      affected websites.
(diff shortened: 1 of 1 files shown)


agektmr:
Screen Shot 2022-08-09 at 21 20 47

site/en/blog/private-network-access-preflight/index.md
```diff
@@ -50,21 +50,25 @@ the change and adjust accordingly.
...
+    {% Aside 'warning' %}
+
+    To limit the influence to the websites not already supporting preflights, we
+    restrict the timeout to 200 milliseconds in M104. The restriction is only
+    applied in warning mode. The special timeout limit would be removed after
+    enabling the enforce mode by switching "Respect the result of Private Network
+    Access preflights" to "Enabled" in `chrome://flags` and the default limit is 5
+    seconds.
+
+    {% endAside %}
```
mihajlija, line 69:
```suggestion
    To limit the effects on websites that do not already support preflights, the
    timeout is restricted to 200 milliseconds in Chrome 104. The restriction is only
    applied in warning mode. The special timeout limit would be removed after
    enabling the enforce mode by switching "Respect the result of Private Network
    Access preflights" to "Enabled" in `chrome://flags` and the default limit is 5
    seconds.
    {% endAside %}
```

mihajlija (approved):

PR #3426, opened 2022-08-10
Remove Interactions from Performance Panel reference

jackfranklin:
These were removed in a recent milestone due to some issues, and
therefore this documentation is out of date.

We will be bringing it back (looking slightly different), and so when
that lands we will update the docs again and ensure they reflect the new
version.


jackfranklin:
(Not sure if I need to do anything to delete the image, not sure where those live)

sofiayem (approved):

PR #3465, opened 2022-08-16
[devtools-translate] French - What's New in DevTools (Chrome 105)

Nicoss54:

- add french translation for what's new in chrome DevTools (chrome 105)


jecfish:
Thanks @Nicoss54 for the PR. @@rachelandrew - iirc, French is not our supported languages. How should we deal with this PR?

Issue #2639, opened 2022-04-20
Change the way how an i18n index page shows its contents


agektmr:
Currently, index pages are displayed based on the contents in `tos.yaml`. For example, privacy sandbox page is rendered based on `/site/_data/i18n/docs/privacy-sandbox.yml` and it's at https://developer.chrome.com/docs/privacy-sandbox/

When a user goes to an i18n page, in the above case https://developer.chrome.com/ja/docs/privacy-sandbox/, they see articles that are already translated (the translated markdown exists) but without `title` property specified in the `tos.yaml`.

Issue 1: I would like to see all articles regardless of translation status. This will help the audience not to miss important contents but not translated yet.

Issue 2: By looking at the current behavior, titles displayed in https://developer.chrome.com/ja/docs/privacy-sandbox/ are translated even if the entry in `tos.yaml` doesn't have a `title` property. I presume this is because 11ty can extract the translated title from the markdown and apply that to the index page. If that is possible, I would not provide string resources as much as possible. This will help:

* Avoid duplicate translation
* It's a good indicator for the audience whether it's translated or not. If the title is in English, they can guess that it's not translated yet.
* Reduce amount of effort to translate

devnook:
@andreban @PaulKinlan FYI - we urgently need resources to solve issues like this one

matthiasrohmer:
Probably related to #2004.

matthiasrohmer:
#3686 tries to fix both issues.

Issue 2 is something we all need to be aware of when authoring for d.c.c: as you already describe, the platform indeed falls back to what is in the frontmatter. The YAMLs could be used to overrule this content which might not make sense in every language. Then the translation keys can simply be removed from the YAML.

agektmr:
Comparing the Privacy Sandbox landing page betwee English and Japanese, I noticed the issue 1 hasn't been addressed yet.
- https://developer.chrome.com/docs/privacy-sandbox/
- https://developer.chrome.com/ja/docs/privacy-sandbox/

For example, the "Get started" section starts with "Relevance and measurement unified origin trial" in English version, but it's missing in Japanese version. My expectation was to show the content in the Japanese page in English even if it's not translated yet.

matthiasrohmer:
Not ideal, but unfortunately for now landing pages still need manual work to show English original content. Follow-up work tracked in #4268.

PR #3494, opened 2022-08-19
[DO NOT MERGE] PS LP updates

heyawhite:

- Small design nits
- Japanese landing page (WIP)

### DIFF
--- a/site/ja/privacy-sandbox/index.njk
+++ b/site/ja/privacy-sandbox/index.njk
@@ -1,18 +1,12 @@
 ---
 layout: 'layouts/landing.njk'
-title: 'The Privacy Sandbox'
-description: 'A series of proposals to satisfy cross-site use cases without third-party cookies or other tracking mechanisms.'
+title: プライバシーサンドボックス
+description: プライバシーサンドボックスは、サードパーティの Cookie やその他の追跡メカニズムを使用せずにクロスサイトのユースケースを満たすことを目指す一連の提案です。
 sections:
   intro:
     - url: /docs/privacy-sandbox/overview
-      title: What is the Privacy Sandbox?
-      description: What's in it, how to get involved, and what it's for.
     - url: /docs/privacy-sandbox/glossary
-      title: Glossary
-      description: Short explanations of key concepts.
     - url: /docs/privacy-sandbox/proposal-lifecycle
-      title: Life of a proposal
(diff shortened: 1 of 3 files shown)


heyawhite:
@agektmr FYI I'm going to start work here, but what I need to figure out next is an i18n import for all of the custom descriptions, instead of relying on total copies.

More work on this next week.

agektmr:
IIUC the preview on the staging server won't include Japanese. I can check it on my local env next week. Thank you, Alexandra!

site/ja/privacy-sandbox/index.njk
```diff
@@ -0,0 +1,170 @@
...
+  "image/VbsHyyQopiec0718rMq2kTE1hke2/QDyL1Otpv6Id7sDTconF.svg",
+  "More documentation",
+  sections.intro_related
+  )
+}}
+
+{{ landingSectionUseCases(
+  "隠されたトラッキングを防ぐ",
+  "Proposals to prevent third-parties from fingerprinting or collecting unlimited sensitive data.",
+  "User-Agent reduction",
```
heyawhite, line 90:
```suggestion
  "User-Agent の情報量削減",
```

heyawhite:
@agektmr I am curious if you're able to preview it, because I wasn't. 

There are a couple of things that should probably be localized to make this document more accessible, such as the "Read the documentation." button and the subtitles for each of the use case sections.

heyawhite (commented):
One l18n effort, copied from the Japanese docs

agektmr:
I've updated Japanese text as much as possible. And here's the screenshot.

localhost_8080_ja_privacy-sandbox_

BTW, I noticed you omit links to a few of the articles that exist in English but not in Japanese. Is this intentional? I think we can keep them the same with English version on all languages, otherwise the readers will miss out and the maintenance overhead is big. I think it's fine English contents are included in other languages. What do you think?

heyawhite:
Amazing, thank you @agektmr! I actually haven't changed any of think links, I totally agree they are still valuable. My guess is because the two attribution reporting docs are new and haven't yet be localized, they won't populate in the Japanese page.

I believe you opened an Issue about this in April: https://github.com/GoogleChrome/developer.chrome.com/issues/2639

agektmr (approved):
I see, thanks @heyawhite for the clarification. If this is a separate issue, I think we are good to merge this.

PR #3503, opened 2022-08-20
Update index.md

evayde:
Chain count increased to 100 since Chrome 104, if I understand it right 

https://chromestatus.com/feature/5710690097561600

Otherwise, feel free to correct me.
I am also unsure whether we want to add a note that this only applies for Chrome >= 104.

- Increase chain count to 100

### DIFF
--- a/site/en/blog/timer-throttling-in-chrome-88/index.md
+++ b/site/en/blog/timer-throttling-in-chrome-88/index.md
@@ -234,7 +234,7 @@ Since timers are a source of excessive CPU use, we're going to continue to look 
 ways we can throttle them without breaking web content, and APIs we can
 add/change to meet use-cases. Personally, I'd like to eliminate the need for
 `animationInterval` in favor of efficient low-frequency animation callbacks. If
-you have any questions, [reach out to me on
+you have any questions, please [reach out to me on
 Twitter](https://twitter.com/jaffathecake/)!
 
 Header photo by [Heather Zabriskie](https://unsplash.com/@heatherz) on
--- a/site/en/blog/timer-throttling-in-chrome-88/index.md
+++ b/site/en/blog/timer-throttling-in-chrome-88/index.md
@@ -9,6 +9,7 @@ description: >
   30 seconds, WebRTC is not in use, and the chain of timers is
   100 or greater.
 date: 2021-01-18
+updated: 2022-08-30
 hero: image/CZmpGM8Eo1dFe0KNhEO9SGO8Ok23/5diIn5KBi44iwy3o6GRi.jpg
 alt: A collection of broken clocks
 ---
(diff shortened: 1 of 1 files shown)


site/en/blog/timer-throttling-in-chrome-88/index.md
```diff
@@ -234,7 +234,7 @@ Since timers are a source of excessive CPU use, we're going to continue to look
 ways we can throttle them without breaking web content, and APIs we can
 add/change to meet use-cases. Personally, I'd like to eliminate the need for
 `animationInterval` in favor of efficient low-frequency animation callbacks. If
-you have any questions, please [reach out to me on
+you have any questions, [reach out to me on
```
jakearchibald, line 237:
```suggestion
you have any questions, please [reach out to me on
```

jakearchibald:
@rachelandrew I've seen the 'please' rule catch folks out a few times. Maybe remove that rule, unless it's catching a lot of cases where it should be removed.

jakearchibald:
It isn't really clear from https://developers.google.com/style/word-list#please which uses of "please" should be avoided, aside from "please note".

jakearchibald:
Need to add an "updated" date before merging?

site/en/blog/timer-throttling-in-chrome-88/index.md
```diff
@@ -7,7 +7,7 @@ description: >
   Intensive throttling takes effect when the page has been hidden
   for more than 5 minutes, the page has been silent for at least
   30 seconds, WebRTC is not in use, and the chain of timers is
-  5 or greater.
+  100 or greater.
 date: 2021-01-18
```
heyawhite, line 11:
```suggestion
date: 2021-01-18
updated: 2022-08-30
```

heyawhite (approved):

jakearchibald:
> Anyway, doesn't matter so much to me in this instance, and the guide does state "recommended".

I think the problem is the guidance isn't clear on cases where "please" is bad. It gives an exception:

> Use please only when you're asking for permission or forgiveness—for example, when what you're asking for benefits us, inconveniences a reader, or suggests a potential issue with a product.

And that matches this case, since we're asking for feedback, which benefits us. Unless I've misunderstood the rule?

I guess I'm still unclear when 'please' is bad, other than the specific example given: "please note…".

heyawhite:
@jakearchibald you're right, it is recommended in this case. 

There's more in the style guide in the [Voice and tone section](https://developers.google.com/style/tone#politeness-and-use-of-please). We often overdo politeness, and should err on the side of brevity.

The guideline should state: "Only use please when we're asking a user to go out of their way to complete a task." (Again, this example could fit this guidance. However, it's not required of our users to take this action, so it's worth consideration of necessity. It could be considered too formal.)

Externally:
- [Making Demands: When to Say Please in Instructions and CTA](https://www.sitepoint.com/making-demands-when-to-say-please-in-instructions-and-ctas/)

PR #3524, opened 2022-08-23
[EXT-DOC] Add headers to MV3 migration checklist 

AmySteam:
## Problem

Currently, the [MV3 Migration checklist](https://developer.chrome.com/docs/extensions/mv3/mv3-migration-checklist/) does not have headers, which makes it hard to go directly to the section one is interested in.

## 👍 Solution

Convert bolded paragraphs to headers with anchor tags.

**Staged link**: [MV3 Migration Checklist](https://deploy-preview-3524--developer-chrome-com.netlify.app/docs/extensions/mv3/mv3-migration-checklist/)

## Additional tweaks 
- Convert inline links into footer links
- Add more helpful links to aid migration.
- Continue using the agreed API convention: Tabs API instead of `chrome.tabs` API.
- Removed an aside gotcha that doesn't apply anymore.

## ❌ Does not include
- Any additional migration checklist items.
- Verification that all items are still valid.
- Other tweaks to the layout.

### DIFF
--- a/site/en/docs/extensions/mv3/mv3-migration-checklist/index.md
+++ b/site/en/docs/extensions/mv3/mv3-migration-checklist/index.md
@@ -110,9 +110,11 @@ and `chrome.extension.getViews()`.*
 There are some changes you may need to make based on changes in security policy. This section lists these changes.
 
 ### Are you making CORS requests in content scripts? {: #security-cors }
+
 - Move these requests to the background service worker.
 
 ### Are you using a custom `content_security_policy` in manifest.json? {: #security-csp }
+
 - Replace `content_security_policy` with `content_security_policy.extension_pages`
   or `content_security_policy.sandbox` as appropriate.
 - Remove references to external domains in `script-src`, `worker-src`, `object-src`, and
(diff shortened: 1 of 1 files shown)


AmySteam:
Hey @rachelandrew! I'd like to know your thoughts on this one 😄 .

We'd like to link directly to each migration suggestion. We converted from bold paragraphs to the next header level (H3). As you can see in the screenshots below, it doesn't seem to look quite right. 

❓Do you have any suggestions that would work well with screen readers?

### No headers - [MV3 migration link](https://developer.chrome.com/docs/extensions/mv3/mv3-migration-checklist/)
Screen Shot 2022-08-24 at 09 41 29

### With Headers - [Staged migration link](https://deploy-preview-3524--developer-chrome-com.netlify.app/docs/extensions/mv3/mv3-migration-checklist/)
Screen Shot 2022-08-24 at 09 46 02

rachelandrew:
I think the headings are fine to be honest, and it's the only way you'll get a nice list in the navigation for people to hop to them so I would keep them. When there isn't a lot of text following they look a bit heavy, but I think this is the best solution in the desk we have now.

dotproto (approved):
LGTM

PR #3553, opened 2022-08-25
Add web vitals library

tunetheweb:

- Adds web vitals library, in similar set up to web.dev
- Adds web vitals attribution tracking
- Adds click event for links with `data-category` and bfcache page transition tracking, same as web.dev


site/_js/analytics.js
```diff
@@ -0,0 +1,156 @@
...
+//   * "devsite-analytics-observation", generated by the Metric class for page load timing
+//   * "devsite-analytics-error", wired up to the onerror handler
+//   * "devsite-analytics-set-dimension", for user response to custom questions
+//   * "devsite-analytics-scope", which constructs complex click events (was used for header)
+//
+// Note that many parts of our code are annotated with an "gc-analytics-event" attribute, but this
+// is actually ignored in DevSite v2. Instead, any links that have `data-category` automatically
+// have clicks logged (see below).
+
+function getAnalyticsDataFromElement(elem, defaultAction = 'click') {
```
philipwalton, line 21: I think we probably don't want to include this function, especially since this code will be temporary, and the functionality isn't being used to right now.

site/_js/analytics.js
```diff
@@ -0,0 +1,156 @@
...
+    action,
+    label,
+    value,
+  };
+}
+
+/**
+ * @param {{ category?: string, action?: string, label?: string, value?: number }} param
+ */
+export function trackEvent({category, action, label, value}) {
```
philipwalton, line 37: Same with this function...

site/_js/analytics.js
```diff
@@ -0,0 +1,156 @@
...
+}
+
+/**
+ * Track an error via Analytics with optional context message and fatal notice.
+ *
+ * @param {Error} error to log
+ * @param {string=} message context to provide around error message
+ * @param {boolean=} fatal whether this is fatal (as per Analytics' logging)
+ */
+export function trackError(error, message = '', fatal = false) {
```
philipwalton, line 53: Same with this function...

site/_js/analytics.js
```diff
@@ -0,0 +1,156 @@
...
+  });
+}
+
+/**
+ * Configure tracking events for any clicks on a link (`<a href="...">`)
+ * or another trackable element (class="gc-analytics-event"), searching
+ * for (requiring at least `data-category`, but also allowing
+ * `data-action`, `data-label` and `data-value`.
+ */
+document.addEventListener(
```
philipwalton, line 112: Probably don't need to include this (similar to above).

types/modules/webdev_analytics/index.d.ts
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright 2020 Google LLC
```
philipwalton, line 2: Super nit:

s/2020/2022/

philipwalton (changes requested):

site/_js/analytics.js
tunetheweb: Yeah I was torn between consistency, and just installing what this PR needs.

Removed them all now.

rollup.config.js
```diff
@@ -13,6 +13,9 @@ import {terser} from 'rollup-plugin-terser';
 // A Rollup plugin for copying files.
 import copy from 'rollup-plugin-copy';
 
+// A Rollup plugin for copying files.
+import json from '@rollup/plugin-json';
```
philipwalton, line 17: What is this being used for? (Also, looks like the comment might have just been copied from above?)

tunetheweb: It's used to read the dimension numbers from [the JSON config file](https://github.com/GoogleChrome/developer.chrome.com/blob/5242c85273759dc21535e8d5d99e479617b37807/site/_data/analytics.json) rather than hardcoding them.

https://github.com/GoogleChrome/developer.chrome.com/blob/5242c85273759dc21535e8d5d99e479617b37807/site/_js/analytics.js#L49-L50

Will update the comment.

philipwalton: Oh, I see; it's to make this line work:

```js
import {dimensions} from '../_data/analytics.json';
```

philipwalton: Aside: not worth doing for just this PR, but I'm curious if there's a more standard-compliant way to do this with Rollup, e.g. https://bobbyhadz.com/blog/javascript-import-json-file

philipwalton (approved):

rollup.config.js
tunetheweb: Actually when I tried to import this rollup TOLD me to install this plugin! 😀

I checked and web.dev uses something different but was beyond me, and this was a simple suggested plugin.

PR #3556, opened 2022-08-26
Update popup link

hanselfmu:
The popup link in this top layer blog post is not correct. This PR updates the link and as a drive-by adjusts the sentence a little bit to make it clear what are top-layer-eligible elements.

### DIFF
--- a/site/en/blog/top-layer-devtools/index.md
+++ b/site/en/blog/top-layer-devtools/index.md
@@ -31,7 +31,7 @@ The [top layer](https://fullscreen.spec.whatwg.org/#new-stacking-layer) can be d
 Multiple elements can be inside the top layer at the same time. When that happens, they stack on top of each other, the last one on top. In other words, all of the top layer elements are placed in a *last in, first out* (LIFO) stack in the top layer.
 
 The `<dialog>` element is not the only element that the browser renders into a top layer. Currently, the top layer elements are:
-[popups](https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/user_interface/Popups), [modal dialogs](https://developer.mozilla.org/docs/Web/HTML/Element/dialog), and elements in a [fullscreen mode](https://developer.mozilla.org/docs/Web/API/Fullscreen_API).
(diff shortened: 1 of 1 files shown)


site/en/blog/top-layer-devtools/index.md
```diff
@@ -31,8 +31,7 @@ The [top layer](https://fullscreen.spec.whatwg.org/#new-stacking-layer) can be d
 Multiple elements can be inside the top layer at the same time. When that happens, they stack on top of each other, the last one on top. In other words, all of the top layer elements are placed in a *last in, first out* (LIFO) stack in the top layer.
 
 The `<dialog>` element is not the only element that the browser renders into a top layer. Currently, the top layer elements are:
-[popups](https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/user_interface/Popups)
-[modal dialogs](https://developer.mozilla.org/docs/Web/HTML/Element/dialog) elements in a [fullscreen mode](https://developer.mozilla.org/docs/Web/API/Fullscreen_API).
+[popups](https://open-ui.org/components/popup.research.explainer), [modal dialogs](https://developer.mozilla.org/docs/Web/HTML/Element/dialog), and elements in a [fullscreen mode](https://developer.mozilla.org/docs/Web/API/Fullscreen_API).
```
mfreed7, line 34: One nit: this should be "pop-ups" with a dash.

mfreed7 (commented):
Thanks, looks good with just one nit!

site/en/blog/top-layer-devtools/index.md
heyawhite: Either with or without a dash is appropriate in this instance, though the new link references the "Pop Up API" and "pop-ups". Our style guide says to [avoid usage](https://developers.google.com/style/word-list#pop-up) however it is appropriate when it's the name of the component.

mfreed7: In this case, it's material. Please see this discussion: https://github.com/openui/open-ui/issues/546#issuecomment-1158190204

The name of the API was purposely changed to "the pop-up API" so that "pop" and "up" are two words, so that the IDL attribute can be called `popUp` with a capital "U", so we don't break the web. Ugh!

So here, ideal would be "pop-ups", but "pop ups" would also be ok. But not-ok is "popups". Sorry.

hanselfmu: Thanks for the explanation! I've updated the PR to reflect this point.

heyawhite (approved):

PR #3603, opened 2022-08-30
Replace deprecated tag that is causing a build error

dotproto:
The 'removals' tag was replaced with 'deprecations-removals' in #3383. This commit updates posts that referenced the old tag in order to fix build errors.


samdutton (approved):
Thanks!

PR #3619, opened 2022-09-01
Added a detailed article about using NgOptimizedImage directive to improve image loading performance in Angular

arungudelli:
Fixes #SOME_ISSUE_NUMBER

-
-
-


rachelandrew:
@arungudelli thanks for the contribution however please don't add further reading sections to articles. This infers that the original author has suggested the resource, which isn't the case.

arungudelli:
@rachelandrew Thanks for your comment. I thought that will be a good addition as the article is very detailed.

PR #3640, opened 2022-09-05
[Anonymous iframes article] Call out `require-corp` is not a typo

tomayac:

- `s/require-corp/require-coep`


agektmr:
This is not a typo. COEP does require CORP for 3p resources to opt-in.

agektmr:
It does require COEP as well for iframes though.

tomayac:
Wow, then this probably should be called out somehow. The article talks about `COEP`, and on the English keyboard `e` and `r` are next to each other, so it looked like a typo. 

PTAL for the new aside that calls this out.

agektmr (approved):
LGTM

PR #3665, opened 2022-09-08
Auto-redirect to translated page if there is one

### LINKED ISSUE #2989: Show localized content when the browser sends a `Accept-Language` header and it's available
When I visit https://developer.chrome.com/docs/privacy-sandbox/ with a browser that sets Japanese as the primary language, the website still shows English contents. I hope it displays localized content when no particular language is specified in the path (`/ja/`) and the translation is available.  
This is also not consistent with the behavior on web.dev. web.dev works as I describe above.

Here is the header my browser sends for example.

```
accept-language: ja,ja-JP;q=0.9,en;q=0.8
```

matthiasrohmer:
Resolves #2989. This adds a redirect handler before express.static kicks in checking if there is a file under the request path prefixed with any of the user's preferred locales.

This way with German added as accepted language in your browser settings (chrome://settings/?search=add+languages), developer.chrome.com/docs/privacy-sandbox/attribution-reporting/ redirects to developer.chrome.com/**de**/docs/privacy-sandbox/attribution-reporting/.


server/language-redirect.js
```diff
@@ -0,0 +1,83 @@
...
+];
+
+/**
+ * @type {express.RequestHandler}
+ */
+const languageRedirectHandler = (req, res, next) => {
+  const pathParts = req.path.substring(1).split('/');
+  const pathPrefix = pathParts[0];
+
+  // If the request path is is a root level path, starts with two chars
```
devnook, line 44: Maybe instead of heuristic based on number of letters etc you can use supported locales from https://github.com/GoogleChrome/developer.chrome.com/blob/main/site/_data/site.json#L8 ?

server/language-redirect.js
```diff
@@ -0,0 +1,83 @@
...
+      rootDir,
+      language,
+      ...pathParts,
+      'index.html'
+    );
+    try {
+      fs.accessSync(translatedPath);
+      return doRedirect(res, path.join('/', language, ...pathParts));
+    } catch (e) {
+      // That's fine. That just means there is no file for
```
devnook, line 75: Consider logging here?

devnook:
reviewed

devnook (changes requested):
Left some comments

server/language-redirect.js
matthiasrohmer: There was a comment/issue/whatever somewhere that things in server should not rely on things from site in order to be able to only ship server and dist directories to App Engine at some point, to improve build times. That's why I did it that weird way 😅  but refactored it as suggested.

server/language-redirect.js
matthiasrohmer: That would fire a lot as that catch gets called for every request that prefers another language than English while most content isn't translated, making the fs call fail. So this might probably drown out meaningful log messages in StackDriver 🤔 That's why I didn't do it. WDYT?

matthiasrohmer:
I added a test suite for this. As I didn't want to abstract the logic away from an actual request it is a little weird. The main reason for it being weird though is that we have no proper env management yet 😅 

Also: I built this under the assumption that we have a CDN in front of d.c.c - if that's not the case we might want to cache the fs lookups 🙂

server/language-redirect.js
devnook: Makes sense!

server/language-redirect.js
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    const translatedPath = path.join(
+      rootDir,
+      language,
+      ...pathParts,
+      'index.html'
+    );
+    try {
+      fs.accessSync(translatedPath);
```
devnook, line 83: I noticed recently that fs calls in GAE are really slow, and this handler gets called before each page render. How about we redirect without checking, and then let the NotFoundHandler serve 'en' backup if needed?

devnook (changes requested):
Let's try the notFoundHandler approach to speed things up

matthiasrohmer:
This doesn't use fs lookups anymore but just stupidly prepends the language matched from the accept-language header and lets the existing handlers do the rest.

devnook (approved):

Issue #2004, opened 2022-01-24
Localized `/blog/` should list articles from the default language


agektmr:
**Describe the bug**
Currently the localized list of articles under `/blog/` only have ones that are already translated. But this is a huge opportunity loss for the language speaker who will miss other non-translated articles. They will have to go back and forth between English and localized contexts.

**To Reproduce**
1. Go to https://developer.chrome.com/ja/blog/

**Expected behavior**
Lists a mixture of articles that are translated and non-translated.

**Screenshots**
Lists just Japanese translated articles.

matthiasrohmer:
Shipped with #3686.

PR #3686, opened 2022-09-12
Fix various issues around internationalised landing and overview pages

### LINKED ISSUE #2639: Change the way how an i18n index page shows its contents

matthiasrohmer:
Fixes #2639 and #2004.

This PR makes the `internalFind` method which is used when building navigations/page lists fallback to then English language directory if there is no match for a specified locale. It also makes a similar change for the method that creates the blog and articles overview by using the English tree as a base and merging in the localized tree.

Additionally it removes the translated strings from the YAMLs where better strings already exist on the page's frontmatter.

/cc @agektmr


site/_utils/addCollectionByDirectory.js
```diff
@@ -0,0 +1,88 @@
...
+ * alongside untranslated ones.
+ */
+
+const path = require('path');
+const {filterOutDrafts} = require('./drafts');
+
+const defaultLocale = 'en';
+
+function addCollectionByDirectory(config, locale, dir) {
+  config.addCollection(`${dir}-${locale}`, collections => {
```
devnook, line 30: I think this function should be defined in the _collections folder: https://github.com/GoogleChrome/developer.chrome.com/tree/main/site/_collections and then the addCollection call can stay in the main eleventy file.

site/_utils/addCollectionByDirectory.js
```diff
@@ -0,0 +1,88 @@
...
+    // sense in trying to merge them, just return the English articles then
+    if (localeCollection.length === 0) {
+      return baseCollection;
+    }
+
+    // Otherwise walk through the English collection while looking
+    // for matches in the locale collection and if it's a match, replace
+    let localeIndex = 0;
+    let collection = [];
+    for (let index = 0; index < baseCollection.length; index++) {
```
devnook, line 65: This got complex enough to deserve a test.

site/_utils/addCollectionByDirectory.js
```diff
@@ -0,0 +1,88 @@
...
+      const item = baseCollection[index];
+      const localizedItem = localeCollection[localeIndex];
+
+      if (item.fileSlug === localizedItem.fileSlug) {
+        collection.push(localeCollection[localeIndex]);
+        localeIndex++;
```
devnook, line 71: If first 3 items in baseCollection are a miss, then on 4th iteration index=3 but localeIndex=0 and they get out of sync. You'll be pushing wrong article to the collection array.

site/_utils/addCollectionByDirectory.js
```diff
@@ -0,0 +1,88 @@
...
+      } else {
+        collection.push(baseCollection[index]);
+      }
+
+      // If there are no items left to potentially localize then simply
+      // append the rest of the articles and stop
+      if (localeIndex === localeCollection.length) {
+        collection = collection.concat(baseCollection.slice(index + 1));
```
devnook, line 79: 11ty collections are sorted by default (by date *I think*), you'll loose this behavior here. There might be parts of the code that silently rely on this characteristic, so maybe worth seorting the collection here.

devnook (changes requested):
Thanks @matthiasrohmer ! Let's also add a test for this collection please.

site/_utils/addCollectionByDirectory.js
matthiasrohmer: Clarified offline. The code works as intended under the assumption that `localeCollection` has the same order as `baseCollection` and is a translated subset of it, not containing any original items.

site/_utils/addCollectionByDirectory.js
matthiasrohmer: Clarified offline. The code works as intended under the assumption that `localeCollection` has the same order as `baseCollection` and is a translated subset of it, not containing any original items.

Agreed to only rely on slugs for matching instead of on date order as well.

devnook:
Hi @matthiasrohmer , where are we standing with this Pr?

devnook (approved):

PR #3687, opened 2022-09-12
Switch from light grey to dark blue for code comment highlighting

### LINKED ISSUE #3357: Change the syntax highlight colour of comments
For example: https://developer.chrome.com/blog/shared-element-transitions-for-spas/

Screen Shot 2022-08-02 at 12 28 28

The syntax highlighting for comments seems to suggest they're not important. I generally disagree with this, but I particularly disagree with it on educational content – comments are just as important as the code around it.

I think we should change the colour so these stand out more. I'm happy to do a PR for this if there's agreement.

Fwiw, I did the same thing [on my blog](https://jakearchibald.com/2021/dom-cross-fade/#why-doesnt-it-work). The default highlighting was 'faded out', so I changed it to something stronger.

matthiasrohmer:
Fixes #3357. I would have liked to use green, but the darkest shade from the blue palette is the only color that satisfies a11y. 

image

### DIFF
--- a/site/_scss/_theme.scss
+++ b/site/_scss/_theme.scss
@@ -24,7 +24,7 @@
   --rgb-button-overlay: 60, 64, 67;
 
   --color-code-bg: #{get-color('grey-50')};
-  --color-code-comment: #{get-color('grey-500')};
+  --color-code-comment: #{get-color('blue-900')};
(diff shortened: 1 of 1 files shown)


jakearchibald (approved):
LGTM!

PR #3692, opened 2022-09-12
Introduce ` {% partial %}` shortcode 

### LINKED ISSUE #3346: Includes are not internationalized
For example this article: https://developer.chrome.com/ja/docs/privacy-sandbox/attribution-reporting-experiment/

Everything is translated except "Join the discussion" and "Get updates".

Looking at the markdown, they are included by:

```
{% include 'content/privacysandbox-partials/ar-join-discussion-long.njk' %}

{% include 'content/privacysandbox-partials/ar-get-updates.njk' %}
```

We need to internationalize things under `_includes/contents`.

matthiasrohmer:
Resolves #3346. This PR introduces a new shortcode `{% partial %}` which is able to include templates living in any of the language folders under _partial. The locale is automatically extracted and it falls back to English if there is no translated version of a partial.

As the partials do not have a permalink (to not get rendered) it was necessary to extend the find utilities to also allow searching for content by path instead of only URL.

The shortcode is implemented as an async nunjucks shortcode as the found partial template needs to be manually rendered as it is not guaranteed that they have already been rendered by 11ty at time of use.

Additionally this PR also moves strings that should be translated from templates to the i18n YAMLs and moves all existing long-form content partials from the _includes folder to the newly created _partial language directories.


matthiasrohmer:
This is now ready for review.

.eleventy.js
```diff
@@ -149,6 +156,7 @@ module.exports = eleventyConfig => {
   eleventyConfig.addPairedShortcode('Aside', Aside);
   eleventyConfig.addPairedShortcode('Label', Label);
   eleventyConfig.addShortcode('LanguageList', LanguageList);
+  eleventyConfig.addNunjucksAsyncShortcode('Partial', Partial);
```
devnook, line 159: Can you check if it is possible to use shortcodes inside a partial now? We had problems before to use a shortcode or an async shortcode inside another async shortcode.

site/_data/i18n/common.yml
```diff
@@ -182,10 +182,51 @@ website:
...
+  gdg:
+    title:
+      en: Google Developer Groups
+    body:
+      en: |
+        Connect with fellow developers and startups in your area, 
+        hear from industry experts, share your skills, and learn.
+    cta:
+      en: Find your local group
+cookie_banner:
```
devnook, line 199: Nit: new line above

site/_data/lib/find.js
```diff
@@ -1,51 +1,42 @@
...
  * @param {EleventyCollectionItem[]} collection
- * @param {string} urlToFind
+ * @param {unique symbol} cacheKey
+ * @param {string} key
+ * @param {string} needle
  * @return {EleventyCollectionItem|undefined}
  */
-const internalFind = (collection, urlToFind) => {
-  /** @type {{[url: string]: EleventyCollectionItem}} */
+const internalFind = (collection, cacheKey, key, needle) => {
```
devnook, line 18: Please add tests in https://github.com/GoogleChrome/developer.chrome.com/blob/main/tests/site/_data/lib/find.js

site/_data/lib/find.js
```diff
@@ -1,51 +1,42 @@
...
+const defaultLocale = 'en';
 
 /**
- * Builds and uses a cache to ensure fast lookup of Eleventy pages.
+ * Builds and uses a cache to ensure fast lookup of Eleventy pages by
+ * a specified key.
  *
  * @param {EleventyCollectionItem[]} collection
- * @param {string} urlToFind
+ * @param {unique symbol} cacheKey
```
devnook, line 13: Link complains here

site/_data/lib/find.js
```diff
@@ -1,51 +1,42 @@
...
+ * @param {string} key
```
devnook, line 14: Please add a description for params, esp. key and needle as they are not self-descriptive

site/_data/lib/find.js
```diff
@@ -1,51 +1,42 @@
...
     collection[cacheKey] = cache;
   }
 
-  if (urlToFind in cache) {
-    return cache[urlToFind];
+  if (needle in cache) {
+    return cache[needle];
   }
 
   // Otherwise, be slow, since we think this is probably pretty rare.
```
devnook, line 34: Curious - in what case would this happen?

site/_data/lib/find.js
```diff
@@ -1,51 +1,42 @@
...
-  const result = collection.find(item => item.url === urlToFind);
+  const result = collection.find(item => item[key] === needle);
```
devnook, line 35: item[key] can be an object. Would comparing objects (by ref) work here?

site/_data/lib/find.js
```diff
@@ -74,15 +65,54 @@ const findByUrl = (collection, url, locale = '') => {
...
 const findByProjectKey = (collection, projectKey, locale) => {
   if (!projectKey || !locale) {
     return undefined;
   }
   const urlToFind = path.join('/', locale, '/docs/', projectKey, '/');
-  return internalFind(collection, urlToFind);
+  return internalFind(collection, urlCacheKey, 'url', urlToFind);
+};
+
+const findByFilePath = (collection, filePath, locale = '') => {
```
devnook, line 79: Please add a doc here, especially regarding what is considered root for the filePath here (. or /site or ...)?

site/_shortcodes/Partial.js
```diff
@@ -0,0 +1,43 @@
...
+  if (!partial) {
+    console.warn(`No partial found at ${partialPath}`);
+    return '';
+  }
+
+  // Proactively render the template to prevent 11ty from complaining
+  // that we are accessing a rendered template (partial.templateContent)
+  // before it has done all it's business
+  const template = partial.template;
+  template.wrapWithLayouts = false;
```
devnook, line 36: TIL!

devnook (changes requested):
Thanks @matthiasrohmer ! I left some comments

site/_data/lib/find.js
matthiasrohmer: Done, also made types more safe.

site/_data/lib/find.js
matthiasrohmer: It can't, I don't know why it was there. Collections can't be altered at build time so if they are not in the lookup table (aka cache here) they don't exist. Hence removed it.

site/_data/lib/find.js
matthiasrohmer: See comment above, removed.

site/_data/lib/find.js
matthiasrohmer: Documented both this and the already existing findByUrl.

.eleventy.js
matthiasrohmer: Partial is the only async shortcode currently though, right? Nested shortcodes just work fine, I did test that. But will add another test once #3783 lands 🙂

site/_data/lib/find.js
matthiasrohmer: Added tests for the new `findByFilePath` methods. That should cover all uses of `internalFind`.

matthiasrohmer:
Will add additional tests after https://github.com/GoogleChrome/developer.chrome.com/pull/3783 landed. But also this one here should probably sit till after the summit to not cause any irritations next week and being able to educate authors 🙂

devnook (approved):
LGTM

devnook:
@matthiasrohmer can you rebase and merge?

matthiasrohmer:
Yep! Now that https://github.com/GoogleChrome/developer.chrome.com/pull/3783 is in I will quickly add some additional tests, then this is good to go 🙂

PR #3693, opened 2022-09-12
Clarify that MV3 uses a single sw.

jpmedley:
Clarify that only a single service worker can be used in Manifest V3, and fix other minor issues.

[Staged link](https://deploy-preview-3693--developer-chrome-com.netlify.app/docs/extensions/mv3/migrating_to_service_workers/)


jpmedley:
@dotproto, do you have any issues with this?

PR #3710, opened 2022-09-14
Update Angular image code snippet

kara:
The name of the Angular image property has changed, so we should update the code example.

- Change "rawSrc" to "ngSrc"


AaronForinton (approved):
Approving this and sourcing this change in terminology: [NgOptimizedImage](https://angular.io/api/common/NgOptimizedImage)

Issue #65, opened 2020-12-17
eleventyIgnore script should pull from .gitignore


robdodson:
@robdodson commented on [Mon Dec 14 2020](https://github.com/GoogleChrome/dcc-private/issues/518)

We have a script in tools/eleventyignore that will dynamically generate an .eleventyignore file for us. We do this so we can toggle folders off with .env vars.

Currently this script has a `ignores` array with a set of items that should always be ignored. Ideally we should just pull this from `.gitignore` instead of having an array.

w0rstpr0grammer:
I would like to work on it, would you please assign it to me.

w0rstpr0grammer:
Have submitted a PR for this (#761)

robdodson:
Ah thanks @w0rstpr0grammer.

I merged your PR but had to revert it in aa13762ed7972a618db9f84c087c62cdef0d12c4 because it seemed to be breaking the build (https://github.com/GoogleChrome/developer.chrome.com/pull/845#pullrequestreview-674159275).

![image](https://user-images.githubusercontent.com/1066253/120525360-e8974e80-c38c-11eb-9c60-299b2bc57320.png)

matthiasrohmer:
Everything that is in .gitignore is automatically ignored by 11ty. Also see https://www.11ty.dev/docs/ignores/.

PR #3765, opened 2022-09-19
Script to automate DevTools tasks

jecfish:
Automate various DevTools tasks:

1. Generate What's New in DevTools blog post templates in 7 languages
2. Generate What's New in DevTools blog post banner images in 7 languages
3. Generate What's New in DevTools blog outlines in 7 languages

... and more. Refer to `README.md` for instructions and all scripts.


matthiasrohmer (changes requested):
That'll make things quicker to handle, nice! Some thoughts and questions: currently, `gulp-tasks` only holds tasks that are necessary to actually build the site, tools that help doing so live in `tools`. Can you please move devtools.js there, into a new folder? 

Ideally the template files `new-in-devtools.md` and `new-in-devtools-banner.svg` then also live alongside in the same directory. Because similar as with the gulp tasks, the template files you created are not consumed directly by 11ty and therefore it would be nice to separate them from it 🙂 

Also a more descriptive name for the tool would be great! Something like devtools-blog-bootstrap maybe? Anything more descriptive would help future maintainers I think.

jecfish:
Thanks @matthiasrohmer. I have rename and organised the files. PTAL. It's a multi step automation process. I split them into different scripts.

.gitignore
```diff
@@ -12,6 +12,7 @@ Thumbs.db
 .lighthouseci
 .idea*
 .python-version
+_temp
```
matthiasrohmer, line 15: Similar to the earlier comments: I think every directory or file at root-level should have some relation to the actual 11ty setup, so I'd move this to tools/devtools/temp. This will also make it obvious in the .gitignore what it's for and won't draw confusion when working on things like https://github.com/GoogleChrome/developer.chrome.com/issues/65

package.json
```diff
@@ -71,6 +71,7 @@
     "lit-html": "^1.3.0",
     "node-fetch": "^2.6.0",
     "npm-run-all": "^4.1.5",
+    "octokit": "^2.0.7",
```
matthiasrohmer, line 74: I try to pin all newly introduced dependencies to have all of them pinned at one point. The reasoning behind this is that having pinned dependencies usually solves a lot of errors with lock files for not so regular contributors.

tools/devtools/templates/new-in-devtools.md
```diff
@@ -0,0 +1,33 @@
+---
```
matthiasrohmer, line 1: Sorry, didn't think about it earlier, also just an idea: as this is essentially a template you could also use any of the template languages that 11ty ships with. Nunjucks for example. Pro: you would not have to deal with RegEx patterns and String.replace, Con: you would need to wrap things like `{% include %}` with `{% raw %}`. Your call 🙂

tools/devtools/utils.mjs
```diff
@@ -0,0 +1,262 @@
+import {readFileSync, writeFileSync, mkdirSync} from 'fs';
```
matthiasrohmer, line 1: At a glance all the methods in your tool that make file operations are `async` already so I'd vote to use the variants from `fs/promises`. That usually makes things easier to refactor later on as well and saves from probably unexpected halts

tools/devtools/utils.mjs
```diff
@@ -0,0 +1,262 @@
...
+ * @returns YYYY-MM-DD
+ */
+function getToday() {
+  const dateObj = new Date();
+  const year = dateObj.getUTCFullYear();
+  const month = (dateObj.getMonth() + 1).toString().padStart(2, '0');
+  const date = dateObj.getUTCDate().toString().padStart(2, '0');
+
+  return `${year}-${month}-${date}`;
+}
```
matthiasrohmer, line 71: It's a shame DateTimeFormat does not allow custom formats, but what you could do is:

`return new Intl.DateTimeFormat("fr-CA", {year: 'numeric', month: '2-digit', day: '2-digit'}).format(Date.now());`

That's a little hacky though as the `fr-CA` locale is what takes care of the order then.

tools/devtools/utils.mjs
```diff
@@ -0,0 +1,262 @@
...
+
+    const outline = outlineRaw.replace(
+      contentHolder,
+      contentHolder + '\n\n' + langOutput
+    );
+
+    writeFileSync(outlineFileName, outline, 'utf-8');
+  }
+
+  console.log(`Append WNDT outlines for: ${langs.join(', ')}`);
```
matthiasrohmer, line 228: Nit: `Appended`, just so it's clear it already happened

tools/devtools/wndt-bootstrap.mjs
```diff
@@ -0,0 +1,13 @@
+import {createWndtBanners, createWndtBlogPosts} from './utils.mjs';
```
matthiasrohmer, line 1: Not too sure, but you'll probably want to move this after the `config()` call to ensure things are really setup before doing anything else

tools/devtools/wndt-bootstrap.mjs
```diff
@@ -0,0 +1,13 @@
...
+import {config} from 'dotenv';
+
+config();
+
+if (!process.env.DEVTOOLS_VERSION) {
+  throw 'Please make sure have a .env file with parameter DEVTOOLS_VERSION="xxx".';
+}
+
+// @ts-ignore
+createWndtBanners(process.env.DEVTOOLS_VERSION, ['en']);
```
matthiasrohmer, line 11: Technically there would be `const {defaultLocale} = require('site/_data/site.json');` but since you need to hardcode locales in other places I think we shouldn't bother.

tools/devtools/wndt-outline.mjs
```diff
@@ -0,0 +1,19 @@
+import {createWndtOutline} from './utils.mjs';
```
matthiasrohmer, line 1: Same as above, this should probably come after `config()`

tools/devtools/wndt-translation-bootstrap.mjs
```diff
@@ -0,0 +1,34 @@
...
+
+const languages = ['es', 'ja', 'ko', 'pt', 'ru', 'zh'];
+
+createWndtBanners(process.env.DEVTOOLS_VERSION, languages);
+createWndtBlogPosts(process.env.DEVTOOLS_VERSION, languages);
+populateTranslationContent(process.env.DEVTOOLS_VERSION, languages);
+createGitHubIssues(
+  process.env.DEVTOOLS_VERSION,
+  process.env.DEVTOOLS_TRANSLATE_DUE,
+  ['es', 'ja', 'ko', 'pt', 'ru', 'zh'],
```
matthiasrohmer, line 32: Why not use `languages` here as well?

matthiasrohmer (commented):
Left some comments, let me know what you think!

.remarkignore
```diff
@@ -0,0 +1 @@
+site/_includes/partials/devtools/*/whats-new.md
```
jecfish, line 1: @matthiasrohmer I took your advise to use 11ty/nunjucks syntax. However, I got a warning from the `lint:md` npm task. 11ty syntax is not allowed (see `no-unescaped-template-tags.js`).  

I checked the `remark-cli` docs, I couldn't find a way to disable the warning for just a line (e.g. `eslint-disable-next-line`). The documentation suggests me to ignore the files in `.remarkignore`. 

At this point, I can either fallback to the previous regex version or ignore this file. Your call.

tools/devtools/utils.mjs
```diff
@@ -0,0 +1,262 @@
+import {readFile, writeFile, mkdir} from 'fs/promises';
+// eslint-disable-next-line node/no-extraneous-import
```
jecfish, line 2: Will remove this line if we add `nunjucks` as direct dependecy.

tools/devtools/utils.mjs
```diff
@@ -0,0 +1,262 @@
...
+}
+
+/**
+ * Generate What's New in DevTools blog post banner
+ * @param {String} version - release version
+ * @param {String[]} langs - languages you want to generate
+ */
+export async function createWndtBanners(version, langs) {
+  for (const lang of langs) {
+    const output = nunjucks.render(bannerTemplate, {
```
jecfish, line 83: Inspired by your previous review comment, I change most of the regex replace syntax with [nunjucks render](https://mozilla.github.io/nunjucks/api.html#render). 

I can't use 11ty syntax as I need to pre-render it beforehand.

tools/devtools/utils.mjs
```diff
@@ -0,0 +1,262 @@
+import {readFile, writeFile, mkdir} from 'fs/promises';
+// eslint-disable-next-line node/no-extraneous-import
+import nunjucks from 'nunjucks';
+import {Octokit} from 'octokit';
+import {dirname} from 'path';
+
+nunjucks.configure({
+  autoescape: false,
+  tags: {blockStart: '<%', blockEnd: '%>'},
```
jecfish, line 9: If you are wondering why I change the block syntax, it's because I don't want nunjucks to render `{% include }`, 11ty should handle it.

Using `{% raw}` to avoid rendering doesn't work in my case because I render the files twice. With `{% raw}`, the second render doesn't work.

jecfish:
@matthiasrohmer, ready for review again!

jecfish:
Hi @matthiasrohmer if you can have a look

jecfish:
Ping @matthiasrohmer for review.

tools/devtools/utils.mjs
```diff
@@ -0,0 +1,280 @@
...
+
+/**
+ * Create Github Issues for translators to work on
+ * @param {String} version - release version
+ * @param {String} dueDate - due date to translate
+ * @param {String[]} langs - languages you want to generate
+ * @param {String} auth - GitHub Token
+ * @param {Object} translators - Translators of different locales { en: 'githubHandler' }
+ */
+export async function createGitHubIssues(
```
matthiasrohmer, line 240: I wonder if it's worth safe guarding this to prevent issues from getting created twice if the script is run a second time with the same variables by accident?

tools/devtools/utils.mjs
matthiasrohmer: Doesn't hurt much, as it's scoped to your tool anyway. Just for my own curiosity: the `{% raw %}` tags would be gone after the first render pass, no? Or am I missing something?

tools/devtools/utils.mjs
matthiasrohmer: Yes, let's add Nunjucks as a direct dependency.

package.json
```diff
@@ -93,6 +94,7 @@
     "@types/js-yaml": "^4.0.3",
     "@types/node": "^14.17.32",
     "@types/node-fetch": "^2.5.12",
+    "@types/nunjucks": "3.2.1",
```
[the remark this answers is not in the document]
matthiasrohmer: Yep pinning is good! npm would complain if there ever is a version mismatch. Also add nunjucks.

.remarkignore
matthiasrohmer: Is this still valid? Now all templates seem to live in tools/devtools/templates. Or is remark complaining after generation?

tools/devtools/utils.mjs
jecfish: I have a safe guard when calling this function. The CLI will explicitly ask if we want to generate, see `wndt-translation-bootstrap` - `readline.question('Do you want to generate GitHub issues? y/[n]:` 

The worst case is we have duplicated issues that we need to manually clean up.

matthiasrohmer: Yep, I have seen this one 🙂 I just can imagine myself seeing pressing `y` in a rush - that's where I was coming from. But maybe even calling out the configured version in the prompt again would save me already.

tools/devtools/templates/new-in-devtools-image-snippet.md
```diff
@@ -0,0 +1,25 @@
+<!-- 1. Save the script below as DevTools snippet. See https://developer.chrome.com/docs/devtools/javascript/sources/#snippets. -->
+<!-- 2. Upload the images in _temp folder manually -->
```
matthiasrohmer, line 2: I was just trying this out and wondered where to upload them?

matthiasrohmer:
Again, sorry for the long wait. Had another look and fiddled around with it. I don't quite get the idea of the image flow yet, but the others look fine to me 🙂

tools/devtools/templates/new-in-devtools-image-snippet.md
jecfish: https://web-dev-uploads.web.app/uploader

jecfish: It would be nice if I can just upload it using firebase api and get the url. But I don't have capacity to dig into that.

tools/devtools/utils.mjs
jecfish: yes, it will go after the first pass. I need two pass.

tools/devtools/utils.mjs
jecfish: Add version number and force it to be capital Y for additional safeguard.

jecfish (commented):
Resolve all the comments. Please take a look!

matthiasrohmer (approved):
Nice! LGTM.

matthiasrohmer:
Resolved the dependency issues. Needs content team approval, because Markdown files have been changed.

rachelandrew (approved):

PR #3768, opened 2022-09-19
[lighthouse] migrate accessibility content from web.dev

brendankenny:
Adds a Lighthouse accessibility docs section and migrates the last of the content from web.dev (the rest of the accessibility audits were redirected to axe docs in https://github.com/GoogleChrome/web.dev/pull/8654).

- all the audit articles only had a change to the `layout` and updated links where needed
- `scoring/index.md` - I regenerated the accessibility weight table because it was super out of date. I don't know if there's a ton of value in seeing the audit weights (we don't have a similar page for best practices/SEO, for instance), but there are ~1500 visitors a month, so we can just keep the page.


site/_data/docs/lighthouse/toc.yml
```diff
@@ -1,13 +1,16 @@
 - url: /docs/lighthouse/overview/
-- url: /docs/lighthouse/seo/
```
brendankenny, line 2: rearranged this so that the TOC is in the same order as the Lighthouse report (perf, a11y, best practices, seo, pwa)

site/_data/docs/lighthouse/toc.yml
```diff
@@ -1,13 +1,16 @@
...
-  title: i18n.docs.seo.intro
-  description: i18n.docs.seo.overview
-- url: /docs/lighthouse/best-practices/
-  title: i18n.docs.best-practices.intro
-  description: i18n.docs.best-practices.overview
 - url: /docs/lighthouse/performance/
   title: i18n.docs.performance.intro
-  description: i18n.docs.performance.overview
+  description: i18n.docs.performance.intro_description
```
brendankenny, line 4: also switched all these to `intro_description`, which seems like what was intended (shorter blurb for the TOC, slightly longer overview on the individual pages)

site/en/docs/lighthouse/accessibility/custom-control-roles/index.md
```diff
@@ -0,0 +1,75 @@
...
+with an actual `<button>`.
+
+If you must keep the `<div>`,
+then add `role="button"` and `aria-pressed="false"` to the `<div>`:
+
+{% Glitch {
+  id: 'gorgeous-raven',
+  path: 'example.html:13:39',
+  height: 346
+} %}
```
brendankenny, line 56: unfortunately several of these rely on a hosted form of chromevox-lite ([in this example](https://gorgeous-raven.glitch.me/) at `https://robdodson.github.io/chromevox-lite/dist/index.min.mjs`) which doesn't exist in that repo anymore and doesn't seem to be hosted elsewhere now. If we can't find a replacement, I guess we'll just delete?

site/en/docs/lighthouse/accessibility/index.md
```diff
@@ -0,0 +1,6 @@
+---
+layout: 'layouts/project-landing.njk'
+title: 'Accessibility Audits'
+description: 'These checks highlight opportunities to improve the accessibility of your web app. Only a subset of accessibility issues can be automatically detected so manual testing is also encouraged.'
```
brendankenny, line 4: since the TOC is already generated per locale, it would be nice if this could use `i18n.docs.lighthouse-accessibility.overview` so you wouldn't need an `index.md` per locale

site/en/docs/lighthouse/accessibility/custom-control-roles/index.md
malchata: I suppose if that's what we need to do until we can find an interim solution. Is there a way to comment out nunjucks syntax? Perhaps it could be fixed later.

site/_data/docs/lighthouse/toc.yml
malchata: Works for me!

malchata (approved):
Cursory look, but LGTM

site/en/docs/lighthouse/accessibility/custom-control-roles/index.md
brendankenny: > I suppose if that's what we need to do until we can find an interim solution. Is there a way to comment out nunjucks syntax? Perhaps it could be fixed later.

Commented it out and left an short explanation.

I was wrong and the other glitches are actually fine, it's just this one, so it's not a huge deal if we can't get a replacement.

Issue #3705, opened 2022-09-14
Slow modules in _data - we should refactor some modules


devnook:
While investigating how to speed up preview generation (PR #3632 ) I noticed that there are 3 slow culprits in Eleventy _data generation:

postsData
chromeApiNamespaces
workboxApiModules (less slow)

Especially postsData seem to read every template to extract info from it - this could be probably dome in a collection instead.

All of them depend on files manipulation, so slow I/O can be the contributing factor. We should look into these modules and see if we can optimize/remove/streamline data generation in them.

PR #3773, opened 2022-09-20
Remove redundancy of the use of the card-authors macro

devnook:
Related to #3705 

The aim is to remove / replace entirely modules that depend on _data/postsData. 
It removes an unnecessary import and replaces frontmatterCard with blogCard.


matthiasrohmer (approved):
LGTM!

PR #3778, opened 2022-09-20
Remove postsData dependency for cardAuthor macro

devnook:


matthiasrohmer (approved):
Works as before, nice 🙂

PR #3783, opened 2022-09-21
Add E2E test setup using freshly built fixtures

### LINKED ISSUE #3752: Add E2E test setup for features that have build output impact
Currently the site has three test suites which are ran using `ava`:

- external: https://github.com/GoogleChrome/developer.chrome.com/tree/main/tests/external
- server: https://github.com/GoogleChrome/developer.chrome.com/tree/main/tests/server
- site: https://github.com/GoogleChrome/developer.chrome.com/tree/main/tests/site

All of them are aiming to be unit tests which works fine for external and server features but falls short for most of the features that are tried to be tested with the site suite. For example take this PR: https://github.com/GoogleChrome/developer.chrome.com/pull/3716/files#diff-97db0848b13415ba7b07d1f0463454d877d9e2a24b5058ccc116aea1737ff573R6

This PR updates a test that has 11ty functionality encoded and failed due to the tested method switching to an updated API in 11ty. That also happens in some other tests, like:

https://github.com/GoogleChrome/developer.chrome.com/blob/ffb3d78f6a021dfbcb556fd3285b103a65addf15/tests/site/_data/lib/find.js#L4-L9

Encodes the Collection API, which might change with an 11ty update which wouldn't be caught by the test.

https://github.com/GoogleChrome/developer.chrome.com/blob/ffb3d78f6a021dfbcb556fd3285b103a65addf15/tests/site/_data/lib/breadcrumbs.js#L11-L18

Encodes the collection item API

In order to implement tests for https://github.com/GoogleChrome/developer.chrome.com/pull/3692 and https://github.com/GoogleChrome/developer.chrome.com/pull/3686 which both depend on certain 11ty internals (template rendering, data cascade, ordering) I would like to introduce a new test suite, gradually replacing tests that have 11ty logic encoded/abstracted.

This new test suite is a second 11ty project living at tests/e2e/.eleventy.js, inheriting from the original config and only containing a few pages that resemble our structure at critical points. For example: a directory that is only partially translated and shows all content an overview page, nested shortcodes, shortcodes in Markdown, ...

I think this will make testing easier, as tests are quicker to write as you don't need to implement an abstraction or heavily isolate your actual functionality, and it also tests the actual output we are producing.

matthiasrohmer:
Resolves #3752. This adds a new testing suite called e2e. It includes a minimal second 11ty setup, inheriting from the original one, but building only a small set of specially created test pages.

Currently this set of test pages holds various usages of the `{% Aside %}` shortcode, to be able to make sure it renders correctly and by this has the potential to replace the test at https://github.com/GoogleChrome/developer.chrome.com/blob/main/tests/site/_shortcodes/Aside.js

The idea of the setup is to be able to test and ensure correct output for logic that is heavily tangled with 11ty internals, like rendering pages from collections.


tests/e2e/utils.js
```diff
@@ -0,0 +1,42 @@
+const fs = require('fs/promises');
```
devnook, line 1: Please add license

tests/e2e/tests/shortcodes/aside.js
```diff
@@ -0,0 +1,31 @@
+const test = require('ava');
```
devnook, line 1: Please add license

package.json
```diff
@@ -34,6 +34,7 @@
     "stage": "NODE_ENV=production npm run production && gcloud app deploy --project=chrome-apps-doc-staging-2020 --quiet",
     "start": "node ./server",
     "test:watch": "ava --watch",
+    "pretest": "cd tests/e2e/fixtures && eleventy",
```
devnook, line 37: is it used somewhere? Or is it intended to be used manually?

site/_includes/layouts/blank.njk
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
```
devnook, line 1: Please add comment of why this template exists and where is used. Or maybe consider renaming to test.njk or similar?

devnook (changes requested):
Reviewed

package.json
matthiasrohmer: You can prefix any script name with `pre` or `post` to make them run with the script with the unprefixed name. So this would automatically run when `npm run test` is called. Also see https://docs.npmjs.com/cli/v8/using-npm/scripts#pre--post-scripts

site/_includes/layouts/blank.njk
matthiasrohmer: Renamed to test and added a comment header.

matthiasrohmer:
License headers added and template renamed.

devnook (approved):

PR #3793, opened 2022-09-22
fix typo in chrome extension documentation

### LINKED ISSUE #3792: typo in chrome extension documentation
the prop name of ScriptInjection object should be `func` instead of `function` [here](https://developer.chrome.com/docs/extensions/mv3/content_scripts/#:~:text=id%20%7D%2C-,function,-%3A%20injectedFunction%0A%20%20%7D)

hoishing:
the prop name of ScriptInjection object should be `func` instead of `function`

Fixes #3792


AmySteam (approved):
@hoishing Thanks for this PR!

@jpmedley This is technically correct. We can merge this 👍

jpmedley (approved):

PR #3794, opened 2022-09-22
[Post] Inside the container query polyfill

devknoll:
Add the "Inside the container query polyfill" blog post, as well as support for Before/After labels for the Compare component.

### DIFF
--- a/site/en/blog/inside-the-container-query-polyfill/index.md
+++ b/site/en/blog/inside-the-container-query-polyfill/index.md
@@ -9,15 +9,15 @@ authors:
   - gbmonaco
 ---
 
-# Inside the container query polyfill
+## Inside the container query polyfill
 
 [Container queries](https://www.youtube.com/watch?v=gCNMyYr7F6w) are a new CSS feature that enables you to write styling logic that targets features of a parent element (for example, its width or height) to style its children. Recently, a [big update](https://developer.chrome.com/blog/cq-polyfill/) to the [polyfill](https://github.com/GoogleChromeLabs/container-query-polyfill) was released, coinciding with support landing in browsers.
 
 In this post, you will be able to take a peek inside how the polyfill works, the challenges it overcomes, and the best practices when using it to provide a great user experience for your visitors.
 
-# Under the hood
+## Under the hood
 
-## Transpilation
+### Transpilation
 
 When the CSS parser inside a browser encounters an unknown at-rule, like the brand new `@container` rule, it will simply discard it as if it never existed. Therefore, the first and most important thing the polyfill must do is transpile an `@container` query into something that won’t be discarded.
 
@@ -122,7 +122,7 @@ This has a number of benefits:
 
 During transpilation, the polyfill will replace this dummy with the attribute selector with the same specificity. To avoid any surprises, the polyfill uses both selectors: the original source selector is used to determine if the element should receive the polyfill attribute, and the transpiled selector is used for styling.
 
-## Pseudo-elements
+### Pseudo-elements
 
 One question you might be asking yourself is: if the polyfill sets some `cq-XYZ` attribute on an element to include the unique container ID `123`, how can pseudo-elements, which can’t have attributes set on them, be supported?
 
@@ -182,7 +182,7 @@ To correct this, the polyfill actually uses _two_ attributes: one that can only 
 
 Since a container will never apply the first attribute (`cq-XYZ-A`) to itself, the first selector will only match if a _different_ parent container has met the container conditions and applied it.
 
-## Container relative units
+### Container relative units
 
 Container queries also come with [a few new units](https://www.w3.org/TR/css-contain-3/#container-lengths) that you can use in your CSS, such as `cqw` and `cqh` for 1% of the width and height (respectively) of the closest appropriate parent container. To support these, the unit is transformed into a `calc(...)` expression using CSS Custom Properties. The polyfill will set the values for these properties via inline styles on the container element.
 
@@ -218,7 +218,7 @@ There are also logical units, like `cqi` and `cqb` for inline size and block siz
 
 Now, the units can be transformed into the appropriate CSS Custom Property just as before.
 
-## Properties
+### Properties
 
 Container queries also add a few new CSS properties like [`container-type`](https://www.w3.org/TR/css-contain-3/#container-type) and [`container-name`](https://www.w3.org/TR/css-contain-3/#container-name). Since APIs like `getComputedStyle(...)` can’t be used with unknown or invalid properties, these are also transformed to CSS Custom Properties _after being parsed_. If a property can’t be parsed (for example, because it contains an invalid or unknown value), it’s simply left alone for the browser to handle.
 
@@ -267,7 +267,7 @@ By default, CSS Custom Properties are inherited, meaning that e.g. any child of 
 }
 ```
 
-# Best practices
+## Best practices
 
 While it’s expected that most visitors will be running browsers with built-in container query support sooner rather than later, it’s still important to give your remaining visitors a good experience.
 
@@ -297,7 +297,7 @@ This approach is recommended for a number of reasons:
 * After the polyfill loads, this `@supports` condition will stop passing, and your content will be revealed.
 * On browsers with built-in support for container queries, the condition will never pass, and so the page will be displayed on the first-paint as expected.
 
-# Conclusion
+## Conclusion
 
 If you're interested in using container queries on older browsers, give the [polyfill](https://github.com/GoogleChromeLabs/container-query-polyfill) a try. Don't hesitate to [file an issue](https://github.com/GoogleChromeLabs/container-query-polyfill/issues/new) if you run into any problems.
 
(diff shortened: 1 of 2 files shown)


rachelandrew (changes requested):
This looks good but we need a thumbnail and/or hero image. 

We often just find something from Unsplash and add a credit at the bottom of the article to the photographer. https://unsplash.com/

site/en/blog/inside-the-container-query-polyfill/index.md
```diff
@@ -0,0 +1,301 @@
...
+date: 2022-10-01
+hero: 'image/0Nc53fqpS4cgxOG9PCAyLwS0QwB3/75EKsp0A2SyU8bOWttSd.jpg'
+alt: "An assortment of neatly arranged drafting tools. In the center is a blank sheet of graphing paper."
+tags:
+  - css
+authors:
+  - gbmonaco
+---
+
+[Container queries](https://www.youtube.com/watch?v=gCNMyYr7F6w) are a new CSS feature that enables you to write styling logic that targets features of a parent element (for example, its width or height) to style its children. Recently, a [big update](https://developer.chrome.com/blog/cq-polyfill/) to the [polyfill](https://github.com/GoogleChromeLabs/container-query-polyfill) was released, coinciding with support landing in browsers.
```
rachelandrew, line 14:
```suggestion
[Container queries](https://www.youtube.com/watch?v=gCNMyYr7F6w) are a new CSS feature that enables you to write styling logic that targets features of a parent element (for example, its width or height) to style its children. Recently, a [big update](/blog/cq-polyfill/) to the [polyfill](https://github.com/GoogleChromeLabs/container-query-polyfill) was released, coinciding with support landing in browsers.
```

site/en/blog/inside-the-container-query-polyfill/index.md
```diff
@@ -0,0 +1,301 @@
...
+
+## Under the hood
+
+### Transpilation
+
+When the CSS parser inside a browser encounters an unknown at-rule, like the brand new `@container` rule, it will simply discard it as if it never existed. Therefore, the first and most important thing the polyfill must do is transpile an `@container` query into something that won’t be discarded.
+
+{% Aside %}
+This sort of transpilation often occurs at build-time, as part of a post-processing step. However, the polyfill does this at runtime instead, for two reasons: you and your visitors. With support for container queries already in [Chromium 105](https://developer.chrome.com/blog/new-in-chrome-105/) and [Safari 16](https://webkit.org/blog/13152/webkit-features-in-safari-16-0/), it’s expected that most of your visitors will likely be running browsers with built-in support sooner rather than later. Offline transpilation would effectively tax you (for having to maintain this offline post-processing step) _and_ your visitors (for having to download transpiled code that may never be used).
```
rachelandrew, line 25:
```suggestion
This sort of transpilation often occurs at build-time, as part of a post-processing step. However, the polyfill does this at runtime instead, for two reasons: you and your visitors. With support for container queries already in [Chromium 105](/blog/new-in-chrome-105/) and [Safari 16](https://webkit.org/blog/13152/webkit-features-in-safari-16-0/), it’s expected that most of your visitors will likely be running browsers with built-in support sooner rather than later. Offline transpilation would effectively tax you (for having to maintain this offline post-processing step) _and_ your visitors (for having to download transpiled code that may never be used).
```

site/en/blog/inside-the-container-query-polyfill/index.md
```diff
@@ -0,0 +1,301 @@
...
+{% endAside %}
+
+The first step in transpilation is to convert the top-level `@container` rule into an [@media](https://developer.mozilla.org/en-US/docs/Web/CSS/@media) query. This mostly ensures that the content remains grouped together. For example, when using [CSSOM](https://developer.mozilla.org/en-US/docs/Web/API/CSS_Object_Model) APIs and when viewing the CSS source.
```
rachelandrew, line 28:
```suggestion
The first step in transpilation is to convert the top-level `@container` rule into an [@media](https://developer.mozilla.org/docs/Web/CSS/@media) query. This mostly ensures that the content remains grouped together. For example, when using [CSSOM](https://developer.mozilla.org/docs/Web/API/CSS_Object_Model) APIs and when viewing the CSS source.
```

site/en/blog/inside-the-container-query-polyfill/index.md
```diff
@@ -0,0 +1,301 @@
...
+```css
+@media all {
+  .card:where([cq-XYZ~="123"]) {
+    /* ... */
+  }
+}
+```
+{% endCompare %}
+
+Notice the use of the [`:where(...)`](https://developer.mozilla.org/en-US/docs/Web/CSS/:where) pseudo-class. Normally, including an additional attribute selector would increase the [specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity) of the selector. With the pseudo-class, the extra condition can be applied while preserving the original specificity. To see why this is crucial, consider the following example:
```
rachelandrew, line 68:
```suggestion
Notice the use of the [`:where(...)`](https://developer.mozilla.org/docs/Web/CSS/:where) pseudo-class. Normally, including an additional attribute selector would increase the [specificity](https://developer.mozilla.org/docs/Web/CSS/Specificity) of the selector. With the pseudo-class, the extra condition can be applied while preserving the original specificity. To see why this is crucial, consider the following example:
```

site/en/blog/inside-the-container-query-polyfill/index.md
```diff
@@ -0,0 +1,301 @@
...
+  }
+}
+```
+{% endCompare %}
+
+Since a container will never apply the first attribute (`cq-XYZ-A`) to itself, the first selector will only match if a _different_ parent container has met the container conditions and applied it.
+
+### Container relative units
+
+Container queries also come with [a few new units](https://www.w3.org/TR/css-contain-3/#container-lengths) that you can use in your CSS, such as `cqw` and `cqh` for 1% of the width and height (respectively) of the closest appropriate parent container. To support these, the unit is transformed into a `calc(...)` expression using [CSS Custom Properties](https://developer.mozilla.org/en-US/docs/Web/CSS/--*). The polyfill will set the values for these properties via inline styles on the container element.
```
rachelandrew, line 179:
```suggestion
Container queries also come with [a few new units](https://www.w3.org/TR/css-contain-3/#container-lengths) that you can use in your CSS, such as `cqw` and `cqh` for 1% of the width and height (respectively) of the closest appropriate parent container. To support these, the unit is transformed into a `calc(...)` expression using [CSS Custom Properties](https://developer.mozilla.org/docs/Web/CSS/--*). The polyfill will set the values for these properties via inline styles on the container element.
```

site/en/blog/inside-the-container-query-polyfill/index.md
```diff
@@ -0,0 +1,301 @@
...
+{% Compare 'better', 'after' %}
+```css
+.card {
+  width: calc(10 * --cq-XYZ-cqw);
+  height: calc(10 * --cq-XYZ-cqh);
+}
+```
+{% endCompare %}
+
+There are also logical units, like `cqi` and `cqb` for inline size and block size (respectively). These are a little bit more complicated, because the inline and block axes are determined by the [`writing-mode`](https://developer.mozilla.org/en-US/docs/Web/CSS/writing-mode) of _the element using the unit_, not the element being queried. To support this, the polyfill applies an inline style to any element whose `writing-mode` differs from its parent.
```
rachelandrew, line 199:
```suggestion
There are also logical units, like `cqi` and `cqb` for inline size and block size (respectively). These are a little bit more complicated, because the inline and block axes are determined by the [`writing-mode`](https://developer.mozilla.org/docs/Web/CSS/writing-mode) of _the element using the unit_, not the element being queried. To support this, the polyfill applies an inline style to any element whose `writing-mode` differs from its parent.
```

site/en/blog/inside-the-container-query-polyfill/index.md
```diff
@@ -0,0 +1,301 @@
...
+{% Compare 'better', 'after' %}
+```css
+.card {
+  --cq-XYZ-container-name: card-container;
+  --cq-XYZ-container-type: inline-size;
+}
+```
+{% endCompare %}
+
+These properties are transformed whenever they’re discovered, allowing the polyfill to play nicely with other CSS features like [`@supports`](https://developer.mozilla.org/en-US/docs/Web/CSS/@supports). This functionality is the basis of the best practices for using the polyfill, as covered below.
```
rachelandrew, line 235:
```suggestion
These properties are transformed whenever they’re discovered, allowing the polyfill to play nicely with other CSS features like [`@supports`](https://developer.mozilla.org/docs/Web/CSS/@supports). This functionality is the basis of the best practices for using the polyfill, as covered below.
```

rachelandrew (commented):
Updating links for the linter

devknoll:
@rachelandrew anything else I can do to help move this forward? 🙏

rachelandrew (approved):

site/en/blog/inside-the-container-query-polyfill/index.md
```diff
@@ -0,0 +1,301 @@
+---
+title: "Inside the container query polyfill"
+description: "Learn how the container query polyfill brings container query support to older browsers."
+layout: 'layouts/blog-post.njk'
+date: 2022-10-01
```
rachelandrew, line 5:
```suggestion
date: 2022-10-10
```

rachelandrew:
@devknoll sorry for the delay, let's publish this on Monday if that works for you?

devknoll:
No worries! Any time should be great 👍

PR #3829, opened 2022-09-26
DevTools 106 > console.createTask

sofiayem:
Documented https://developer.chrome.com/blog/devtools-modern-web-debugging/#linked-stack-traces

### DIFF
--- a/site/en/docs/devtools/console/api/index.md
+++ b/site/en/docs/devtools/console/api/index.md
@@ -73,6 +73,24 @@ console.countReset();
 console.countReset('coffee');
 ```
 
+## console.createTask(name) {: #createtask }
+
+{% Aside 'gotchas' %}
+This method implements the Async Stack Tagging API. If you're using a framework that handles the scheduling and async execution, you don't need to use this method. In that case, it is up to the framework to implement the API.
+{% endAside %}
+
+Returns a `Task` instance that associates the current async stack trace with the created `task`. You can later use the `task` to run its content (`f` in the following example). The `task.run(f)` executes an arbitrary payload and forwards the return value back to the caller.
+
+```js
+// Task creation
+const task = console.createTask(name);
+
+// Task execution
+task.run(f); // instead of f();
+```
+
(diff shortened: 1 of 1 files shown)


sofiayem:
Hi @victorporof! Can you please fact-check this?

site/en/docs/devtools/console/api/index.md
```diff
@@ -73,6 +73,24 @@ console.countReset();
 console.countReset('coffee');
 ```
 
+## console.createTask(name) {: #createtask }
+
+{% Aside 'gotchas' %}
+This method implements the Async Stack Tagging API. If you're using a framework that handles the scheduling and async execution, you don't need to use this method. In that case, it is up to the framework to implement the API.
```
victorporof, line 79: I'd word this as "This method is known as the Async Stack Tagging API. It you're using a framework or abstraction that already started using this under the hood to keep track of scheduling and async execution, then you don't need to use this method directly."

site/en/docs/devtools/console/api/index.md
```diff
@@ -73,6 +73,24 @@ console.countReset();
...
+{% endAside %}
+
+Returns a `Task` instance that associates the current async stack trace with the created `task`. You can later use the `task` to run its content (`f` in the following example). The `task.run(f)` executes an arbitrary payload and forwards the return value back to the caller.
```
victorporof, line 82: I'd word this as "Returns a `Task` instance that associates the current stack trace with the created `task` object. You can later use this `task` object to run a function (`f` in the following example). The `task.run(f)` executes an arbitrary payload and forwards the return value back to the caller."

victorporof (changes requested):
Looks good with some wording nits. Thanks!

site/en/docs/devtools/console/api/index.md
victorporof: I'd remove "In that case, it is up to the framework to implement the API.". Frameworks don't _implement_ this API, they use it. We (DevTools) implemented the API.

victorporof (approved):

PR #3843, opened 2022-09-28
Fix intents zxing link

Viicos:

- The [previous link](https://code.google.com/p/zxing/source/browse/trunk/android/AndroidManifest.xml#97) is doing an incorrect redirect. You might want to have a custom code.google.com link instead.


Pedro-vk (commented):
Thanks for the fix!
Hopefully will be fixed soon :+1:

sofiayem (approved):

Viicos:
Hi @sofiayem, thanks for the merge! Was just wondering if the date was labeled correctly on the docs page?

See for example https://developer.chrome.com/docs/multidevice/android/intents/#see-also:

```Last updated: Friday, February 28, 2014```. Maybe `Created` would be the correct one?

sofiayem:
@Viicos As per recent change of policies, we don't update the date for small changes such as this.

PR #3870, opened 2022-09-29
Add example Curl command to CrUX API docs

tunetheweb:

- Adds example Curl command line as discussed on CrUX chat.

CC: @powdercloud @paulirish


site/en/docs/crux/api/index.md
```diff
@@ -250,7 +251,17 @@ Queries are submitted as JSON objects via a POST request to `https://chromeuxrep
 }
 ```
 
-Page-level data is available through the API by passing a `url` property in the query:
+For example, this can be called from `curl` with the following command line (replacing API_KEY with your key):
```
malchata, line 254:
```suggestion
For example, this can be called from `curl` with the following command line (replacing `API_KEY` with your key):
```

malchata (approved):
LGTM

PR #3907, opened 2022-10-04
Fix navigation tracking on pageviews

tunetheweb:
Noticed we're not tracking page navigation types except back_forward_cache (which has special handling).

This PR implements same stuff we have on web.dev: https://github.com/GoogleChrome/web.dev/blob/main/src/site/_includes/partials/analytics.njk

### DIFF
--- a/site/_data/analytics.json
+++ b/site/_data/analytics.json
@@ -5,5 +5,5 @@
     "NAVIGATION_TYPE": "dimension2",
     "WEB_VITALS_DEBUG": "dimension3"
   },
-  "TRACKING_VERSION": "3.0"
+  "version": 4
 }
--- a/site/_includes/partials/script.js
+++ b/site/_includes/partials/script.js
@@ -39,7 +39,11 @@
     '{{ analytics.TRACKING_VERSION }}'
   );
   try {
-    ga('set', '{{ analytics.dimensions.NAVIGATION_TYPE }}', performance.getEntriesByType('navigation')[0].type);
+    ga(
+      'set',
+      '{{ analytics.dimensions.NAVIGATION_TYPE }}',
+      performance.getEntriesByType('navigation')[0].type
+    );
   } catch (error) {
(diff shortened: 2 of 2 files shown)


site/_includes/partials/script.js
```diff
@@ -38,6 +38,15 @@
     '{{ analytics.dimensions.TRACKING_VERSION }}',
     '{{ analytics.TRACKING_VERSION }}'
   );
+  try {
+    ga(
+      'set',
+      '{{ analytics.dimensions.NAVIGATION_TYPE }}',
+      performance.getEntriesByType('navigation')[0].type
```
philipwalton, line 45: Since we're already using the navigation type reported by the `web-vitals` library, to be consistent with that let's standardize on using kebab-case.

```suggestion
      performance.getEntriesByType('navigation')[0].type.replace(/_/g, '-')
```

philipwalton (approved):
LGTM (with one comment)

philipwalton:
Actually, you should probably also bump the tracking version...

site/_includes/partials/script.js
tunetheweb: I noticed it was inconsistent on web.dev and wondered should we change bfcache one! But happy to change this instead.

philipwalton: Yeah, I think we should probably also change web.dev...

Jdzimmo:
Yes sir...

tunetheweb:
> Actually, you should probably also bump the tracking version...

Done now. I also cleaned it up a little (single version number rather than decimal, and rename variable to avoid confusion) to bring it more inline with web.dev

site/_data/analytics.json
```diff
@@ -5,5 +5,5 @@
     "NAVIGATION_TYPE": "dimension2",
     "WEB_VITALS_DEBUG": "dimension3"
   },
-  "TRACKING_VERSION": "3.0"
+  "version": 4
```
tunetheweb, line 8: You happy enough with this change @philipwalton ?

PR #3912, opened 2022-10-05
Enhanced select

johnrich85:

Adds a web component that progressively enhances the native select element.

Design:

image

To test:

```
<enhanced-select>
      <select id="aaa" name="aaa" >
        <option value="1">One</option>
        <option value="2">Two</option>
        <option value="3">Three</option>
        <option value="4">Four</option>
        <option value="5">Five</option>
      </select>
</enhanced-select>
```


matthiasrohmer:
Haven't had a detailed look yet, but it might be worth looking at https://github.com/GoogleChrome/web.dev/pull/7177/files#diff-747808994e25221313c98080756bef4444f8969dda0b2ff1ae3a4ed39e27cde8R90 which tries to implement the same for web.dev and seems to handle a lot of the a11y pitfalls gracefully.

johnrich85:
Awesome, thanks @matthiasrohmer - only wish I found this earlier!  Keyboard navigation/a11y is one of the next items on my list so will definitely take a look before moving on

devnook:
Hi John, can you add the select to the https://developer.chrome.com/docs/handbook/style-guide/ page? It will serve both as testing page and later as documentation.

johnrich85:
Sure, ill add that to my list.

site/_js/main.js
```diff
@@ -27,6 +27,7 @@ import './web-components/share-button';
 import './web-components/web-tabs';
 import './web-components/language-select';
 import './third-party/announcement-banner/announcement-banner';
+import './web-components/enhanced-select';
```
devnook, line 30: This is not used on every page, so it is adding bloat to the main js unnecessarily.

site/_js/web-components/enhanced-select.js
```diff
@@ -0,0 +1,424 @@
...
+
+export class EnhancedSelect extends BaseElement {
+  static get formAssociated() {
+    return true;
+  }
+
+  constructor() {
+    super();
+
+    // @ts-ignore
```
devnook, line 38: What are exactly you ignoring here?

devnook (commented):
Thanks @johnrich85 ! This looks good. Can you please add a test for this component? We try to get test coverage for new addons.

site/_js/web-components/enhanced-select.js
johnrich85: I'm getting an erroneous  'property does not exist' error without the ignore statement...

devnook (approved):

PR #3932, opened 2022-10-07
Add /preview url and handler

devnook:


server/render.js
```diff
@@ -0,0 +1,131 @@
...
+/**
+ * Escapes and formats frontmatter provided by user to 11ty format.
+ * @param {string} inputFm Frontmatter string
+ */
+const parseFrontmatter = inputFm => {
+  const /** @type {any} */ parsedFm = yaml.load(inputFm) || {};
+  const date = parsedFm.date?.toISOString ? parsedFm.date : new Date();
+  return {
+    layout: xss(parsedFm.layout) || 'layouts/blog-post.njk',
+    date: xss(date.toISOString().split('T')[0]),
```
matthiasrohmer, line 63: That probably doesn't need to be piped through `xss` as `Date` wouldn't produce malicious strings, no?

server/render.js
```diff
@@ -0,0 +1,131 @@
...
+const supportedLayouts = ['layouts/blog-post.njk', 'layouts/doc-post.njk'];
+
+const renderHandler = async (req, res) => {
+  // Check sign in
+  const userId = await verifyLogin(req.body.t).catch(() => {
+    res.sendStatus(403);
+  });
+  if (!userId) {
+    return;
+  }
```
matthiasrohmer, line 79:
```suggestion
  // Check sign in
  try {
    const userId = await verifyLogin(req.body.t);
    if (!userId) {
      return res.sendStatus(401);
    }
  } catch (e) {
    return res.sendStatus(401);
  }
```

I think we shouldn't mix async/await and .then/.catch, 403 assumes we know the user's identity, so 401 might be a better fit and 78 wasn't sending a response.

server/render.js
```diff
@@ -0,0 +1,131 @@
...
+
+  try {
+    const elev = new Eleventy(inputPath, 'dist', {
+      configPath: '.eleventy.js',
+      inputDir: './site',
+      config: function (eleventyConfig) {
+        eleventyConfig.addPlugin(EleventyRenderPlugin);
+        eleventyConfig.addGlobalData('fm', fm);
+        eleventyConfig.addGlobalData('body', body);
+        eleventyConfig.addGlobalData('date', fm.date);
```
matthiasrohmer, line 103: Not sure what is cleaner/faster/easier to read but you could probably also manually write them to `this.ctx` in the render shortcode...?

matthiasrohmer (approved):
Left some questions and suggestions, but nothing that stands in the way of the actual functionality 🙂

server/render.js
```diff
@@ -0,0 +1,131 @@
...
+  return payload && payload['sub']; // userid
+}
+
+/**
+ * Escapes and formats frontmatter provided by user to 11ty format.
+ * @param {string} inputFm Frontmatter string
+ */
+const parseFrontmatter = inputFm => {
+  const /** @type {any} */ parsedFm = yaml.load(inputFm) || {};
+  const date = parsedFm.date?.toISOString ? parsedFm.date : new Date();
```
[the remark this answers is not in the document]
devnook: inputFm comes from the body of a POST request, so it is defined by the user (request origin). We don't control it, so the user can put a real Date or random string in the date field.

server/render.js
devnook: right:)

server/render.js
```diff
@@ -0,0 +1,131 @@
...
+const renderHandler = async (req, res) => {
+  // Check sign in
+  const userId = await verifyLogin(req.body.t).catch(() => {
+    res.sendStatus(403);
+  });
+  if (!userId) {
+    return;
+  }
+
+  process.env.ELEVENTY_IGNORE_EXTENSIONS = 'true';
```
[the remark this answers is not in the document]
devnook: I think this should be solvable, but not in this PR. I added an issue for it: https://github.com/GoogleChrome/developer.chrome.com/issues/4000

server/render.js
devnook: Actually it does not work when set directly on this.ctx object. I don't know why, but I'd say passing it as global data is consistent enough with the rest of the site.

PR #3940, opened 2022-10-07
Correct blog sort order

tunetheweb:
As noted on team chat currently sorted by alphabetical order, when should be reverse data order.

This seems to have been broken by #3686 but a lot of change in that so don't fully understand it to be honest. However forcing the sort order fixes the immediate issue.

Suggest @matthiasrohmer have a further look on Monday to make sure that it wasn't set to alphabetical for a reason.


matthiasrohmer (approved):

matthiasrohmer:
Approving and about to merge to get site back in a healthy state. The sorting by fileSlug was coming from this discussion https://github.com/GoogleChrome/developer.chrome.com/pull/3686#discussion_r974271226 with the idea of only relying on the fileSlug as a matching criteria for translated content instead of both date order and fileSlug. Will revisit on Monday.

PR #3869, opened 2022-09-29
Add screen sharing controls blog post + web platform documentation

beaufortfrancois:
FIX https://github.com/GoogleChrome/developer.chrome.com/issues/3672

Live preview: 
- https://deploy-preview-3869--developer-chrome-com.netlify.app/blog/avoiding-oversharing-when-screen-sharing/
- https://deploy-preview-3869--developer-chrome-com.netlify.app/docs/web-platform/screen-sharing-controls/
- https://deploy-preview-3869--developer-chrome-com.netlify.app/docs/web-platform/#shipped
- https://deploy-preview-3869--developer-chrome-com.netlify.app/authors/eladalon/

cc @eladalon1983

PR #3947, opened 2022-10-08
Update index.md

guest271314:
Notify users that getDisplayMedia() does not capture speechSythesis.speak() executed on the same Tab, and when echoCancellation: false is not set channelCount of MediaStreamTrack will be 1.

Fixes #SOME_ISSUE_NUMBER

-
-
-


beaufortfrancois:
https://github.com/GoogleChrome/developer.chrome.com/pull/3869#issuecomment-1273078706 suggests we should not add this note.

@eladalon1983 Do you confirm?

eladalon1983:
> [#3869 (comment)](https://github.com/GoogleChrome/developer.chrome.com/pull/3869#issuecomment-1273078706) suggests we should not add this note.
> 
> @eladalon1983 Do you confirm?

I thank @guest271314 for his contribution, but I am still against this addition. I think a very small percentage of the readership will be aware of `speechSynthesis`, and so most of our readers will just be confused by this.

beaufortfrancois:
Thank you @eladalon1983!

PR #3950, opened 2022-10-09
Deferring `<script />` load for faster web page load.

sanjaiyan-dev:
[Deferring `<script />` load for faster web page load.](https://web.dev/efficiently-load-third-party-javascript/#defer)

_Sorry if I made any mistakes :(_

### DIFF
--- a/site/_includes/layouts/base.njk
+++ b/site/_includes/layouts/base.njk
@@ -65,8 +65,6 @@
       })();
     </script>
     {% endif %}
-    
-    {% block chromescripts %}{% endblock %}
   </head>
   <body>
     <div class="scaffold">
--- a/site/_includes/layouts/base.njk
+++ b/site/_includes/layouts/base.njk
@@ -65,6 +65,8 @@
       })();
     </script>
     {% endif %}
+    
+    {% block chromescripts %}{% endblock %}
   </head>
   <body>
     <div class="scaffold">
@@ -91,8 +93,6 @@
       {% endblock %}
     </div>
     {% include 'partials/cookie-banner.njk' %}
-
-    {% block chromescripts %}{% endblock %}
 
     {% block chromeanalytics %}{% endblock %}
   </body>
--- a/site/_includes/layouts/chrome-100.njk
+++ b/site/_includes/layouts/chrome-100.njk
@@ -5,9 +5,9 @@
 {% endblock %}
 
 {% block chromescripts %}
-  <script src="https://cdnjs.cloudflare.com/ajax/libs/gsap/3.9.1/gsap.min.js"></script>
-  <script src="https://cdnjs.cloudflare.com/ajax/libs/gsap/3.9.1/ScrollTrigger.min.js"></script>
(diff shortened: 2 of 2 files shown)


tophf:
This change doesn't seem to improve anything because `defer` keeps the scripts in practically the same position they were previously, i.e. just before DOMContentLoaded. The only negligible difference is now they are swapped with the analytics scripts.

devnook:
@tunetheweb ptal

site/_includes/layouts/base.njk
```diff
@@ -65,6 +65,8 @@
       })();
     </script>
     {% endif %}
+    
+    {% block chromescripts %}{% endblock %}
```
tunetheweb, line 69: I'd prefer not to move the `chromescripts` to the `<head>` since they are not critical and, even though they are defered, this can still cause them to be fetched earlier than necessary.

```suggestion
```

site/_includes/layouts/base.njk
```diff
@@ -92,8 +94,6 @@
     </div>
     {% include 'partials/cookie-banner.njk' %}
 
-    {% block chromescripts %}{% endblock %}
-
```
tunetheweb, line 96: Reverse this change.

site/_includes/layouts/chrome-100.njk
```diff
@@ -5,9 +5,9 @@
 {% endblock %}
 
 {% block chromescripts %}
-  <script src="https://cdnjs.cloudflare.com/ajax/libs/gsap/3.9.1/gsap.min.js"></script>
-  <script src="https://cdnjs.cloudflare.com/ajax/libs/gsap/3.9.1/ScrollTrigger.min.js"></script>
-  <script src="https://cdnjs.cloudflare.com/ajax/libs/gsap/3.9.1/ScrollToPlugin.min.js"></script>
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/gsap/3.9.1/gsap.min.js" defer></script>
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/gsap/3.9.1/ScrollTrigger.min.js" defer></script>
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/gsap/3.9.1/ScrollToPlugin.min.js" defer></script>
```
tunetheweb, line 10: This change seems fine, but a bit of a micro optimisation since it only affects that one page. But still, best to practice what we preach.

If you revert the other change, I'm happy to accept this change.

tunetheweb (changes requested):

site/_includes/layouts/base.njk
tunetheweb: You need to put that line back, since you're now not moving that to the head since you reversed the line above.

Also looks like things have changed so you have merge conflicts. It might be easiest to just close this PR and open a new one, from `main` with just the `defer` change?

sanjaiyan-dev: Ok 🤝

sanjaiyan-dev:
https://github.com/GoogleChrome/developer.chrome.com/pull/3950#discussion_r1094639664

PR #3968, opened 2022-10-10
Fix grammar in PNA deprecation trial post

johnathan79717:
The original sentence sounds weird. I changed it to how I guess it was intended to be.

cc @letitz @iVanlIsh


heyawhite (approved):

letitz:
My bad, the sentence was indeed awkward. I think this changes the meaning however. What I meant to say is that the header need only be served on resources when all the following conditions hold:

1) resource is a main resource / HTML document
2) for a navigation fetch
3) the HTML document is planning to make use of the feature

The new sentence's meaning is more "if 1 and 2, then 3". Which is (pedantically, I'll admit) not true, most documents will not make use of the deprecated feature.

PR #3970, opened 2022-10-10
Fix mangled aside.

jpmedley:
@AmySteam I'm taking a guess that this is what it's trying to say. If you don't know whether my update is correct, assign this to Simeon.

Fixes #SOME_ISSUE_NUMBER

-
-
-


AmySteam (approved):
Your revision does make the content of this aside more explicit. The information is correct 👍

PR #3980, opened 2022-10-11
Adds the new FilterSheet shortcode

johnrich85:

Adds a new shortcode: FilterSheet

Design:
image


devnook:
We try to keep shortcodes for authoring elements that can be embedded in the article body (so mainly used in .md files). This filters element is intended to be used in layout and is not a reusable component - there's no need to make it a shortcode.

I'd consider putting it directly in the calendar page, or extracting potentially reusable elements into a macro (but not a shortcode).

johnrich85:
Cheers @devnook. I've made the requested changes

devnook (approved):

PR #3998, opened 2022-10-12
Adds Checkbox group

johnrich85:

Adds new CheckboxGroup component.

Design:

image


devnook:
Please add tests for the new component and then ping for a review. Thanks!

johnrich85:
Apologies @devnook, this one ended up a little larger than I'd hoped after adding tests (as we didn't have any way to test WCs up until now)

site/_data/i18n/checkbox_group.yml
```diff
@@ -0,0 +1,5 @@
+select_all:
+  en: 'Select all'
+
+reset:
+    en: 'Reset'
```
devnook, line 5: indentation

site/_scss/blocks/_checkbox-group.scss
```diff
@@ -0,0 +1,44 @@
+checkbox-group {
+  label {
+    align-items: center;
+    color: get-color('grey-700');
```
devnook, line 4: I won't block on this, since we do not support dark theme as of now. However, we should start thinking of adding universal color definitions (like you use --primary below) to support theming instead of definitive colors when we add new css.

tests/e2e/tests/web-components/checkbox-group.js
```diff
@@ -0,0 +1,242 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+const test = require('ava');
+const {withPage, addPageScript} = require('../../puppeteer');
+const {html} = require('common-tags');
+
+test(
+  'checkbox group: can restrict the number of items shown',
```
devnook, line 22: I'm quite sure thiese should be unit tests, not e2e tests? I think the gist of our e2e tests is that it requires a build step (running 11ty). Please move it to https://github.com/GoogleChrome/developer.chrome.com/tree/main/tests/site/_js (or tell me if I'm wrong here)

devnook (changes requested):
Reviewed

site/_scss/blocks/_checkbox-group.scss
johnrich85: Cheers @devnook . Noted for future PRs

tests/e2e/tests/web-components/checkbox-group.js
johnrich85: Fair enough - I think that makes sense. I was kind of on the fence since were testing in browser/from a user perspective but also understand we are testing a single unit rather than the end product as a whole.

Unless i hear otherwise, I'll move the puppeteer/rollup stuff to the root test directory (so that it can be used by any test type if needs be) and the tests themselves to the /site/_js dir as mentioned.

package.json
```diff
@@ -34,7 +34,7 @@
     "stage": "NODE_ENV=production npm run production && gcloud app deploy --project=web-dev-staging --quiet",
     "start": "node ./server",
     "test:watch": "ava --watch",
-    "pretest": "cd tests/e2e/fixtures && eleventy",
+    "pretest": "cd tests && npx rollup -c && cd e2e/fixtures && eleventy",
```
devnook, line 37: @matthiasrohmer WDYT? This means we run both eleventy and rollup before e2e tests as well as unit tests. should we split it apart?

matthiasrohmer: With the new puppeteer testes introduced here I think that makes sense.

devnook (approved):

PR #4007, opened 2022-10-13
WIP: Add staticProxy middleware

devnook:


matthiasrohmer (commented):
Tested this further for the staging setup. This would only work with long existing buckets. By now there are org policies that disallow both object level and bucket level public serving 😞 

Got a variant going with using the Node.js library and response streaming. Will investigate serving times.

PR #4016, opened 2022-10-14
Update webdev-infra to 1.0.33

matthiasrohmer:
This updates webdev-infra to 1.0.33, which includes https://github.com/GoogleChrome/webdev-infra/pull/35.

With this change Cloud Build's deploy step should be significantly faster as inlined SVGs in pages do no change with every build and by this requiring a smaller subset of files to update/upload.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -69,7 +69,6 @@ jobs:
         run: npm run lint:scss
 
       - name: Build External
-        if: ${{ steps.filter.outputs._external == 'true' }}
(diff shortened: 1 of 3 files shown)


matthiasrohmer:
Updated the check workflow to always run build-external. It currently almost always does anyway, as almost all PRs start an 11ty build, but in the rare cases where we do not start a build the E2E tests can not finish as they rely on the external files to be built.

devnook (approved):

PR #4019, opened 2022-10-14
Misc deprecation updates

tunetheweb:

- Fixes deprecation and removal redirects that currently don't work (e.g. try https://developer.chrome.com/tags/removals/ to see)
- Updates Web-SQL post to use new tag (so it doesn't even need the redirect) - FYI @tomayac 
- Updates HTTP/2 Server Push notice with links

### DIFF
--- a/redirects.yaml
+++ b/redirects.yaml
@@ -469,11 +469,11 @@ redirects:
 
 # Tag Redirects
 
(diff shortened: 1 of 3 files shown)


tomayac (approved):
LGTM, thanks, Barry!

PR #4020, opened 2022-10-14
Adding chrometober banner

tropicadri:
Adding a banner to promote chrometober content. Copy created by Harleen


rachelandrew (changes requested):
Same comment as on web.dev, I'm not happy with the container queries working across all browsers wording, because they don't. That's a confusing message.

rachelandrew (changes requested):
Same comment as on web.dev, I'm not happy with the container queries working across all browsers wording, because they don't. That's a confusing message.

tropicadri:
Harleen updated the wording. Thank you for checking.

rachelandrew (approved):

PR #4036, opened 2022-10-17
Specify only one service_worker key is allowed.

xcollantes:
Fixes ambiguity of how many `service_worker` fields can be specified https://developer.chrome.com/docs/extensions/mv3/mv3-migration/#man-sw.  

Changes proposed in this pull request: Add description to specify only one `service_worker` field can be defined in the `background` field. 

-
-
-


AmySteam (commented):
@xcollantes Thanks for this contribution! It's important to highlight the fact that MV3 extensions only support one service worker. 💯 

@jpmedley MV3 supporting a single service worker is technically correct. Can you perform an editorial review to make this aside more concise and explicit? Other than that, LGMT 👍

jpmedley (changes requested):

PR #4059, opened 2022-10-19
Events  - Content updates

johnrich85:
Adds some new events and updates others.


rachelandrew (changes requested):
Looking at the preview, there seems to be a duplicate event for the PerfNow pre-event, does this PR add a duplicate?

devnook:
@rachelandrew PTAL

johnrich85:
Ah yeah, good spot. That was supposed to be an update rather than an addition. I've deleted the duplicate and updated the original instead

johnrich85:
Apologies @rachelandrew . That should be sorted now.

rachelandrew (approved):

PR #4062, opened 2022-10-19
Remove broken link in the Shared Storage landing page

kevinkiklee:
This PR removes the link that is no longer valid

### DIFF
--- a/site/en/docs/privacy-sandbox/shared-storage/index.md
+++ b/site/en/docs/privacy-sandbox/shared-storage/index.md
@@ -92,7 +92,7 @@ storage worklet.
 
 The shared storage data can be used for:
 
-*  [**URL selection**](/docs/privacy-sandbox/shared-storage/url-selection): 
+*  **URL selection**: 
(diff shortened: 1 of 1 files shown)


site/en/docs/privacy-sandbox/shared-storage/index.md
```diff
@@ -92,7 +92,7 @@ storage worklet.
 
 The shared storage data can be used for:
 
-*  [**URL selection**](/docs/privacy-sandbox/shared-storage/url-selection): 
+*  **URL selection**:
```
heyawhite, line 95: Should this be titled frequency control and directed there?

heyawhite (approved):

site/en/docs/privacy-sandbox/shared-storage/index.md
kevinkiklee: This section is talking about the gate rather than the use case, so I think it's okay to leave it as "URL selection".

PR #4112, opened 2022-10-26
Region Capture is no longer only for self-capture tabs

beaufortfrancois:
This PR makes it clear the Region Capture API is no longer only for self-capture tabs, but Chrome only supports self-capture tabs. 

@eladalon1983 PTAL

### DIFF
--- a/site/en/docs/web-platform/region-capture/index.md
+++ b/site/en/docs/web-platform/region-capture/index.md
@@ -7,14 +7,14 @@ authors:
   - beaufortfrancois
   - eladalon
 date: 2022-06-21
-updated: 2022-08-04
+updated: 2022-10-26
 hero: image/vvhSqZboQoZZN9wBvoXq72wzGAf1/05M2IjnWmaeFbIDWaznm.jpeg
 alt: A woman holding her hands up and making a frame with her fingers. Above her the sky is blue with a few clouds.
 tags:
   - chrome-104
 ---
 
-The web platform already allows a web app to capture a video track of the current tab. It now ships with Region Capture, a mechanism for cropping these video tracks. The web app designates a portion of the current tab as its area of interest, and the browser crops all pixels outside that area.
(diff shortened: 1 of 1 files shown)


beaufortfrancois:
@eladalon1983 gentle ping

site/en/docs/web-platform/region-capture/index.md
```diff
@@ -111,7 +111,7 @@ If unsuccessful, the Promise will be rejected. This will happen if:
 The `cropTo()` method is exposed on any tab-capture video track, and not just for self-capture. It is therefore advisable to check if the user selected the current tab, before attempting to crop the track. This can be accomplished using [Capture Handle]. It is also possible to ask the browser to nudge the user towards self-capture using [`preferCurrentTab`].
```
eladalon1983, line 111: See this:
> - The current track is not a self-capture video track; see below.

And this:
> The `cropTo()` method is exposed on any tab-capture video track, and not just for self-capture.

And this:
>It is also possible to ask the browser to nudge the user towards self-capture using [`preferCurrentTab`].

I suggest scanning the doc for "self" and "own".

beaufortfrancois: I've tried to update the article so that when talking about the Region Capture API in general, I don't use self-capture and when talking about Chrome specifically, I add mention that it only supports self-capture video tracks.

eladalon1983: Such a nuance would have eluded me if I were the reader. I think it would be wiser to [KISS](https://en.wikipedia.org/wiki/KISS_principle). How about we wait until we implement this, and then we can overhaul the doc to call out and demonstrate other-tab-cropping, possibly with a useful demo too?

beaufortfrancois:
I'm thinking of closing this PR and re-open a new one when Region Capture API changes actually occur in Chrome. @eladalon1983 Is that okay with you?

eladalon1983:
I think we'll eventually do this, so no need to remove the PR. Although I guess it can be reopened later.
What changes that are actually happening in Chrome are you referring to, btw?

PR #4155, opened 2022-10-30
CWS documentation: Fix dead link to /cws-dashboard-enterprise/

andyfischer:
The url https://developer.chrome.com/docs/webstore/cws-dashboard-enterprise/ gives a 404 error.

Replacing that same link with this url https://developer.chrome.com/docs/webstore/cws-enterprise/

Also replaced `[enterprise]` with `[cws-enterprise]` since it's the same link now.

These fixes are for the page: https://developer.chrome.com/docs/webstore/publish/


AmySteam (approved):
@andyfischer Thanks for this PR 👍 🙂

@jpmedley LGTM, we can merge it.

jpmedley (approved):

PR #4171, opened 2022-11-01
Add a note about curl command for Windows-based systems

tunetheweb:
Fixes https://github.com/GoogleChrome/web.dev/issues/6696

- Add a note about `curl` command and how it might have to be altered for other systems


malchata (approved):
Been a while since I used Windows, but this seems to align with what I remember about cmd.exe. LGTM.

tunetheweb:
Yeah I tested it on my aging Windows machine.

PR #4188, opened 2022-11-02
Arch patches

jpmedley:
This PR reworks how asynchronous methods are documented. Specifically:

* Puts promises first since they're the more modern pattern.
* Adds information about feature detecting promises.
* Removes information about using JavaScript methods because it's out of scope.

### DIFF
--- a/site/en/docs/extensions/mv3/architecture-overview/index.md
+++ b/site/en/docs/extensions/mv3/architecture-overview/index.md
@@ -29,7 +29,7 @@ An extension's architecture will depend on its functionality, but all extensions
 ### The manifest {: #manifest }
 
 The manifest file, titled `manifest.json`, gives the browser information about the extension, such
-as the most important files and the capabilities the extension might use. An example is shown below. For information and descriptions of what the manifest can contain, see [Mainfest file format][docs-manifest].
+as the most important files and the capabilities the extension might use. 
 
 ```json
 {
--- a/site/en/docs/extensions/mv3/architecture-overview/index.md
+++ b/site/en/docs/extensions/mv3/architecture-overview/index.md
@@ -241,7 +241,7 @@ Most Chrome API methods are asynchronous; they return immediately without waitin
 * Use the returned promise.
 * Pass a callback function into the method.
 
-Note that these choiced are mutually exclusive. If you pass a callback to a method, no promise
+Note that these choices are mutually exclusive. If you pass a callback to a method, no promise
 will be returned. If you use the returned promise, do not pass a callback.
 
 Generally, you should prefer promises to callbacks. Not all methods in extensions APIs support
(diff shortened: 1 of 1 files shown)


site/en/docs/extensions/mv3/architecture-overview/index.md
```diff
@@ -234,65 +234,33 @@ using [chrome.tabs.create()][api-create-tab] instead.
...
 
-A method is asynchronous when the callback parameter is available in its signature.
+* Use the returned promise.
+* Pass a callback function into the method.
 
-```js
-// Signature for an asynchronous method
-chrome.tabs.query(object queryInfo, function callback)
-```
+Note that these choiced are mutually exclusive. If you pass a callback to a method, no promise
```
AmySteam, line 244:
```suggestion
Note that these choices are mutually exclusive. If you pass a callback to a method, no promise
```

site/en/docs/extensions/mv3/architecture-overview/index.md
```diff
@@ -234,65 +234,33 @@ using [chrome.tabs.create()][api-create-tab] instead.
...
+will be returned. If you use the returned promise, do not pass a callback.
 
-If the extension needed to navigate the user's currently selected tab to a new URL, it would need to
-get the current tab's ID and then update that tab's address to the new URL.
+Generally, you should prefer promises to callbacks. Not all methods in extensions APIs support
+promises, but newer methods do. You can verify whether a method supports promises by checking
+its API reference page. If you need to support both promises and callbacks for the same
+function (because your users have older browsers), you can test whether the method returns a
+promise using `typeof` and
+[optional chaning](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Operators/Optional_chaining).
```
AmySteam, line 252:
```suggestion
[optional chaining](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Operators/Optional_chaining).
```

site/en/docs/extensions/mv3/architecture-overview/index.md
AmySteam: Should this be an aside? 🤔 I think this is important not to miss, but you can make this call.

AmySteam (changes requested):
@jpmedley Great job on the PR 👍 . Just found a couple of details to fix and one observation.

site/en/docs/extensions/mv3/architecture-overview/index.md
jpmedley: I debated that myself. I think it's unlikely that anyone will actually try to use both methods, so I see no reason to draw attention to it.

jpmedley:
PTAL

AmySteam (approved):

PR #4189, opened 2022-11-03
Change default value for selfBrowserSurface

beaufortfrancois:
This change will make it compliant with Chrome behavior.

@eladalon1983 FYI

### DIFF
--- a/site/en/docs/web-platform/screen-sharing-controls/index.md
+++ b/site/en/docs/web-platform/screen-sharing-controls/index.md
@@ -6,7 +6,7 @@ description: >
 authors:
   - beaufortfrancois
   - eladalon
-date: 2022-09-30
+date: 2022-11-03
 hero: image/vvhSqZboQoZZN9wBvoXq72wzGAf1/5VpvNzrvEM3qSxasqP1j.jpeg
 alt: toddler holding her lips photo
 tags:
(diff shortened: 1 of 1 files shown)


beaufortfrancois:
With https://chromium-review.googlesource.com/c/chromium/src/+/4008501 now merged, @eladalon1983 do you want to wait for Chrome 108 to reach stable channel or do you want to land this PR now?

eladalon1983:
The question is trickier than it might initially appear, because it's a gradual rollout. That means that different developers will be exposed to the change at different times, and it might also get reverted at some moment. Wdyt, given that additional information?

beaufortfrancois:
I'd rather wait for the gradual rollout to reach 100% before making this change then.

beaufortfrancois:
@eladalon1983 What's the status?

eladalon1983:
https://critique.corp.google.com/cl/490713846

eladalon1983 (commented):
At the moment, this is at 50%, which means 'include' and 'exclude' are equally correct, and soon this will be at 100%. So merging right away makes sense to me. LGTM.

beaufortfrancois:
@rachelandrew PTAL

rachelandrew (approved):

Malikjavedibrahim:
Where am I and what is written, I don't understand, so please help

PR #4076, opened 2022-10-20
Added note about NodeJS support for early hints

richardeschloss:
Changes proposed in this pull request:

- Added note about NodeJS support for early hints. It appears as though support landed in NodeJS 18.x on [Aug 17, 2022](https://github.com/nodejs/node/pull/44180)

PR #4191, opened 2022-11-03
Update Early Hints blog post to mention current node.js support

voxpelli:
Created independently of #4076 and submitted individually as the CLA for that one is not signed.

This adds a mention of the two API:s that now supports sending Early Hints + the version number where it was introduced and a link to the release notes for that version number.


AaronForinton (approved):
Approving this. Thank you @voxpelli for taking the time to adjust this.

PR #4196, opened 2022-11-03
Events - partial render

johnrich85:

- Renders the first 5 events from each list, as opposed to rendering all of them to reduce DOM size on load.
- Conditionally adds a 'load more' button to each list, allowing the user to fetch additional events if necessary.
- Fixes: #4306


johnrich85:
I've left this one as a draft for now @devnook as I still need to add tests for the load-more script. Everything else is done though so feel free to review if you want to get ahead of it.

site/_js/misc/render-event-card.js
```diff
@@ -0,0 +1,272 @@
...
+ *
+ * @param {object} event
+ * @param {object} icons
+ * @returns string
+ */
+function RenderEventCard(event, icons) {
+  const buttonLabel = event.isPastEvent ? 'See details' : "See who's joining";
+
+  return html`
+    <enhanced-event-card>
```
devnook, line 31: This is used on backend code, right? in _js/ folder, only frontend js files should live.

If you want to use this as aboth backend lib and frontend lib, I worry that:
1. we will run into issues using require() modules in frontend code
2. you use a LOT of js, with additional imports, to achieve fairly straightforward declarative html output.

How about rendering the events on backend, and sending down html partials instead of json? or embed output html as json field? 

Or alternatively, making it a web component, and always render on the client side?

devnook (changes requested):
Hi John, I'm a bit concerned about the RenderEventCard logic moved to frontend, mostly because performance and mixing module types point of view. Happy to sync on chat if needed.

site/_js/misc/load-more.js
```diff
@@ -0,0 +1,92 @@
...
+
+    if (loading) return;
+
+    toggleLoading();
+
+    button.setAttribute('disabled', '');
+
+    const items = await fetchItems(currentOffset, take);
+
+    items.forEach(html => container.insertAdjacentHTML('beforeend', html));
```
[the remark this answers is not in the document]
johnrich85: Went a slightly different way with this and instead compiled the collection down to a single string and then inserted that. As I understand it that'll achieve the same thing without some of the overheads. Correct me if am wrong!

site/_js/utils/events.js
```diff
@@ -0,0 +1,14 @@
+const startOfDay = new Date();
+startOfDay.setHours(0, 0, 0, 0);
+
+const isPastEvent = event => {
+  return new Date(event.date).getTime() < startOfDay.getTime();
+};
+
+const sortDescByTime = (a, b) =>
+  new Date(b.date).getTime() - new Date(a.date).getTime();
```
matthiasrohmer, line 9: EleventyCollectionItem.date already is a `Date` AFAIK, so there wouldn't be a need to recast it as a date in those sorting functions. Or are you sorting anything else with those?

site/events.njk
```diff
@@ -2,4 +2,16 @@
...
+      {% include 'partials/event-card.njk' %}
+    {% endset %}
+    {
+      "id": {{event.id|dump|safe}},
+      "html": {{ html|minifyHtml|dump|safe}},
+      "date": {{ event.date|dump|safe }},
+      "isPastEvent": {{ event.isPastEvent|dump|safe }}
+    } {% if loop.last === false %},{% endif %}
+  {% endfor %}
+]
```
matthiasrohmer, line 17: This all looks sus to me 😅  I know one-off filters are also not ideal, but putting this logic into a `prerenderEvents` filter or something like this appears cleaner to me. @devnook, WDYT?

site/_js/utils/events.js
johnrich85: So this is used again on the frontend after the events have been serialized to JSON, at which point they are strings. It's not strictly necessary at this point as they'll be in the right order in the dump anyway. However, there is an outstanding requirement to dynamically move elapsed events over to the past events tab at which point this will be necessary.

site/events.njk
johnrich85: Is it possible to render templates/partials in filters? I wasn't able to find an API allowing that, or were you thinking something else?

site/_js/utils/events.js
matthiasrohmer: Ah I see. Now I also remember having stumbled over this before... I personally not a fan of mixing "backend" (11ty) code with frontend code but also can not say if that is already the case anywhere else in the codebase. @devnook, WDYT?

site/events.njk
matthiasrohmer: Something like for the partial shortcode might work. But I think one might also be able to extract the nunjucks engine from a shortcode context to doe stuff with it, but need to dig around. Let's wait for Ewa's opinion on this 🙂

site/_transforms/minify-html.js
```diff
@@ -38,4 +40,4 @@ const minifyHtml = (content, outputPath) => {
   return content;
 };
 
-module.exports = {minifyHtml};
+module.exports = {minifyHtml, options};
```
devnook, line 43: options is definitely not a descriptive name for a module export

site/events.njk
devnook: Best would be to prepare html in the events.js file, where the rest of json is generated. BUT njk engine is not available there yet. So we have to render it in the events.njk template. Wheter we do it with a filter, shortode or include, is't all the same I think. Storing it in variable is as good as any other way, just please double check if it's not escaping things twice.

site/_js/utils/events.js
matthiasrohmer: Yep, I like that!

PR #4208, opened 2022-11-04
[doc-ext] Remove no longer relevant aside in service worker article

AmySteam:
This PR removes an aside that warned developers to place the service worker file in the root. This was true during the first releases of MV3, but it is no longer the case.

**Staged link**
[Service worker](https://deploy-preview-4208--developer-chrome-com.netlify.app/docs/extensions/mv3/service_workers/)


jpmedley:
There's nothing here to fix. Forgive me for asking how sure you are of this. The text you're removing is actually what I would expect knowing how web service workers function.

AmySteam:
@jpmedley confirmed it with Simeon and tested it locally 👍. The service worker file can be located inside a subdirectory.

jpmedley (approved):

PR #4232, opened 2022-11-08
Adds event tags collection

johnrich85:

- Adds a new collection that will eventually be used to populate the event filters (as seen in the attached)

image

To preview the collection: {{collections.eventTags | dump | safe }}


johnrich85:
This one generates a new collection behind the scenes but it's not actually used anywhere yet @devnook . I thought id separate it out anyway in this instance as it's the only 'block' that can be safely separated from the next PR (adding filters to ui and hooking up), which i don't want to get too complex.

devnook (approved):

PR #4235, opened 2022-11-08
Switch to improved webdev-infra CSS transform

matthiasrohmer:
This replaces the current css-purify transformer with the one implemented in https://github.com/GoogleChrome/webdev-infra/pull/36 while also using the newly introduced InlineCss shortcode, which also allows pages not using main.css to inline their CSS.

The main performance improvements come from by reading the JS from the disk once instead of letting PurgeCSS read it again for every page, so over 15k saved fs calls and by optimizing the CSS once before purging as this lowers the amount of data PurgeCSS needs to handle.


matthiasrohmer:
Closed in favor of https://github.com/GoogleChrome/developer.chrome.com/pull/4316.

PR #4237, opened 2022-11-08
Add information about trusted testers to Web Store help.

jpmedley:
**NOTE:** There was no obvious place to tack on this information. I required minor refactoring of the text. The alternative was undermining the internal logic of the help, which is not good for users.


AmySteam:
Here's another place where trusted testers is mentioned:
https://developer.chrome.com/docs/webstore/cws-dashboard-distribution/#private-visibility-trusted-testers

PR #4269, opened 2022-11-11
URL.createObjectURL is not supported in service workers

jsalvata:

- Mention that service workers don't support `URL.createObjectURL`


jpmedley:
@dotproto This is getting a little long in the tooth. It's a single line. Can you verify this real quick.

dotproto (approved):
LGTM. Thanks for the suggestion!

PR #4275, opened 2022-11-11
Improvements to messaging article.

jpmedley:
@dotproto, I realized this morning that I needed to understand messaging better. If I'm reading something, the cost of improving it is marginal. I made some V3-specific updates to this article. You can find them at lines 27, 33, 43, and 274. Please ignore the other stuff. I'll ask @AmySteam to do a proper copy edit before I merge.

### DIFF
--- a/site/en/docs/extensions/mv3/messaging/index.md
+++ b/site/en/docs/extensions/mv3/messaging/index.md
@@ -2,7 +2,7 @@
 layout: "layouts/doc-post.njk"
 title: "Message passing"
 date: 2012-09-18
-updated: 2021-06-01
+updated: 2022-11-11
 description: How to pass messages between extensions and content scripts.
 ---
 
@@ -14,36 +14,34 @@ order to display a page action icon for that page.
 Communication between extensions and their content scripts works by using message passing. Either
 side can listen for messages sent from the other end, and respond on the same channel. A message can
 contain any valid JSON object (null, boolean, number, string, array, or object). There is a simple
-API for [one-time requests][1] and a more complex API that allows you to have [long-lived
-connections][2] for exchanging multiple messages with a shared context. It is also possible to send
-a message to another extension if you know its ID, which is covered in the [cross-extension
+API for [one-time requests][1] and a more complex API for [long-lived
+connections][2] to exchange multiple messages within shared context. It is also possible to send
+a message to another extension if you know its ID. That is covered in the [cross-extension
 messages][3] section.
 
 ## Simple one-time requests {: #simple }
 
 If you only need to send a single message to another part of your extension (and optionally get a
-response back), you should use the simplified [runtime.sendMessage][4] or [tabs.sendMessage][5].
-This lets you send a one-time JSON-serializable message from a content script to extension, or vice
-versa, respectively. An optional callback parameter allows you to handle the response from the other
-side, if there is one.
+response back), use the simplified [runtime.sendMessage()][4] or [tabs.sendMessage()][5].
+This lets you send a one-time JSON-serializable message from a content script to the extension, or vice
+versa. To handle the response, use the returned Promise. (For backward compatibility, you can still
+pass a callback as the last argument.)
 
 Sending a request from a content script looks like this:
 
 ```js
-chrome.runtime.sendMessage({greeting: "hello"}, function(response) {
-  console.log(response.farewell);
-});
+await response = chrome.runtime.sendMessage({greeting: "hello"});
+console.log(response.farewell);
 ```
 
-Sending a request from the extension to a content script looks very similar, except that you need to
+Sending a request from the extension to a content script is similar, except that you need to
 specify which tab to send it to. This example demonstrates sending a message to the content script
 in the selected tab.
 
 ```js
 chrome.tabs.query({active: true, currentWindow: true}, function(tabs) {
-  chrome.tabs.sendMessage(tabs[0].id, {greeting: "hello"}, function(response) {
-    console.log(response.farewell);
-  });
+  await response = chrome.tabs.sendMessage(tabs[0].id, {greeting: "hello"});
+  console.log(response.farewell);
 });
 ```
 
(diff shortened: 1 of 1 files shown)


site/en/docs/extensions/mv3/messaging/index.md
```diff
@@ -14,36 +14,34 @@ order to display a page action icon for that page.
...
+console.log(response.farewell);
 ```
 
-Sending a request from the extension to a content script looks very similar, except that you need to
+Sending a request from the extension to a content script is similar, except that you need to
 specify which tab to send it to. This example demonstrates sending a message to the content script
 in the selected tab.
 
 ```js
 chrome.tabs.query({active: true, currentWindow: true}, function(tabs) {
```
dotproto, line 42:
```suggestion
chrome.tabs.query({active: true, lastFocusedWindow: true}, function(tabs) {
```

Let's update this to match the query parameters used in the [Tabs API example](https://developer.chrome.com/docs/extensions/reference/tabs/#get-the-current-tab).

site/en/docs/extensions/mv3/messaging/index.md
```diff
@@ -344,3 +350,4 @@ directory.
 [40]: https://github.com/GoogleChrome/chrome-extensions-samples/tree/master/mv2-archive/api/messaging
 [41]: /docs/apps/nativeMessaging/#examples
 [42]: /docs/extensions/mv3/samples
+[43]; /docs/extensions/mv3/manifest/externally_connectable
```
dotproto, line 353: Typo

```suggestion
[43]: /docs/extensions/mv3/manifest/externally_connectable
```

site/en/docs/extensions/mv3/messaging/index.md
```diff
@@ -265,7 +271,7 @@ scripts.
 When receiving a message from a content script or another extension, your scripts should be careful
 not to fall victim to [cross-site scripting][39]. This advice applies to scripts running inside the
 extension background page as well as to content scripts running inside other web origins.
-Specifically, avoid using dangerous APIs such as the ones below:
+Specifically, avoid using dangerous APIs, specificaslly `chrome.scripting.executeScript()`, `eval()`, and `new Function()`. For example:
```
dotproto, line 274: This section is a bit odd now because of CSP changes in Manifest V3. Ian is working on #3837 to update our CSP docs, but the short version is that Chrome has a minimum CSP for extension contexts that does not allow developers to call `eval()` or `new Function()`. For reference, the minimum CSP is `script-src 'self' 'wasm-unsafe-eval'; object-src 'self';`.

It may be best to use a slightly different approach for this section. Since cross-site scripting isn't as much a concern, we could lean into "cross-site attacks," where a compromised render could use the message channel to send the extension messages in order to extract data or make the extension do something dangerous.

Cross-site scripting is still a concern for content scripts. This is another method by which a content script could potentially be hijacked into sending dangerous messages back to the extension.

dotproto (changes requested):

site/en/docs/extensions/mv3/messaging/index.md
AmySteam, line 274:
```suggestion
Specifically, avoid using dangerous APIs, specifically `chrome.scripting.executeScript()`, `eval()`, and `new Function()`. For example:
```

AmySteam (changes requested):
@jpmedley Good improvements.

Since you are making this article MV3 compatible, can you change the mention of background pages to extension service worker?

> For example, an RSS reader extension might use content scripts to detect the presence of an RSS feed on a page, then notify the background page in order to display a page action icon for that page.

This sentence mentions a background page and a page action. Both are not supported in MV3.

site/en/docs/extensions/mv3/messaging/index.md
jpmedley: I'm going to remove the new text so that we can merge this. I'll circle back when I have more information.

AmySteam (changes requested):
LGTM 👍 I just have some small change requests. Thanks!

AmySteam (approved):

jpmedley:
@AmySteam Thank you for the review. Are there any fact that you think @dotproto needs to verify before we merge this?

AmySteam:
@jpmedley Nope, I think this can be merged as is 👍

site/en/docs/extensions/mv3/messaging/index.md
```diff
@@ -14,36 +14,34 @@ order to display a page action icon for that page.
...
 specify which tab to send it to. This example demonstrates sending a message to the content script
 in the selected tab.
 
 ```js
-chrome.tabs.query({active: true, currentWindow: true}, function(tabs) {
-  chrome.tabs.sendMessage(tabs[0].id, {greeting: "hello"}, function(response) {
-    console.log(response.farewell);
-  });
+chrome.tabs.query({active: true, lastFocusedWindow: true}, function(tabs) {
+  await response = chrome.tabs.sendMessage(tabs[0].id, {greeting: "hello"});
```
tophf, line 43: This is invalid JS syntax.

site/en/docs/extensions/mv3/messaging/index.md
```diff
@@ -14,36 +14,34 @@ order to display a page action icon for that page.
...
+versa. To handle the response, use the returned Promise. (For backward compatibility, you can still
+pass a callback as the last argument.)
 
 Sending a request from a content script looks like this:
 
 ```js
-chrome.runtime.sendMessage({greeting: "hello"}, function(response) {
-  console.log(response.farewell);
-});
+await response = chrome.runtime.sendMessage({greeting: "hello"});
```
tophf, line 33: This is invalid JS syntax.

tophf:
Fixing the snippets is a little tricky because you need to show a complete `async` function, otherwise people who only learn JS will waste time (probably) a lot.

The first snippet:

```js
(async () => {
  const response = await chrome.runtime.sendMessage({greeting: "hello"});
  // do something with response here, not outside the function
  console.log(response);
})();
```

The second snippet should be rewritten:
```js
(async () => {
  const [tab] = await chrome.tabs.query({active: true, lastFocusedWindow: true});
  const response = await chrome.tabs.sendMessage(tabs[0].id, {greeting: "hello"});
  // do something with response here, not outside the function
  console.log(response);
})();
```

Generally you should convert all `function` declarations to `await` as shown above or to arrow functions - it is important in modern frameworks that need `this` to be bound to the app instance also inside the API callback.

Also make sure to at least verify each JS snippet you modify or add in devtools console, especially if JS is not your forte.

site/en/docs/extensions/mv3/messaging/index.md
jpmedley: Thank you for your feedback. We had an off day there.

See #4518

jpmedley:
We know that. I had an off day.

Issue #3595, opened 2022-08-30
Use clearer terminology to describe how extension service workers behave


dotproto:
**Describe the bug**
Documentation that describes how extension service workers behave 

**To Reproduce**
1. Go to https://developer.chrome.com/docs/extensions/mv3/service_workers/
2. Read through the documentation and look for references to how service workers are started and stopped.

**Expected behavior**
The documentation should have clear, consistent terminology that describes extension service worker behavior.

**Observed behavior**
The linked document uses several different terms to describe how extension service workers behave. 

* start
  * react
  * react to listeners
* run
  * keeps running
* stop
  * stay dormant
  * lie dormant
  * unload
  * suspend

This inconsistency likely also applies to other documentation in this corpus. 

**Additional context**
https://groups.google.com/a/chromium.org/g/chromium-extensions/c/y5hxPcavRfU/m/CF7EffJFAwAJ

PR #4312, opened 2022-11-15
Use clearer terminology in service worker docs

jpmedley:
Addresses #3595. There were few things to correct. As long as I was reading, I tightened a few other things.

[Staging site](https://deploy-preview-4312--developer-chrome-com.netlify.app/docs/extensions/mv3/service_workers/)

### DIFF
--- a/site/en/docs/extensions/mv3/service_workers/index.md
+++ b/site/en/docs/extensions/mv3/service_workers/index.md
@@ -120,7 +120,7 @@ Extensions can remove listeners from their background scripts by calling `remove
 
 Use APIs that support [event filters][8] to restrict listeners to specific use cases. If an
 extension is listening for the [`tabs.onUpdated`][9] event, try using the
-[`webNavigation.onCompleted`][10] event with filters instead, as the tabs API does not support
+[`webNavigation.onCompleted`][10] event with filters instead, as the `onUpdated` event does not support
 filters.
 
 ```js
--- a/site/en/docs/extensions/mv3/service_workers/index.md
+++ b/site/en/docs/extensions/mv3/service_workers/index.md
@@ -8,18 +8,18 @@ description: How to respond to browser triggers (events) from a Chrome Extension
 
 Extensions are event-based programs used to modify or enhance the Chrome browsing experience. Events
 are browser triggers, such as navigating to a new page, removing a bookmark, or closing a tab.
-Extensions monitor these events using scripts in their [extension service
-worker][doc-sw-migration], which then react with specified instructions.
+Extensions monitor these events using scripts in an [extension service
+worker][doc-sw-migration] (previously called a background script), which then executes specified instructions.
 
-An extension service worker (previously called a background script) is loaded when it is needed,
-and unloaded when it goes idle. Some examples include:
+An extension service worker is loaded when it is needed, and unloaded when it goes dormant.
+Examples include:
 
 - The extension is first installed or updated to a new version.
-- The extension service workers was listening for an event, and the event is dispatched.
+- An extension event is dispatched.
 - A content script or other extension [sends a message.][1]
 - Another view in the extension, such as a popup, calls [`runtime.getBackgroundPage()`][2].
 
-Once it has been loaded, an extension service worker generally keeps running as long as it is
+Once it has been loaded, an extension service worker generally runs as long as it is
 performing an action such as calling a Chrome API or issuing a network request.
 
 {% Aside %}
@@ -29,7 +29,7 @@ loaded.
 
 {% endAside %}
 
-Extension service workers stay dormant until an event they are listening for fires. At which point, they execute the appropriate event litener, then unlod.
+Extension service workers stay dormant until an event they are listening for fires. At which point, they execute the appropriate event litener, then unload.
 
 ## Register the service worker {: #manifest }
 
@@ -79,9 +79,10 @@ Structure extension service workers around events the extension depends on. Defi
 events allows background scripts to lie dormant until those events are fired and prevents the
 extension from missing important triggers.
 
-Listeners must be registered at the start of your extension service worker. Notice that the listener for creating bookmarks is at the top level of the script.
+Listeners must be registered at the start of your extension service worker. Notice that this
+example the listener for creating bookmarks is at the top level of the script.
 
-{% Compare 'worse' %}
+{% Compare 'better' %}
 ```js/8-10
 chrome.runtime.onInstalled.addListener(() => {
   chrome.contextMenus.create({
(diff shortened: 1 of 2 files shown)


site/en/docs/extensions/mv3/service_workers/index.md
```diff
@@ -42,7 +41,7 @@ field. This field uses the `"service_worker"` key, which specifies a single Java
   "name": "Awesome Test Extension",
   ...
   "background": {
-    "service_worker": "background.js"
+    "service_worker": "service-worker.js"
```
dotproto, line 44: At the moment the canonical example filename for the main background entry point is `bacgkround.js`. This practice was established at some point in he past and carried over to Manifest V3. Do you want to change this across all our docs and samples?

site/en/docs/extensions/mv3/service_workers/index.md
```diff
@@ -76,13 +75,15 @@ chrome.runtime.onInstalled.addListener(() => {
 
 ## Set up listeners {: #listeners }
 
-Structure background scripts around events the extension depends on. Defining functionally relevant
+Structure extension service workers around events the extension depends on. Defining functionally relevant
 events allows background scripts to lie dormant until those events are fired and prevents the
```
dotproto, line 79:
```suggestion
events allows the service wroker to lie dormant until those events are fired and prevents the
```

site/en/docs/extensions/mv3/service_workers/index.md
```diff
@@ -96,9 +97,12 @@ chrome.bookmarks.onCreated.addListener(() => {
   // do something
 });
 ```
+{% endCompare %}
 
-Do not register listeners asynchronously, as they will not be properly triggered.
+If you put the same listener somewhere else, for examle, inside the `installed` listener, it may
```
dotproto, line 102: Typo. 

```suggestion
If you put the same listener somewhere else, for example, inside the `installed` listener, it may
```

site/en/docs/extensions/mv3/service_workers/index.md
```diff
@@ -96,9 +97,12 @@ chrome.bookmarks.onCreated.addListener(() => {
...
+not be called when it is needed because it may not even be initialized yet.
```
dotproto, line 103: Typo + rewrite for clarity. Might be good to discuss this one as it's a bit weird to wrap your head around.

```suggestion
If you put the same listener somewhere else, for example, inside the `installed` listener, 
the event will be dropped because the nested listener will not be registered when the browser 
attempts to dispatch it.
```

site/en/docs/extensions/mv3/service_workers/index.md
```diff
@@ -108,21 +112,14 @@ chrome.runtime.onInstalled.addListener(() => {
...
-```
+Extensions can remove listeners from their background scripts by calling `removeListener()`. 
 
 ## Filter events {: #filters }
 
-Use APIs that support [event filters][8] to restrict listeners to the cases the extension cares
-about. If an extension is listening for the [`tabs.onUpdated`][9] event, try using the
+Use APIs that support [event filters][8] to restrict listeners to specific use cases. If an
+extension is listening for the [`tabs.onUpdated`][9] event, try using the
 [`webNavigation.onCompleted`][10] event with filters instead, as the tabs API does not support
```
dotproto, line 123: Avoid overuse of "API".

```suggestion
[`webNavigation.onCompleted`][10] event with filters instead, as the `onUpdated` event does not support
```

dotproto (changes requested):

site/en/docs/extensions/mv3/service_workers/index.md
jpmedley: I actually do. I've observed conversations on StackOverflow where one party was talking about V2 and the other was talking about V3, but they didn't know it because they were both using the term 'background.js'.

site/en/docs/extensions/mv3/service_workers/index.md
jpmedley: I like your rewrite, but for some reason, I can't commit it.

site/en/docs/extensions/mv3/service_workers/index.md
jpmedley: Have you given this any thought?

site/en/docs/extensions/mv3/service_workers/index.md
dotproto: Looks like the suggestion is outdated. I think you'll need to manually commit the change.

site/en/docs/extensions/mv3/service_workers/index.md
jpmedley: Will do.

jpmedley (commented):
PTAL

site/en/docs/extensions/mv3/service_workers/index.md
```diff
@@ -108,22 +113,15 @@ chrome.runtime.onInstalled.addListener(() => {
...
 
 ## Filter events {: #filters }
 
-Use APIs that support [event filters][8] to restrict listeners to the cases the extension cares
-about. If an extension is listening for the [`tabs.onUpdated`][9] event, try using the
-[`webNavigation.onCompleted`][10] event with filters instead, as the tabs API does not support
+Use APIs that support [event filters][8] to restrict listeners to specific use cases. If an
+extension is listening for the [`tabs.onUpdated`][9] event, try using the
+[`webNavigation.onCompleted`][10] event with filters instead, as the `onUpdated` event does not support
 filters.
```
dotproto, line 125: tabs.onUpdated` and `webNavigation.onCompleted` are aimed at different uses, so this suggestion felt weird to me. As is my way, I inserted more context on this scenario to make it read better to me. I defer to you on whether this is a good change.

```suggestion
Use APIs that support [event filters][8] to restrict listeners to specific use cases. If an
extension is listening for the [`tabs.onUpdated`][9] event to detect when a user navigates
to a specific website, try using the [`webNavigation.onCompleted`][10] event with filters
instead, as the `tabs.onUpdated` event does not support filters.
```

dotproto (changes requested):

site/en/docs/extensions/mv3/service_workers/index.md
```diff
@@ -96,9 +97,13 @@ chrome.bookmarks.onCreated.addListener(() => {
...
 
-Do not register listeners asynchronously, as they will not be properly triggered.
+If you register the same listener asynchronously (for example, inside a `runtime.onInstalled` listener), 
+the event will be dropped because the nested listener will not be registered when the browser 
+attempts to dispatch it.
 
+{% Compare 'worse' %}
 ```js
 chrome.runtime.onInstalled.addListener(() => {
   // ERROR! Events must be registered synchronously from the start of
```
dotproto, line 109: Ah! Almost missed this. We should update the terminology in this comment to reflect the change made to the second paragraph after the "Set up listeners" header.

site/en/docs/extensions/mv3/service_workers/index.md
```diff
@@ -76,13 +75,15 @@ chrome.runtime.onInstalled.addListener(() => {
...
+events allows the service wroker to lie dormant until those events are fired and prevents the
 extension from missing important triggers.
 
-Listeners must be registered synchronously from the start of the page.
+Listeners must be registered at the start of your extension service worker. Notice that in this
+example the listener for creating bookmarks is at the top level of the script.
 
-```js
+{% Compare 'better' %}
+```js/8-10
```
dotproto, line 86: Looks like we might have an off by one error here; I'm assuming that the comment wasn't supposed to be included in the highlight range and we should include the closing curly brace.

Screenshot 2023-01-05 at 6 24 00 PM


````suggestion
```js/9-11
````

PR #4316, opened 2022-11-16
Add new HTML minify and inline CSS transformers

matthiasrohmer:
Replaces #4235 and additionally adds the new HTML minifier based on @swc/html. For context see https://github.com/GoogleChrome/webdev-infra/pull/38.

This also removes the JS minifying pipeline as this is taken care of by @swc/html. By this DOM transformer also no longer needs to be async. Terser is a peerDependency by the rollup-plugin anyways and hence doesn't need to be a direct dependency of ours.

CSSO and purgeCss are now dependencies of webdev-infra.


matthiasrohmer:
John implemented a html-minify filter in https://github.com/GoogleChrome/developer.chrome.com/pull/4196. Will add this to https://github.com/GoogleChrome/webdev-infra/pull/38 before publishing and merging this.

matthiasrohmer:
I am mad. I moved pooling for the CSS transform to webdev-infra, again using async-transforms. I didn't gave it a second thought, thinking we'd be using the most recent version here and blindly installing that one for webdev-infra.

After I've done that, 11ty was never exiting after the build. Turns out, the most recent version of async-transforms has a bug that keeps the MessageChannel open, preventing using apps from shutdown 🫠

After much pointless debugging I switched to [piscina](https://github.com/piscinajs/piscina), which didn't help fixing the bug as the dependency version coming from webdev-infra already ended up in the package-lock.json, which I only found out after desperately diffing the only file I haven't already diffed.

image

matthiasrohmer:
Deploy times on Netlify.

Before: [Copied 721 files / Wrote 2084 files in 114.69 seconds (55.0ms each, v1.0.1)](https://app.netlify.com/sites/developer-chrome-com/deploys/637cc1710627a800093df30f)
After: [Copied 721 files / Wrote 2084 files in 87.38 seconds (41.9ms each, v1.0.1)](https://app.netlify.com/sites/developer-chrome-com/deploys/637cc1710627a800093df30f)

matthiasrohmer:
Okay now we're good. /100 and /meet-the-team also working as expected.

devnook (approved):

PR #4339, opened 2022-11-18
Add note about search engine override.

jpmedley:

- Fixes https://bugs.chromium.org/p/chromium/issues/detail?id=1363488
- Corrects a few formatting issues.


site/en/docs/extensions/mv3/settings_override/index.md
```diff
@@ -70,22 +73,20 @@ Values in the manifest can be customized in the following ways:
...
   first '/' in the URL.
 
 ## Reference {: #reference }
 
 An extension can override one or more of the following properties in the manifest:
 
-- **`homepage` (string)** - optional
-
-  New value for the homepage.
+**`homepage` (string)** - optional
```
dotproto, line 85:
```suggestion
`homepage` (string) - optional
```

Using `**bold**` here may not be necessary here due to the way our styling rules and how Markdown definition list processing works.

The paragraph before a line with a `: ` prefix is wrapped in a `<dt>` tag. These tags are styled with given a higher-than-normal `font-weight` of 600. `<strong>` tags have a font-weight of bold (700). While there _is_ a visible difference in the final rendering, it's minor.

**Rendering in Chrome with `<strong>` inside the `<dt>`**
screenshot-1

**Rendering  in Chrome with only a `<dt>`** 
screenshot-2

site/en/docs/extensions/mv3/settings_override/index.md
```diff
@@ -70,22 +73,20 @@ Values in the manifest can be customized in the following ways:
...
+: New value for the homepage.
 
-- **`search_provider` (object)** - optional
-
-  A search engine
+**`search_provider` (object)** - optional
```
dotproto, line 88:
```suggestion
`search_provider` (object) - optional
```

dotproto (changes requested):
I find the structure of the definitions on this page is a bit odd and hard to parse. At the moment I'm not sure how best to restructure this info, so I don't want to block the PR on those concerns.

There are a couple Markdown issues we should clean up, but once they're addressed this should be good to merge.

jpmedley:
@dotproto Thank you for the feedback on the reference section. I've restored the original layout, which shortens this PR considerably.

PTAL

dotproto (approved):
LGTM

PR #4345, opened 2022-11-19
Correction to examples in $() and $$()

### LINKED ISSUE #4344: Correction to article: Console Utilities API reference
https://developer.chrome.com/docs/devtools/console/utilities/

There are a couple of minor mistakes in this article:
1. The example text for $(selector, startNode) didn't match the screenshot and explanation is not particularly clear. The "currently selected node" would have been represented as $0, and "after" is used where "descendant" would be clearer and more correct.
2. Quoted text: `$$(selector) returns an array of elements that match the given CSS selector. This command is equivalent to calling document.querySelectorAll().` This isn't quite correct, as querySelectorAll() returns a NodeList, not an Array.

I've made proposed changes and will have a pull request imminently.

acropup:
Fixes #4344
CLA signed as Shane Burgess.

Addresses two minor mistakes in the Console Utilities API reference:

1. Example text for $(selector, startNode) didn't match the screenshot. The "currently selected node" would have been represented as $0.
2. Clarified that $$(selector) returns an array, whereas document.querySelectorAll() returns a NodeList.
_Note that an inline `<code>` block was used instead of backticks in order to maintain the hyperlink within._

### DIFF
--- a/site/en/docs/devtools/console/utilities/index.md
+++ b/site/en/docs/devtools/console/utilities/index.md
@@ -75,7 +75,7 @@ property:
 This function also supports a second parameter, `startNode`, that specifies an 'element' or Node from
 which to search for elements. The default value of this parameter is `document`.
 
-The following example returns a reference to the first element after the currently selected Node and
+The following example returns a reference to the first `img` element that is a descendant of `devsite-header-background`, and
 displays its `src` property:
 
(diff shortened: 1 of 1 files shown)


sofiayem (approved):

sofiayem:
@acropup Thank you for catching these!

PR #4351, opened 2022-11-20
Add scripting

stefanvd:
Fixes the documentation. The `"scripting"` is necessary to run the script (in Manifest V3). That was not mentioned in the Google Chrome documentation.
https://developer.chrome.com/docs/extensions/mv3/manifest/activeTab/


AmySteam (approved):
Hey @stefanvd! Thanks for this PR; good catch 👍

Issue #3955, opened 2022-10-10
Revamp Percy setup


matthiasrohmer:
Currently the site uses [Percy](https://percy.io/google-chrome/developer.chrome.com) to test PRs for visual changes and potential diffs they introduce. The current config only checks a small subset of hardcoded pages or rather page types:

https://github.com/GoogleChrome/developer.chrome.com/blob/6be350b9a675c68aa22ec7daf6403abf20163ed3/.percy.js#L1-L9

A Percy run on those URLs is performed for every commit that changes a file with those extensions:

https://github.com/GoogleChrome/developer.chrome.com/blob/6be350b9a675c68aa22ec7daf6403abf20163ed3/.github/workflows/percy.yml#L8-L12

This approach is not ideal, as changes affecting the UI might not be reflected in the tested URLs or unapproved snapshots might trigger false-positives. To improve the situation we want to do the following

- [ ] Identify all types of pages (templates) and create an elaborate example page for all of those templates, using all supported components for that type
- [ ] Create a new test case (snapshot) for each of those new example pages
- [ ] Create example pages using all variations of [every component](https://developer.chrome.com/docs/handbook/components/), both layout and prose components while also checking the [shortcode implementations](https://github.com/GoogleChrome/developer.chrome.com/tree/main/site/_shortcodes) to verify all features are documented
- [ ] Create snapshots for component example pages
- [ ] Improve when Percy tests are actually run. Ideally only run when visual changes are likely, instead of every commit

PR #4354, opened 2022-11-21
Percy revamp

johnrich85:
First steps towards #3955:

- Adds new 11ty instance to tests/visual
- Updates percy so that it uses the new instance
- Adds the first test page


.percy.js
```diff
@@ -1,11 +1,6 @@
 const SNAPSHOTS = [
-  '/index.html', // Home
-  '/docs/index.html', // Docs
-  '/docs/native-client/index.html', // Docs landing page
-  '/docs/extensions/mv3/index.html', // Docs page
-  '/docs/extensions/reference/action/index.html', // API reference page
-  '/docs/handbook/test-post/index.html', // Test blog post
-  '/docs/handbook/components/index.html', // Handbook components
+  '/en/page-types/author-individual/index.html',
+  '/en/test-post/index.html',
```
devnook, line 3: Would this make sense to move test-post to the page-types dir?

.eleventy.js
```diff
@@ -169,7 +169,7 @@ module.exports = eleventyConfig => {
     eleventyConfig.addTransform('inlineCss', (new InlineCssTransform()).configure({
       cssBasePath: path.join(__dirname, 'dist'),
       jsPaths: [
-        path.join(__dirname, 'dist/js/**/*.js'),
+        path.join(__dirname, 'dist/js/**/*.js').split(path.sep).join('/')
```
devnook, line 172: Please add a comment that this is wwindows related fix.

gulp-tasks/preTestVisuals.js
```diff
@@ -0,0 +1,32 @@
...
+  // eslint-disable-next-line
+  const {execa} = await import('execa');
+
+  rimraf.sync('tests/visual/fixtures/dist');
+  mkdirSync('tests/visual/fixtures/dist', {recursive: true});
+
+  await Promise.all([execa('gulp'), execa('npx', ['rollup', '-c'])]);
+  // Copy all of whats normally built into dist by the default tasks
+  // into the test dist directory as well
+  cp('dist', 'tests/visual/fixtures/dist', {recursive: true});
```
devnook, line 31: I wonder if it would make sense to just use files from dist instead of copying here.

.percy.js
johnrich85: Yeah that makes total sense. I've actually done so in the second [PR](https://github.com/GoogleChrome/developer.chrome.com/pull/4450/files) which will be ready to go pretty much straight away once this one is merged in.

devnook (approved):

PR #4356, opened 2022-11-21
Show TOC in sidebar while scrolling

### LINKED ISSUE #3097: Show TOC in sidebar while scrolling
While scrolling through a long page like product documentation, it's hard to jump between sections given that the table of contents (TOC) is only available at the top of the page.

The side bar is currently used for related pages for the same project, but beyond a certain scroll position the space is blank and unused. It might be a nice compromise to pin the page TOC in the side bar after the other content scrolls away.

Relatedly, the Home/Docs/Blog/Articles navigation icons and links could also be sticky so that they're visible regardless of scroll position.

The pinned TOC is similar to how docs were presented on developers.google.com, for example: https://developers.google.com/web/tools/chrome-user-experience-report/

A fixed position TOC button that is always present in the viewport. Clicking it expands into the full, scrollable TOC.

**Additional context**
image

image

johnrich85:
Fixes #3097

### DIFF
--- a/site/_includes/partials/navigation-tree.njk
+++ b/site/_includes/partials/navigation-tree.njk
@@ -14,4 +14,7 @@
     {% from 'macros/navigation-tree-sections.njk' import navigationTreeSections with context %}
     {{ navigationTreeSections() }}
   </div>
+
+ {% include 'partials/toc-under-tree.njk' %}
+
 </navigation-tree>
--- a/site/_includes/partials/toc-under-tree.njk
+++ b/site/_includes/partials/toc-under-tree.njk
@@ -1,0 +1,11 @@
+{#
+  This is used as the TOC on smaller screens (<= 992) that makes use of the white space
+  under the navigation tree when the user scrolls down, in response to:
+  https://github.com/GoogleChrome/developer.chrome.com/issues/3097
+#}
+<div class="toc-container toc-container-under-navigation-tree display-none md:display-block xl:display-none pad-right-400 {% if not tocContents %}visibility-hidden{% endif %}">
+  <toc-active class="scrollbar">
+    <div class="type--h6 color-secondary-text">{{ 'i18n.common.toc' | i18n(locale) }}</div>
+    <div class="toc__wrapper">{{ tocContents | safe }}</div>
+  </toc-active>
+</div>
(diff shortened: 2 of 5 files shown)


devnook:
Hi John, this PR fixes the navbar but not the sidebar, and #3097 asked primarily to make the sidebar sticky. Can you take a look?

johnrich85:
Cheers @devnook . Just to clarify it makes the newly introduced TOC sticky too (screen width < 992px & > mobile) - just tested on: https://deploy-preview-4356--developer-chrome-com.netlify.app/docs/crux/methodology/. Is it supposed to make the whole sidebar sticky instead?

site/_includes/partials/toc-under-tree.njk
```diff
@@ -0,0 +1,11 @@
+{#
+  This is used as the TOC on smaller screens (<= 992) that makes use of the white space
+  under the navigation tree when the user scrolls down, in response to:
+  https://github.com/GoogleChrome/developer.chrome.com/issues/3097
+#}
+<div class="toc-container toc-container-under-navigation-tree display-none md:display-block xl:display-none pad-right-400 {% if not tocContents %}visibility-hidden{% endif %}">
```
devnook, line 6: This is repeating the content of toc-side.njk

Can we reuse the same template, and modify display properties in the css class? 

Also, we generally use BEM notation for non-utility classes. So this class should be something like toc-container--sticky or so. However, since this class is used mostly for positioning of the element, it would be better to add a css rule in side-nav instead (https://github.com/GoogleChrome/developer.chrome.com/blob/main/site/_scss/blocks/_side-nav.scss), e.g. 

navigation-tree toc-container {...}

site/_includes/partials/navigation-tree.njk
```diff
@@ -14,4 +14,7 @@
     {% from 'macros/navigation-tree-sections.njk' import navigationTreeSections with context %}
     {{ navigationTreeSections() }}
   </div>
+
+ {% include 'partials/toc-under-tree.njk' %}
```
devnook, line 18: Please reuse toc-side.njk (see other comment)

devnook (changes requested):
Ah no, my bad - I realized that on long sidebar content toc will become inaccessible. So let's keep the solution as is function-wise.

PR #4364, opened 2022-11-22
Revamp check workflow

matthiasrohmer:
This consolidates the check workflow to have a more detailed overview of which checks pass or fail and why. This PR also integrates the lighthouse check into the default check workflow while reusing the Eleventy build that is performed to verify 11ty integrity. The same is done for Percy. This allows us to establish dependencies between the jobs and not running Eleventy three times as currently and having it fail three times separately in the worst case.

image

image


.github/workflows/check.yml
```diff
@@ -1,81 +1,229 @@
...
+          NODE_OPTIONS: --max_old_space_size=4096
+          ELEVENTY_IGNORE_HANDBOOK: true
+          ELEVENTY_IGNORE_LIGHTHOUSE: true
+          ELEVENTY_IGNORE_MULTIDEVICE: true
+          ELEVENTY_IGNORE_NACL: true
+          ELEVENTY_IGNORE_PRIVACY_SANDBOX: true
+          ELEVENTY_IGNORE_VERSIONHISTORY: true
+          ELEVENTY_IGNORE_WEBSTORE: true
+          ELEVENTY_IGNORE_WORKBOX: true
+          # ELEVENTY_INCLUDE_TRANSLATED: true
```
matthiasrohmer, line 122: This can be trimmed down once https://github.com/GoogleChrome/developer.chrome.com/pull/4316 is in.

devnook (approved):

PR #4397, opened 2022-11-24
Fix view-transitions article’s updated field

bramus:
The view-transitions article says that the last update was in August, yet it was committed/merged earlier this week. Small oversight I guess?


jakearchibald (approved):
Hah, I always forget this

AaronForinton (approved):

PR #4413, opened 2022-11-25
DevTools 108: Settings > Ignore listing

sofiayem:
Documented https://developer.chrome.com/blog/new-in-devtools-108/#ignore-list
 
Publish on Nov 29.

P.S. These sections are to be reworked and moved to the upcoming Settings reference.


AaronForinton (approved):
LGTM

PR #4478, opened 2022-12-01
Secure Payment Confirmation shipping in Android Chrome blog post

agektmr:
Secure Payment Confirmation shipping in Android Chrome blog post.

### DIFF
--- a/site/en/blog/spc-on-android/index.md
+++ b/site/en/blog/spc-on-android/index.md
@@ -16,7 +16,7 @@ date: 2022-12-01
 Secure Payment Confirmation (SPC) is a [proposed web
 standard](https://www.w3.org/TR/secure-payment-confirmation/) that allows
 customers to authenticate with a credit card issuer, bank, or other payment
-service provider using a platform authenticator — typically activated with a
+service provider using a platform authenticator—typically activated with a
 device's screen unlock feature such as a fingerprint sensor. This usually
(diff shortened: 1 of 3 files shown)


agektmr:
cc: @stephenmcgruer

site/en/blog/spc-on-android/index.md
```diff
@@ -0,0 +1,148 @@
...
+{% Aside %}
+
+As SPC is specialised in payment authentication, you can use this API directly
+only if you are a credit card issuer, bank, or other payment service provider.
+Otherwise, please reach out to your credit card issuer, bank or payment service
+provider if you are interested in using this feature.
+
+{% endAside %}
+
+## Strong Authentication for payments
```
anusmitaray, line 56: Just one thing to double-check—are we using "Strong Authentication" as an official name—is that why we have it in title case?

I'm aware of [SCA](https://www.fca.org.uk/firms/strong-customer-authentication), is that the reference being made here? If not, should it not be "Strong authentication for payments"?

anusmitaray (approved):
Made minor edits and left a comment to just double-check something, but other than that LGTM

site/en/blog/spc-on-android/index.md
agektmr: Chatted offline. We'll use "Strong **a**uthentication" to avoid conflicting with the SCA.

PR #4431, opened 2022-11-28
feat: add cards component for Landing Pages

mamieorine:
Fixes #3960

Changes proposed in this pull request:

- Rename an exiting `featuredCard` component to `featuredBlogCard` component
- Creates an `featuredCard` component 
- Creates a new `featuredSection` and adds into **Workbox** landing page

PR #4484, opened 2022-12-01
feat: disable escaping for landing page content strings 

### LINKED ISSUE #4116: Disable automatic escaping for landing page content strings

At the moment it's not possible to style the content inside landing page macros. This is causing issues because it prevents content authors from using HTML entities (`&mdash;`) or basic Markdown styling (`*emphasis*`) inside these strings.

It looks like the most direct solution is to update how strings inserted to process them as Markdown and avoid escaping HTML.

**site/_includes/macros/landing-deco.njk** starting on line 14
```diff
          <span>
-            {{ title }}
+            {{ title | md | safe }}
          </span>
        </h2>
        <p class="gap-top-300 gap-bottom-600 md-only">
-          {{ desc }}
+          {{ desc | md | safe }} 
        </p>
```

This change would also need to be applied to other landing macros that output similar strings.

The inability to use HTML entities is my primary concern. If the team would rather not support Markdown in these strings, we could just use the `safe` pipe instead of both `md` and `safe`.

The only other alternative I briefly considered is to use tags instead of macros, but I wrote this off as it would essentially be a complete rewrite of the landing page system.

**Additional context**

N/A

mamieorine:
Fixes #4116

- Disable automatic escaping for landing page content strings


site/_includes/partials/toc-under-tree.njk
```diff
@@ -0,0 +1,11 @@
+{#
+  This is used as the TOC on smaller screens (<= 992) that makes use of the white space
```
devnook, line 2: Is this file intended for this Pr?

site/_includes/macros/landing-section.njk
```diff
@@ -24,18 +24,18 @@
               <div class="item">
                 <div class="item__inner height-full display-flex direction-column">
                   <div class="item__title">
-                    <a href="{{ post.url }}">{{ post.title or post.data.title }}</a>
+                    <a href="{{ post.url }}">{{ post.title or post.data.title | md | safe }} </a>
```
devnook, line 27: Once https://github.com/GoogleChrome/developer.chrome.com/pull/4431 lands, can we add examples of this component,  including markdown, to document & test that markdown is allowed and working?

mamieorine: Yes, we can add the example of using markdown with this component into the 'add a landing page' page

site/_includes/partials/toc-under-tree.njk
mamieorine: I have removed it

site/_includes/macros/landing-section.njk
```diff
@@ -24,18 +24,18 @@
...
+                  <div class="desc flex-1">{{ post.description or post.data.description | md | safe }}</div>
                 </div>
               </div>
             </div>
           {% else %}
             <div class="item height-full">
               <div class="item__inner
                 height-full">
-                <div class="item__title"><a href="{{ post.url }}">{{ post.title or post.data.title }}</a></div>
-                <p class="desc flex-1">{{ post.description or post.data.description }}</p>
```
devnook, line 38: Why change from p to div?

devnook:
@mamieorine ptal

site/_includes/macros/landing-section.njk
mamieorine: I brought back the p tag. I accidentally changed it while updating this section

mamieorine:
@devnook  I brought back the p tag and merged the main brach into this branch. Please help review this PR again. Thank you!

devnook:
Can you also please add the example to the components page?

devnook (approved):

Issue #2007, opened 2022-01-24
Required components for feature parity with web.dev


rachelandrew:
The following components are missing from developer.chrome.com when compared to web.dev or require different markup. I've ordered these in terms of priority.

1. Codelab callout `{% CodelabsCallout 'codelab-fix-sneaky-404', lang %}`
2. Figure - dcc requires the classes, web.dev now does not. Would be great to have these the same.
3. [CodePen](https://web.dev/handbook/web-dev-components/#codepen)
4. [Banners](https://web.dev/handbook/web-dev-components/#banners) 
5. [Blockquote with shortcode](https://web.dev/handbook/web-dev-components/#block-quotes)
6. [Columns](https://web.dev/handbook/web-dev-components/#columns) - on dcc these use a shortcode, web.dev uses the `switcher` class.
7. [Instructions](https://web.dev/handbook/web-dev-components/#instructions)
8. [Compat](https://web.dev/handbook/web-dev-components/#browsercompat) - lower priority on dcc as a lot of stuff on dcc this won't work on as we don't have compat data yet for experimental things. Though if we could design a good way to show this status so we could use the same widget that would be cool.

There are other components that don't exist on dcc but I'm not sure if they need to right now:

- [Self assessments](https://web.dev/handbook/self-assessment-components/)
- [Patterns](https://web.dev/handbook/web-dev-components/#codepattern)

devnook:
A tracker list for prioritized components:

- [ ] Codelab callout {% CodelabsCallout 'codelab-fix-sneaky-404', lang %}
- [ ] Figure - dcc requires the classes, web.dev now does not. Would be great to have these the same.
- [ ] CodePen
- [ ] Banners
- [ ] Blockquote with shortcode
- [ ] Columns - on dcc these use a shortcode, web.dev uses the switcher class.
- [ ] Instructions

mihajlija:
Adding thumbnails to the wishlist 🙏

PR #4502, opened 2022-12-03
Add BrowserCompat shortcode

bramus:
Adds a `BrowserCompat` shortcode. Code was copied in (and adapted from) web.dev’s repo.

Example post where I thought this was needed: media-query-range-syntax

Also see #2007


matthiasrohmer:
Closed in favor of https://github.com/GoogleChrome/developer.chrome.com/pull/4736.

PR #4522, opened 2022-12-05
Replace 'MV3+' with 'Manifest V3 or later'

jpmedley:
Example
- https://deploy-preview-4522--developer-chrome-com.netlify.app/docs/extensions/reference/action/#method-disable


dotproto (approved):
Nice, thanks for this, @jpmedley!

AmySteam (approved):

PR #4540, opened 2022-12-07
fix: prevent account selector from reappearing

### LINKED ISSUE #4012: Prevent account selector from re-appearing when accessing preview
After you initially logged in for the preview handler for updates you are again asked to select the account you want to use to authenticate.

via @tomayac.

mamieorine:
Fixes #4012

- Adds a `data-skip_prompt_cookie` attribute to prevent the google button re-rendering after logged in
- Ref: https://developers.google.com/identity/gsi/web/guides/toggle-display-with-cookies


mamieorine:
@devnook 

I have done this by adding `data-skip_prompt_cookie` attribute (Ref: https://developers.google.com/identity/gsi/web/guides/toggle-display-with-cookies)

However, since the exiting Client ID allows only accounts from Google organisation to sign in to access this page and I don't have one, so I cannot sign in to DDC Preview to test this issue.

devnook (approved):

PR #4554, opened 2022-12-08
Use the new 'extensions-news' tag.

jpmedley:
Fixes #SOME_ISSUE_NUMBER

-
-
-


site/en/blog/cws-policy-revamp/index.md
```diff
@@ -11,7 +11,7 @@ date: 2022-11-01
 hero: 'image/WlD8wC6g8khYWPJUsQceQkhXSlv1/Z3DIUjSZNKZzmdYcNL6m.jpg'
 alt: 'Photo by Sara Cohen on Unsplash'
 tags:
-  - extensions
+  - extensions-news
```
dotproto, line 14: I was going to ask if we could use both `extensions` and `extensions-news` until we have the new approach fully fleshed out. My concern was that I've historically pointed developers to https://developer.chrome.com/tags/extensions and https://developer.chrome.com/feeds/extensions.xml, so I was worried we would break their existing links or RSS subscriptions.

It looks like https://developer.chrome.com/tags/extensions already redirects to https://developer.chrome.com/tags/extensions-news. Could we also add a redirect for the XML feed? I'm good with either doing that in this PR or handling it in a separate one, dealer's choice :)

```suggestion
  - extensions
  - extensions-news
```

dotproto (approved):
LGTM, but I have a request in a comment we should resolve before merging.

site/en/blog/cws-policy-revamp/index.md
jpmedley: I'll add an XML redirect. It didn't occur to me because I didn't think anyone used RSS any more.

PR #4559, opened 2022-12-09
Tag cleanup

jpmedley:
Let's use "- extensions-news" for ephemeral extensions items and reserve "- extensions" for things that a perennial. (We may never need a perennial tag since all extensions articles are under the same structure. I want to make this change now, while it is easy.


jpmedley:
Something we wrong. This has things in it that it should not.

PR #4564, opened 2022-12-09
Clean up extensions tags.

jpmedley:
Let's use "- extensions-news" for ephemeral extensions items and reserve "- extensions" for things that a perennial. (We may never need a perennial tag since all extensions articles are in the same folder structure, but I want to make this change now, while it is easy.


jpmedley:
@IanStanion-google @dotproto: please read the PR description.

AmySteam (approved):
LGTM 👍 This is a great idea! 

@jpmedley Would you be opposed to keeping track of new conventions and procedures in the [Extension docs handbook](https://developer.chrome.com/docs/handbook/extensions/)?

Feel free to assign this task to me 🙂

IanStanion-google:
@jpmedley Could you define ephemeral and perennial in this context for my understanding?

PR #4567, opened 2022-12-09
Add information about Chrome 108 permissions changes.

jpmedley:
This puts an [item from Known Issues](https://developer.chrome.com/docs/extensions/mv3/known-issues/#webrequest-onauthrequired) in the API reference.

### DIFF
--- a/site/en/docs/extensions/reference/webRequest/index.md
+++ b/site/en/docs/extensions/reference/webRequest/index.md
@@ -4,11 +4,9 @@ api: webRequest
 
 ## Manifest
 
-You must declare the "webRequest" permission in the [extension manifest][1] to use the web request
+You must declare the `"webRequest"` permission in the [extension manifest][1] to use the web request
 API, along with the necessary [host permissions][2]. To intercept a sub-resource request, the
-extension needs to have access to both the requested URL and its initiator. If you want to use the
-web request API in a blocking fashion, you need to request the "webRequestBlocking" permission in
-addition. For example:
+extension needs to have access to both the requested URL and its initiator. For example:
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


site/en/docs/extensions/reference/webRequest/index.md
```diff
@@ -22,6 +20,11 @@ addition. For example:
 }
 ```
 
+If you are building policy-installed extensions for enterprises, and want to use the web request API
+in a blocking fashion, you need to request the `"webRequestBlocking"` permission. As of Chrome 108,
```
oliverdunk, line 24: Would it make sense to move the sentence about that to a separate paragraph? While everything here is technically correct I feel like it could be read as implying that only policy-installed extensions can use `webRequestAuthProvider`, when in reality it's available to all extensions.

oliverdunk (approved):
Happy from a technical review POV. @jpmedley, I'm not sure if it makes sense to re-assign this back to you so I'll let you decide where it goes next :)

jpmedley:
>Happy from a technical review POV. @jpmedley, I'm not sure if it makes sense to re-assign this back to you so I'll let you decide where it goes next :)

I din't go into enough detail about process yesterday. I apologize. 
1. Use GitHub's approval feature when you're done. One of the reasons our throughput is bad is that we're not giving clear signals that something is done. 
1. After your approval, assign it back to me for the typo review.
1. After I approve it, the PR creator merges it. (this is standard practice in Chrome DevRel. Sometimes articles have to publish on a particular day. The PR creator keeps track of that better than anyone else.

Also, we should all know the immutable law of text reviews: **There's always something to fix.**

No matter how many times you go through a piece of writing, there will always be something to fix. 

Therefore:

Unless you find a problem that's particularly severe (it brings down the apocalypse. Dogs and cats living together... MASS HYSTERIA!) open a ticket and move on with the review.

site/en/docs/extensions/reference/webRequest/index.md
```diff
@@ -330,7 +335,7 @@ chrome.webRequest.onBeforeRequest.addListener(
 ```
 
 As this function uses a blocking event handler, it requires the "webRequest" as well as the
```
AmySteam, line 337:
```suggestion
As this function uses a blocking event handler, it requires the `"webRequest"` as well as the
```

AmySteam (approved):
LGTM, just a quick suggestion and should be good to go 👍

PR #4571, opened 2022-12-09
Add landing page for favicon namespace

AmySteam:
- Adds a landing page for the Favicon API.

I will add an intro and examples on a separate PR


dotproto:
This API is a bit weird because it's not actually exposed in JavaScript like the others. I think adding this file may cause the build job to log an error about not seeing an API by that name.

AmySteam:
@dotproto I guess we can close this, and I will follow up with you on a better way to document this API. Thanks! 👍

PR #4573, opened 2022-12-09
WIP [lighthouse] use the web.dev guides for the Lighthouse CWV metrics

brendankenny:
WIP: staging for feedback, please do not merge

In practice, the Lighthouse docs covering the major CWV metrics (and their lab equivalents) have major overlap with the official CWV docs over on web.dev. Lighthouse should just use the core CWV guides for documentation since they explain what the metrics are, have more eyes on them keeping them up to date, and each one already has an extensive optimization guide that link to other Lighthouse audits when they're useful.

[d.c.c FCP](https://developer.chrome.com/docs/lighthouse/performance/first-contentful-paint/) -> [web.dev FCP](https://web.dev/fcp/)
[d.c.c LCP](https://developer.chrome.com/docs/lighthouse/performance/lighthouse-largest-contentful-paint/) -> [web.dev LCP](https://web.dev/lcp/)
[d.c.c TBT](https://developer.chrome.com/docs/lighthouse/performance/lighthouse-total-blocking-time/) -> [web.dev TBT](https://web.dev/tbt/)
[d.c.c TTI](https://developer.chrome.com/docs/lighthouse/performance/interactive/) -> [web.dev TTI](https://web.dev/tti/)

(note that the Lighthouse CLS metric has always linked to web.dev/cls, this just brings the others around)

--

also moves the old FMP, first-cpu-idle, and max-potential-FID metric guides to a different section in the performance table of contents to better differentiate them as deprecated and no longer present in the current Lighthouse report.


brendankenny:
I think we're not going to do this and try to make these articles strictly focused on Lighthouse specifics instead, everything else deferred to the main CWV docs.

PR #4574, opened 2022-12-10
Style updates 

AaronForinton:
Minor things to bring some docs up to speed with the [style guide](https://developers.google.com/style/word-list):

- i.e. and e.g. removals
- a few sentences switched to active voice
- removed some brackets
- removal 'please' and 'please note' and 'surface' as a transitive verb
- removed TL;DR. I've replaced these with 'Summary'. Those paragraphs could be labelled 'Introduction' more precisely but I'm not 100% convinced on that working better. In pretty much every case the heading could also be scrapped.
       
Definitely recommend viewing all my diffs with [whitespace hidden](https://github.com/GoogleChrome/developer.chrome.com/pull/4574/files?diff=split&w=1) because they come in looking very choppy otherwise :@


Adamiok:
Also fixes #4600

PR #4586, opened 2022-12-10
[Lighthouse] Fixed GitHub reference links

AaronForinton:


brendankenny (approved):
Ah, great catch! Thanks for updating these.

> I accidentally also fixed the ones being deleted in https://github.com/GoogleChrome/developer.chrome.com/pull/4573, so we may want to wait on merging this one until after that one

No problem. I still have to get some opinions from folks on that PR, so definitely don't hold up this change for that one. I'll just rebase when this lands.

rachelandrew (approved):

PR #4641, opened 2022-12-15
fix: translation of  [Does not use]

myamolane:
Fixes #SOME_ISSUE_NUMBER

- In the English version, combining the context of this article, I think it means that we should use HTTP2 for as many resources as we can. But in the Chinese version, the meaning changed to [For some resources, you should not use http2]. So I changed the translation to [You didn't use http2 for all resouces]. Or I can change it to '为所有的资源使用 Http/2' [Use HTTP/2 for all of your resources] 
-
-


AaronForinton:
Hey @jecfish, could you sign off on this change?

jecfish (approved):

AaronForinton (approved):

Issue #4327, opened 2022-11-17
Improve error messages for missing data during build


matthiasrohmer:
https://github.com/GoogleChrome/developer.chrome.com/pull/4277 migrated articles from web.dev without bringing author data along. This broke the 11ty build together with the rest of the checks.

The error message in this case was

```
[11ty] Problem writing Eleventy templates: (more in DEBUG output)
[11ty] 1. Having trouble writing template: "dist/en/articles/index.html" (via EleventyTemplateError)
[11ty] 2. (./site/_includes/layouts/blog-landing.njk)
[11ty]   Template render error: (/Users/mrohmer/Workspace/developers-chrome-com/site/_includes/partials/tags.njk) [Line 6, Column 16]
[11ty]   EleventyShortcodeError: Error with Nunjucks shortcode `Img` (via Template render error)
[11ty] 3. ERROR IN ./site/en/articles/index.md, IMG image/tcFciHGuF3MxnTr1y5ue01OGLBn2/PFaMfvDZoPorronbpdU8.svg: alt text must be a string, received a object (via Template render error)
[11ty]
[11ty] Original error stack trace: Error: ERROR IN ./site/en/articles/index.md, IMG image/tcFciHGuF3MxnTr1y5ue01OGLBn2/PFaMfvDZoPorronbpdU8.svg: alt text must be a string, received a object
```

Which unfortunately is entirely unhelpful and only gives a slight hint with the image url pointing to the default author avatar.

matthiasrohmer:
Idea: implement a `assert` shortcode in webdev-infra that can be called in templates to raise a specific error.

matthiasrohmer:
We decided to go for the underlying issues instead and make them less likely to happen. For example https://github.com/GoogleChrome/developer.chrome.com/pull/4966 fixed the particular issue described above.

PR #4738, opened 2022-12-29
Use Assert shortcode to check author data integrity

matthiasrohmer:
Fixes https://github.com/GoogleChrome/developer.chrome.com/issues/4327 by using the new `{% Assert %}` shortcode, introduced in https://github.com/GoogleChrome/webdev-infra/pull/42.

This provides a more helpful error message in case of missing author data. Before:

```
[11ty] 2. (./site/_includes/layouts/home.njk) [Line 23, Column 27]
[11ty]   EleventyShortcodeError: Error with Nunjucks shortcode `Img` (via Template render error)
[11ty] 3. ERROR IN ./site/en/index.md, IMG image/tcFciHGuF3MxnTr1y5ue01OGLBn2/PFaMfvDZoPorronbpdU8.svg: alt text must be a string, received a object (via Template render error)
[11ty]
```

After:

```
[11ty] 1. Having trouble writing template: "dist/en/blog/index.html" (via EleventyTemplateError)
[11ty] 2. (./site/_includes/layouts/blog-landing.njk)
[11ty]   Template render error: (/Users/mrohmer/Workspace/developers-chrome-com/site/_includes/partials/tags.njk) [Line 6, Column 16]
[11ty]   EleventyShortcodeError: Error with Nunjucks shortcode `Assert` (via Template render error)
[11ty] 3. No i18n data for author matthias (via Template render error)
[11ty]
```

Needs https://github.com/GoogleChrome/webdev-infra/pull/42 to be merged and published.

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -38,6 +38,7 @@ const {Aside} = require('./site/_shortcodes/Aside');
 const includeRaw = require('./site/_shortcodes/includeRaw');
 const {LanguageList} = require('./site/_shortcodes/LanguageList');
 const {Partial} = require('./site/_shortcodes/Partial');
+const {Assert} = require('webdev-infra/shortcodes/Assert');
 
 // Transforms
 const {domTransformer} = require('./site/_transforms/dom-transformer-pool');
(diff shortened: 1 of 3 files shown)


devnook:
See comments on the 

In short, my complaints:
1. We introduce shortcode not intended for authors
2. Low readability of code in case of this shortcode
3. Possibly the error text will be repeated on each usage of this shortcode
3. ... and on the other hand, if we only use it in few places, maybe it's better to add this logic to i18n filter directly?

An alternative would be to add {{ ... | assert }} filter, which always just logs that "Value xyz.abc was not found / is undefined". This is shorter, more readable and not a shortcode.

matthiasrohmer:
Discussed this in chat with @devnook. We rather invest in checking data integrity before template rendering kicks off and if that's not possible provide proper fallbacks at template level.

PR #4747, opened 2022-12-30
I18N extension docs inconsistency with observed behavior

rubi-rastko:
_locale message ~reference~ defnition should be without \_\_MSG\_\_

Fixes <not-found>

- \_\_MSG_tooltip\_\_ where ~referenced~ defined in ~manifest.json~ messages.json should be "tooltip"
-
-


jpmedley:
@rubi-rastko thank you for your submission. Unless I'm missing something, the existing content is correct. `"__MSG_tooltip__"` is a variable defined in the appropriate `messages.json` and dereferenced to "tooltip" using `manifest.json'.

PR #4791, opened 2023-01-05
Make content-type demo pages use example tag and author

matthiasrohmer:
@rachelandrew noticed that the new Demo pages for all of our content types, introduced with https://github.com/GoogleChrome/developer.chrome.com/pull/4450 and ff., are showing up in places where they shouldn't, namely the author detail page: https://developer.chrome.com/authors/mihajlija/

This PR adds a new author and tag, both named Example, which are excluded from their respective collections which prevents a /tags/example and a /author/example page from showing up.


devnook (changes requested):
I think we need to exclude this content at the creation of authors collection point of time: https://github.com/GoogleChrome/developer.chrome.com/blob/main/site/_collections/authors.js#L31

and be more granular about what we exclude (e.g. all Handbook section) rather than rely on noindex.

matthiasrohmer:
Updated the PR to have a a new author and tag, both named `Example`, which are excluded from their respective collections.

devnook (approved):

PR #4797, opened 2023-01-05
Updated origin trial docs to cover third-party token troubleshooting in more detail

samdutton:
Also minor fixes to URLs, wording and layout.

### DIFF
--- a/site/en/docs/web-platform/origin-trial-troubleshooting/index.md
+++ b/site/en/docs/web-platform/origin-trial-troubleshooting/index.md
@@ -10,7 +10,7 @@ description: >
 authors:
   - samdutton
 date: 2021-08-11
-updated: 2022-08-31
+updated: 2023-01-05
 hero: image/80mq7dk16vVEg8BBhsVe42n6zn82/b52LlVcFfbFtxgfT0BoF.jpg
 alt: Test tubes in a metal rack, one containing clear green liquid.
 tags:
@@ -71,7 +71,7 @@ To troubleshoot an origin trial, work through each of the issues below using the
   <br>
   <input class="w-checkbox" type="checkbox" id="check-token-third-script">
   <label for="check-token-third-script" class="w-ml--l"><a href="#token-third-script">Third-party 
-token is provided via an external script, not a meta tag or inline script</a></label>
+token is provided via an external script, not a meta tag, HTTP header or inline script</a></label>
   <br>
   <input class="w-checkbox" type="checkbox" id="check-token-method">
   <label for="check-token-method" class="w-ml--l"><a href="#token-method">Origin trial feature access 
@@ -120,6 +120,7 @@ The demos below show each of the ways to provide an origin trial token and acces
 * [ot-header.glitch.me](https://ot-header.glitch.me): token in an `Origin-Trial` response header
 * [ot-3p.glitch.me](https://ot-3p.glitch.me): token injected by a third-party script
 * [ot-iframe.glitch.me](https://ot-iframe.glitch.me): origin trial feature accessed in an iframe
+* [ot-iframe-3p.glitch.me](https://ot-iframe-3p.glitch.me): cross-origin iframe examples
 
 
 ## Use Chrome DevTools to check tokens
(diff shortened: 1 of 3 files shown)


site/en/docs/web-platform/origin-trial-troubleshooting/index.md
```diff
@@ -563,12 +565,24 @@ Check that the origin trial is [enabled for the Chrome versions accessing your s
...
+cross-origin iframe examples.
 
+{% Aside 'caution' %}
+
+When you register for a trial, don't select _Third-party matching_ just because you plan to access a
+trial feature from an iframe!
+
+A [third-party token](/docs/web-platform/third-party-origin-trials/#register-for-a-third-party-origin-trial)
+will only activate a trial feature if it's provided in an external JavaScript file included via a
+`<script>` element: a third-party token won't work in a meta tag, inline script or HTTP header.
```
heyawhite, line 583:
```suggestion
A [third-party token](/docs/web-platform/third-party-origin-trials/#register-for-a-third-party-origin-trial)
only activates a trial feature if it's provided in an external JavaScript file, included via a
`<script>` element. A third-party token won't work when provided in a meta tag, inline script or HTTP header.
```

site/en/docs/web-platform/origin-trials/index.md
```diff
@@ -138,8 +137,22 @@ registration and provide a new token, for each origin enrolled in the trial.
 3. Copy the new token and make it available for every page that should remain enrolled in the trial.
 You can provide multiple tokens if necessary: Chrome will ignore invalid or expired tokens.
 
+{% Aside %}
+
+Usually if an API lands unchanged after a successful origin trial, there is a
+short period between the end of the origin trial and the date the implementation ships in the
+browser when the API is unavailable. This is by design. If Chrome were to avoid the
+mandatory total-breakage period, that would bias toward avoiding breakages in the API surface. Such
+breakages are often needed to improve the API. The final shipping API might be worse for it.
```
heyawhite, line 146: "If Chrome were to avoid the mandatory total-breakage period, that would bias toward avoiding breakages in the API surface."

Who is the owner of "bias"? I'm a bit confused by the sentence in context.

Would this be correct:

```suggestion
Usually if an API lands unchanged after a successful origin trial, there is a
short period between the end of the origin trial and the date the implementation
ships in the browser when the API is unavailable. This is intentional and known
as the total-breakage period. Chrome allows an opportunity for breakage in the
API surface, which is often needed to improve the API. The final shipping API
may be worse without this period.
```

site/en/docs/web-platform/origin-trials/index.md
```diff
@@ -138,8 +137,22 @@ registration and provide a new token, for each origin enrolled in the trial.
...
+
+In some circumstances, if there was clear evidence that developers engaged with the origin trial and
```
heyawhite, line 148:
```suggestion
In some circumstances, if there is clear evidence that developers engaged with the origin trial and
```

heyawhite (approved):
Two nits and one question around understanding the breakage period for your attention.

site/en/docs/web-platform/origin-trials/index.md
samdutton: I agree this is a bit hard to understand!

I like your rephrasing. @jpchase Does Alexandra's version sound OK to you here?

samdutton:
> A thing to look at either in this PR or the next is heading length. The headings in [Troubleshoot Chrome origin trials](https://pr-4797-static-dot-dcc-staging.uc.r.appspot.com/docs/web-platform/origin-trial-troubleshooting/) are very long, which squishes the main body content and can create a scroll bar at the bottom.
> 
> I recommend shortening headings to 8 words or less, ideally under 55 characters (with spaces).
> 
> Screenshot highlighting visual issues

On [Troubleshoot Chrome origin trials](https://developer.chrome.com/docs/web-platform/origin-trial-troubleshooting/) on the live site, I can't reproduce this (but I may not be trying hard enough!)

I agree that the current approach results in very long headings (which aren't really headings...)

heyawhite:
@samdutton it looks like eng resolved this problem on the live site by allowing word wrap! So, need to adjust for technical reasons!

site/en/docs/web-platform/origin-trials/index.md
jpchase: The requirement for a breakage period has actually been [removed](https://www.chromium.org/blink/launching-features/#step-5-optional-origin-trial:~:text=%5BNote%3A%20the-,required%20breaking%20period,-was%20removed%20in). We're ~9 months into the 1 year temporary removal, so it seems likely to become permanent.

I suggest removing the section about breakage entirely. If they do decide to reinstate, we can always add something back.

samdutton: Thanks @jpchase! I've updated the aside to explain the change (with links).

samdutton:
Thanks @heyawhite! I've merged.

PR #4863, opened 2023-01-12
Update document.domain timeline

agektmr:
document.domain setter deprecation is postponed to Chrome 111. This change adds a note about the update.


site/en/blog/immutable-document-domain/index.md
```diff
@@ -18,6 +18,8 @@ alt: >
 
 **Updates**
 
+- **January 12, 2023**: Updated timeline—`document.domain` setter will be removed
+    by default starting from Chrome 111.
```
anusmitaray, line 22: Just confused about whether the update is the same for a different version.

PR #4864, opened 2023-01-12
Chrome 110 beta

rachelandrew:


zloirock:
Hey, I see that https://github.com/tc39/proposal-change-array-by-copy support is unflagged in 110. Why is nothing about it in this post or will it not be exposed to the stable release?

PR #4902, opened 2023-01-15
feat(social): add social links to post author macro, fixes #926

### LINKED ISSUE #926: Add twitter, github etc attribution to posts, just like web.dev does
Can we add these attribution links to the author section at the top of blog posts? e.g.

https://developer.chrome.com/blog/renderingng/

has no attribution, but I'm pretty sure @rachelandrew added them in the metadata. My understanding is that this feature isn't part of the template for rendering posts yet.

MichaelSolati:
Fixes #926

- Add social links to post-author macro
- Add simple CSS for links

image

@devnook I think that the links for the post card and the authors page may be able to be made into one macro, and just have a flag to display images or text, but I didn't want to introduce that unless you agreed. I think this is the minimum required for the feature.

### DIFF
--- a/site/_includes/macros/post-author.njk
+++ b/site/_includes/macros/post-author.njk
@@ -5,9 +5,10 @@
 #}
 {% macro postAuthor(authorId, locale) %}
 
+{% set author = authorsData[authorId] %}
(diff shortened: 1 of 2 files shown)


devnook (approved):
Thanks @MichaelSolati !

PR #4907, opened 2023-01-16
Clarify behaviour when MacCtrl is used on other platforms.

oliverdunk:
A comment about MacCtrl in the commands docs was unclear about what would happen if this keyword was used on platforms other than macOS. I've reworded it to avoid using "uninstallable" and hopefully clarify the intended meaning.


jpmedley:
@oliverdunk DevRel policy is that the author of the PR does the merge. Please do so now.

PR #4920, opened 2023-01-16
Add clean-up task for staging environment

### LINKED ISSUE #4706: Clean-up staging bucket directories and instances
https://github.com/GoogleChrome/developer.chrome.com/pull/4527 introduces a new staging environment which creates directory in Google Cloud Storage and sometimes a new instance in App Engine for every PR. As soon as a PR is merged or closed those need to be cleaned.

As the Google Cloud Build trigger does not listen to the PR close event, there are two alternatives:

1) Create a GitHub action workflow that listens on the PR close event and then calls a Google Cloud Build job via webhook that removes unused artifacts. Webhook is the only option for this way round, as there are no allowed ways to auth GitHub to Google Cloud.
2) Create a Google Cloud Build job that runs on interval, gets all closed PRs from the GitHub API and checks for which there are still artifacts and removes them.

matthiasrohmer:
The new staging workflow leaves some artifacts with every deployment: at least a directory named `pr-{number}` in the Cloud Storage bucket and if serving logic changed also an App Engine version with a version ID of `pr-{number}-app`.

Those remains need to be cleaned up as soon as a PR is closed. Therefore this gulp task which can be called manually, if there is an exported `GITHUB_TOKEN` and an active `gcloud` CLI authentication, or on schedule via Google Cloud Build.

It fetches the list of open PRs from GitHub (limited to the latest 100, as no pagination logic yet) and compares them to the root level objects of the bucket and the App Engine versions and removes the ones that don't appear in the GitHub API response.

Fairly simple, the thing that makes it a little confusing is normalizing version names and bucket object paths to PR numbers.

Resolves #4706.


matthiasrohmer:
I have no idea why the linter is complaining about the md filter, it hasn't been touched in this PR: https://github.com/GoogleChrome/developer.chrome.com/actions/runs/3932951334/jobs/6726120805#step:6:15

gulp-tasks/stageGitHub/cleanUpGoogleCloud.js
```diff
@@ -0,0 +1,192 @@
...
+const PROJECT_ID = 'dcc-staging';
+
+/** This folder contains the base version of the site and should
+ * never be deleted.
+ */
+const FALLBACK_DIRECTORY = 'main/';
+
+/**
+ * The version name of the default App Engine version, which should
+ * never be cleaned up (but also can't and would )
```
devnook, line 46: fix comment

matthiasrohmer: Done!

gulp-tasks/stageGitHub/cleanUpGoogleCloud.js
```diff
@@ -0,0 +1,192 @@
...
+/**
+ * Fetches the open pull requests from GitHub and
+ * maps the API response to only get the pull request numbers.
+ * @returns {Promise<Array<number>>} [4124, 4125, 4126] for example
+ */
+async function getOpenPullRequestsFromGitHub() {
+  try {
+    const apiResponse = await requestGitHubApi(
+      'GET',
+      'pulls?state=open&per_page=100'
```
devnook, line 87: I'd record that 100 in some top-level variable, to make it clear we only check last 100. I think 100 is sufficient, just to make it more future proof, in case someone wonders why not PRs are accoutned for

devnook (approved):

gulp-tasks/stageGitHub/cleanUpGoogleCloud.js
```diff
@@ -0,0 +1,192 @@
...
+  if (staleDeployments.length === 0) {
+    return;
+  }
+
+  await cleanUpCloudStorage(staleDeployments);
+  const appEngineVersions = await getAppEngineVersions();
+  // There is a posibility that there are only static deployments
+  // and no instances except the default one, then we can exit early
+  if (appEngineVersions.length === 1) {
+    console.log('No app deployments found.');
```
devnook, line 184: Exit?

gulp-tasks/stageGitHub/cleanUpGoogleCloud.js
matthiasrohmer: Added a comment instead. As the job is intended to run periodically it will pick up overflown items with the next run, even if we wouldn't set per_page and stick to the default.

PR #4921, opened 2023-01-16
Update Google Cloud Build staging job to use secret version 2

matthiasrohmer:
Well, that's a simple one for a change.

The previous secret stored a token that was only valid for 30 days. I now issued a token valid for a year, which is the current maximum for the new fine-grained GitHub PATs. The new token will expire on Sat, Jan 13 2024. Advice on what's the best way to track this is much appreciated.


matthiasrohmer:
Superseded by #5128.

PR #4926, opened 2023-01-16
Add RSS feed for articles

### LINKED ISSUE #4924: RSS feed for articles
The blog has a RSS feed set up on https://developer.chrome.com/feeds/blog.xml. Would be nice if [the articles section](https://developer.chrome.com/articles/) also had a feed, to make them more easy to follow.

bramus:
If this feature is wanted, this PR fixes #4924.

- Add an RSS feed for the articles

### DIFF
--- a/site/_includes/layouts/base.njk
+++ b/site/_includes/layouts/base.njk
@@ -9,7 +9,8 @@
 
     {% include 'partials/meta.njk' %}
     {% block feed %}
-      <link rel="alternate" type="application/rss+xml" title="{{ site.title }}" href="{{ site.url }}/feeds/blog.xml">
+      <link rel="alternate" type="application/rss+xml" title="{{ site.title }} - {{ 'i18n.common.blog' | i18n(locale) }}" href="{{ site.url }}/feeds/blog.xml">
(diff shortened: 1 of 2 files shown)


site/_includes/layouts/base.njk
```diff
@@ -9,7 +9,8 @@
 
     {% include 'partials/meta.njk' %}
     {% block feed %}
-      <link rel="alternate" type="application/rss+xml" title="{{ site.title }}" href="{{ site.url }}/feeds/blog.xml">
+      <link rel="alternate" type="application/rss+xml" title="{{ site.title }} - {{ 'i18n.common.blog' | i18n(locale) }}" href="{{ site.url }}/feeds/blog.xml">
```
devnook, line 12: Since this is the main layout, shouldn't we link to all.xml instead?
@rachelandrew wdyt?

bramus: Checking the contents of that all feed, it seems utterly broken. This might explain why it wasn’t exposed in the first place.

devnook: I think many RSS readers will just pick the first linked feed anyways, and ignore the other though. So it would be better to fix all.xml, or combine blog + articles in one feed instead of exposing both urls.

bramus: Good call. I can easily correct the “all” feed generation and list it first.

matthiasrohmer (approved):

emanruse:
There are no RSS feeds any more.

PR #4932, opened 2023-01-17
added more description about process

sourav0v0:

- Introduced main thread concept 
- The whole description dosent introduces the main thread concept which dosnt gives a clear idea to understand the releationship between process and thread


sourav0v0:
@rachelandrew Please review

rachelandrew (commented):
This directly contradicts the previous sentence which states the threads are optional. This is quite an old article and probably needs more work, I'll see if someone can take a proper look at it.

PR #4937, opened 2023-01-17
Update FPS instructions

mihajlija:
Removes the clear data flag from FPS instructions because it's on by default now.

To do:

- [ ] @rowan-m remove the flag from the demo

### DIFF
--- a/site/en/blog/first-party-sets-testing-instructions/index.md
+++ b/site/en/blog/first-party-sets-testing-instructions/index.md
@@ -67,7 +67,7 @@ To see a FPS demo in action, visit [https://first-party-sets.glitch.me/](https:/
 
 ```text
 google-chrome \
---enable-features="FirstPartySets:FirstPartySetsClearSiteDataOnChangedSets/1,StorageAccessAPI,StorageAccessAPIForOriginExtension,PageInfoCookiesSubpage,PrivacySandboxFirstPartySetsUI" \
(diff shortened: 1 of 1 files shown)


site/en/blog/first-party-sets-testing-instructions/index.md
```diff
@@ -83,10 +83,12 @@ To enable FPS locally, you need to use Chrome's  `--enable-features` option with
...
 
 ```text
-FirstPartySets:FirstPartySetsClearSiteDataOnChangedSets/1
+FirstPartySets
 ```
 
-Starting in Chrome M109, the additional parameter `FirstPartySetsClearSiteDataOnChangedSets/1`  will enable [clearing site data in case members of a set change](https://github.com/WICG/first-party-sets#clearing-site-data-on-set-transitions). This prevents linking a user's identity across all the sets a site has historically been in. This behavior will be the default in Chrome.
+{% Aside %}
+Starting in Chrome M109, [site data will be cleared in case members of a set change](https://github.com/WICG/first-party-sets#clearing-site-data-on-set-transitions). This prevents linking a user's identity across all the sets a site has historically been in. 
+{% endAside %}
```
krgovind, line 91: Can we add something to the effect of: In M108, this functionality can be enabled using the flag `FirstPartySets:FirstPartySetsClearSiteDataOnChangedSets/1`.

PR #4961, opened 2023-01-19
Popover API article + docs update for resolutions

jh3y:
Hey 👋 

There were some new resolutions at the back end of last year that changed the shape of the API slightly removing some events and attributes, etc.

I've updated the article and docs along with the corresponding demos to reflect this. Also fixed up the redirects.


matthiasrohmer:
This has been superseded by https://github.com/GoogleChrome/developer.chrome.com/pull/6197 which moves the Popover API docs to MDN.

Issue #4954, opened 2023-01-18
"[object Object]" in published site


paulirish:
seems similar to #4930

**Describe the bug**
https://developer.chrome.com/blog/reporting-api-migration/#whats-changing has [object Object] in the page
<img width="518" alt="image" src="https://user-images.githubusercontent.com/39191/213285876-75ab2ae0-93b9-4867-b521-f268a81b347b.png">

Presumably because of some generator use?

That page's source: https://github.com/GoogleChrome/developer.chrome.com/blob/main/site/en/blog/reporting-api-migration/index.md
<img width="595" alt="image" src="https://user-images.githubusercontent.com/39191/213285958-3f89fb13-db03-4193-906e-6e257dea7d78.png">

**To Reproduce**
1. Loook at https://developer.chrome.com/blog/reporting-api-migration/#whats-changing

Issue #4930, opened 2023-01-17
RSS feeds have corrupt `<title>`, reading `Chrome Developers - [object Object]`


matthiasrohmer:
While looking at #4924 I noticed some of the RSS feeds have `Chrome Developers - [object Object]` in their `<title>` instead of the proper one:

- developer.chrome.com/feeds/api.xml
- developer.chrome.com/feeds/backpressure.xml
- developer.chrome.com/feeds/caching.xml
- developer.chrome.com/feeds/contacts.xml
- developer.chrome.com/feeds/cors.xml
- developer.chrome.com/feeds/feedback.xml
- developer.chrome.com/feeds/idle-detection.xml
- developer.chrome.com/feeds/index.xml
- developer.chrome.com/feeds/monitoring.xml
- developer.chrome.com/feeds/notification-triggers.xml
- developer.chrome.com/feeds/notifications.xml
- developer.chrome.com/feeds/sensors.xml
- developer.chrome.com/feeds/service-worker.xml
- developer.chrome.com/feeds/shape-detection.xml
- developer.chrome.com/feeds/video.xml
- developer.chrome.com/feeds/wake-lock.xml
- developer.chrome.com/feeds/web-app-manifest.xml
- developer.chrome.com/feeds/websocket.xml
- developer.chrome.com/feeds/websocketstream.xml

PR #4966, opened 2023-01-19
Add missing i18n for tags, properly handle free-form labels for `{% Compare %}`, use webdev-infra i18n filter

matthiasrohmer:
Fixes https://github.com/GoogleChrome/developer.chrome.com/issues/4930, fixes https://github.com/GoogleChrome/developer.chrome.com/issues/4954. Follow-up to https://github.com/GoogleChrome/developer.chrome.com/issues/4327.

https://github.com/GoogleChrome/developer.chrome.com/pull/3998 introduced a new `<checkbox-group>` web component, originally meant for the events page but currently not used, which uses an object for interface copy. To enable this, the i18n filter was changed to flatten translations into an object, if no specific translation could be found. 

By this change some templates now receive an empty object from the i18n filter, instead of undefined. This leads to `[object Object]` showing up in weird places, like in #4930 or #4954 or confusing error messages like described in #4327.

To fix, this PR makes d.c.c use the i18n filter from the webdev-infra package, which is an optimized version of the one from web.dev. This version of the filter throws an error instead of returning `{}` for an undefined translation. That also revealed that for a lot of tags there weren't any translations, preventing tag pages for those tags getting generated or showing up on the tags overview page, and also made `[object Object]` showing up in a lot of RSS feeds.

The missing author descriptions are taken from the people's LinkedIn profiles, verified it's them by looking at their linked Twitter profiles.

Additionally this PR alters the `{% Compare %}` shortcode to properly handle the `labelOverride` argument (#4954). The fix without updating the i18n filter would have been to check if `i18n` returned an empty object, which would be weird. With the behaviour described above it can now properly try/catch. By looking through the current uses of the `labelOverride` argument, which almost always is free-form, this seems to follow the user expectation of what that argument is there for.

All in all this PR brings down the results for searching for `[object Object]` in the build output (`./dist`) down from 180x to just 1x in the main.js bundle, where it's actually expected.

The change has been tested by diffing the `dist` directory of `npm run production` before and after the change, which only lists the new tags pages and the article pages that previously had `[object Object]` in their tag lists. Additionally I searched for `i18n.` in the build output, which revealed that some of the `alt` attributes on the /meet-the-team page also weren't translated correctly, that's also fixed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -53,7 +53,7 @@
         "unistore": "^3.5.2",
         "uslug": "^1.0.4",
         "web-vitals": "^3.0.4",
-        "webdev-infra": "1.0.41",
+        "webdev-infra": "1.0.42",
         "xss": "^1.0.14"
       },
       "devDependencies": {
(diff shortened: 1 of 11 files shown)


matthiasrohmer:
As https://github.com/GoogleChrome/developer.chrome.com/pull/4736, this requires webdev-infra 1.0.42.

PR #5006, opened 2023-01-23
New Post - `scrollend`

argyleink:
ToDo:
- [x] get final reviews
- [x] update date
- [x] remove shortcode for BCD if it isn't ready to go yet.


rachelandrew (approved):
This LGTM, the compat component is outputting 'this short code is not supported", I'm chasing up to see if we can get that merged in yet, if not we can remove the compat line and I'll put it back in once it is.

argyleink:
> this short code is not supported

i saw this too and left it in because of previous conversations we had about these "just working" later when the module is supported. i'll wait to hear back from ya!

PR #5016, opened 2023-01-24
feat: expose IDs on landingsection macros

### LINKED ISSUE #4053: Expose IDs on landingSection macros

At the moment it's not possible to link directly to given section on a landing page because sections do not have IDs. Currently, the closest thing we can do is use a text link to reference the title of a section. For example: https://developer.chrome.com/workbox/#:~:text=what%20you%20need%20to%20know

IDs for section titles should be automatically assigned similar to how headers are automatically given IDs. Ideally, content authors would also be able to manually provide IDs to account for string changes.

At the moment, the only workaround I can think of is to manually embed an anchor with an ID before a section (e.g. `<a id="example">`). That said, I haven't experimented with this approach.

**Additional context**

N/A

simgobb:
Fixes #4053

- Expose id for landingsection macros
- Allow custom id string, if not specified use the section title

### DIFF
--- a/site/_includes/macros/landing-section-expanded.njk
+++ b/site/_includes/macros/landing-section-expanded.njk
@@ -2,14 +2,15 @@
 {% from 'macros/landing-deco.njk' import landingDeco with context %}
 {% from 'macros/related-articles.njk' import relatedArticles with context %}
 
-{% macro landingSectionExpanded(title, desc, items, color, type, ordered, imgSrc, imgAlt, imgWidth, imgHeight, smallImgSrc, relatedTitle, relatedItems) %}
+{% macro landingSectionExpanded(title, desc, items, color, type, ordered, imgSrc, imgAlt, imgWidth, imgHeight, smallImgSrc, relatedTitle, relatedItems, customId) %}
+{% set sectionId = customId if customId else title %}
 
 <section
   class="landing-section landing-section--{{type}} {% if ordered %} landing-section--ordered {% endif %} hairline rounded-lg width-full display-grid gap-top-400"
-  style="--number-color: var(--color-{{ color }}-medium); --dot-color: var(--color-{{ color }}-lighter);">
+  style="--number-color: var(--color-{{ color }}-medium); --dot-color: var(--color-{{ color }}-lighter);"
+  id="{{sectionId | lower | replace(" ", "-")}}">
 
   {{ landingDeco(title, desc, color, imgSrc, imgAlt, imgWidth, imgHeight, smallImgSrc) }}
-
   <div class="gap-top-200 gap-left-200 gap-right-200">
     <div class="landing-section__links">
       {% for item in items %}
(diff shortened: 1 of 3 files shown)


site/_includes/macros/landing-section-expanded.njk
```diff
@@ -2,14 +2,15 @@
 {% from 'macros/landing-deco.njk' import landingDeco with context %}
 {% from 'macros/related-articles.njk' import relatedArticles with context %}
 
-{% macro landingSectionExpanded(title, desc, items, color, type, ordered, imgSrc, imgAlt, imgWidth, imgHeight, smallImgSrc, relatedTitle, relatedItems) %}
+{% macro landingSectionExpanded(title, desc, items, color, type, ordered, imgSrc, imgAlt, imgWidth, imgHeight, smallImgSrc, relatedTitle, relatedItems, customId) %}
+{% set sectionId = customId if customId else title %}
```
devnook, line 6: You need to ensure that the id is unique in some way.

devnook (commented):
Hi @simgobb ! Please make sure the new field is documented, so that authors can learn how to use it. If the component you add field to is not documented yet, you need to create the documentation for the whole component. Most probably here: https://developer.chrome.com/docs/handbook/how-to/add-a-landing-page/

simgobb:
@matthiasrohmer @devnook 

Looked into adding documentation for these components. I have added sample ones here: https://developer.chrome.com/docs/handbook/content-types/landing/.

As per previous comment I've tried adding documentation here as well: https://developer.chrome.com/docs/handbook/how-to/add-a-landing-page/

The issue is that the landing page components don't render correctly within a blog post. Mentioned this to Matthias who suggested raising a separate issue for this in the future to include documentation on how to build landing pages

site/_filters/unique-id.js
```diff
@@ -0,0 +1,38 @@
...
+    }).length;
+
+    pageIds.push(idUrl);
+
+    return `${id}-${count}`;
+  }
+
+  pageIds.push(idUrl);
+  return id;
+}
```
matthiasrohmer, line 34: I'd simplify this quite a bit. The idea makes sense, but I would leave it to the author to choose correct IDs, as they might also be a simple copy-paste errors. No one will probably build landing pages in a hurry anyway where it would make sense to auto-fix it for them. This would be my approach:

```js
/**
 * @param {string} id of the section
 * @this {any} The eleventy context
 * @return {string} unique id for the section
 */

function ensureUniqueHref(id) {
  const key = `${this.ctx.page.url}#${id}`;

  // Check if there is already an URL with this key, and if not, add it
  // to the set and exit early.
  if (!pageIds.has(key)) {
    pageIds.add(key);
    return id;
  }

  throw new Error(
    `There is another element with the ID ${id} on the page ${this.ctx.page.url}. Please make sure all IDs are unique.`
  );
}
```

So just throwing if an ID duplicates should be fine.

matthiasrohmer (approved):

PR #5026, opened 2023-01-25
feat: add outdated announcement banner

### LINKED ISSUE #4830: A standard way to indicate a blog post is outdated and link to the new material
Now that we have a documentation project ongoing, I'm going to be a bit more active in chasing docs out of the blog in preference for blog posts being news (ref: [where to publish](https://developer.chrome.com/docs/handbook/where-to-publish/)). As they are news, what we want to avoid is people going back and making major edits to an old post due to the API changing. This rewrites history, means that people can't track the evolution of ideas, and is just confusing.

That said, we do want to avoid people stumbling across old blog posts and thinking they are current, so I'd like to have some kind of standard callout for the top of posts that we can add without fundamentally changing the content of the post, but that clearly highlights that this is historical info, and points to new material (new blog posts, the MDN docs etc.)

mamieorine:
Fixes #4830 

- Adds `newMaterialURL` frontmatter key to indicate outdated posts and link to new material. 
- Adds example usage in _example.md and description in /handbook/how-to/
- Changes the global red-darkest colour var from `#A50e0E` to `#9F0E0E`


site/_includes/layouts/blog-post.njk
```diff
@@ -28,6 +28,10 @@
   <div class="display-flex gap-top-300 lg:gap-top-400">
     <div class="display-flex justify-content-center width-full">
       <article class="post stack measure-long width-full pad-left-400 pad-right-400">
+        {% if newMaterialURL %}
```
matthiasrohmer, line 31: All of our multi-word YAML keys are snake case (by it being related to Python): `stack_overflow_tag`, `project_icon_color`, all of i18nk keys,
`project_key`, `origin_trial`. So it would be good to follow that scheme for this one. `new_material_url` also sounds quite specific. `new_material` would already work for me.

site/_includes/partials/post-banner.njk
```diff
@@ -0,0 +1,3 @@
+<div class="post-banner">
+  This content is outdated and no longer accurate. <a href={{ newMaterialURL }}>New material available</a>.
```
matthiasrohmer, line 2: Please don't use fully qualified URLs in the frontmatter. They may end up broken if the content is removed/moved. Please use the path to the document instead and derive the URL via 11ty filters.

site/_scss/blocks/_post-banner.scss
```diff
@@ -0,0 +1,13 @@
+.post-banner {
+  --link-color: var(--color-red-darkest);
+  --link-visited-color: var(--color-red-darkest);
+  --link-rgb-background: var(--rgb-red-darkest);
+  background-color: get-color('red-50');
```
matthiasrohmer, line 5: Isn't that `var(--color-red-lightest)` as in the warning aside?

site/_scss/globals/_colors.scss
```diff
@@ -11,7 +11,7 @@ $google-colors: (
   'blue-200': #aecbfa,
   'blue-100': #d2e3fc,
   'blue-50': #e8f0fe,
-  'red-900': #a50e0e,
+  'red-900': #9f0e0e,
```
matthiasrohmer, line 14: Those are the hardcoded material colors and may not be changed. Please note that would also change all other red hues on the site.

matthiasrohmer: By looking at the example from the PR I see the change is veeeery slight. Anyway, I know that has been a problem with other PRs before. Ewa's call :)

site/_includes/partials/post-banner.njk
matthiasrohmer: To be more specific: we have a helper `findByPath` that you could export from /site/_data/helpers.js. In the templates that should allow you to construct a URL by using `helpers.findByPath(new_material).url` then, which should be less error-prone.

site/_scss/blocks/_post-banner.scss
mamieorine: `var(--color-red-lightest)` in the warning aside is using 'red-50' var with 0.5 opacity. I asked Dan to confirm about this and he said this announcement banner is floating banner so it makes sense to have it solid.

site/_scss/globals/_colors.scss
mamieorine: According to the note from Dan, the red-darkest colour var has been darkened slightly to achieve AAA contrast (#a50e0e > #9f0e0e). That change will need to be applied globally.
Screenshot 2023-01-26 at 16 04 24

Also, I have seen that `red-900` had been used only for `--rgb-red-darkest` var. So I decided to change the HEX code here to achieve AAA contrast with `--rgb-red-lightest`.

devnook: I called this out specifically to Dan that our corrent palette does not fulfill accessibility requirements, with an ask to get it updated globally. So this is as expected.

site/_scss/blocks/_post-banner.scss
devnook: Generally, we should be using colors from _theme.scss, because if you use colord directly you cannot implement theme in a correct way (e.g. imagine you need to introduce dark theme in next sprint, which is actually very realistic).

I don't see why it wouldn't be --color-red-lightest or --color-red-lighter.

devnook (commented):
Thanks @mamieorine ! Left some comments.

site/_scss/blocks/_post-banner.scss
mamieorine: The existing code is:
```
  --color-red-lightest: #{rgb(get-color('red-50'), 0.4)};
  --color-red-lighter: #{rgb(get-color('red-50'), 0.5)};
```

However, Dan wants the new announcement banner to have the solid colour (without opacity). So I cannot use `--color-red-lightest` or `--color-red-lighter`. Is that okay if I add a new variable to the theme,  like `--color-red-solidLighter: #{get-color('red-50')};`?

site/_includes/partials/post-banner.njk
mamieorine: I have fixed this 😃

site/_includes/layouts/blog-post.njk
mamieorine: I have updated YAML frontmatter key from camelcase to underscored

site/_includes/layouts/doc-post.njk
```diff
@@ -16,6 +16,10 @@
     {# The article is wrapped again so that the ToC aligns on its right on desktop #}
     <div class="display-flex justify-content-center width-full">
       <article class="stack measure-long width-full pad-left-400 pad-right-400">
+        {% if new_material %}
```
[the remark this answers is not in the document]
mamieorine: I added `isOutdated` frontmatter key

site/_includes/layouts/blog-post.njk
```diff
@@ -28,6 +28,10 @@
   <div class="display-flex gap-top-300 lg:gap-top-400">
     <div class="display-flex justify-content-center width-full">
       <article class="post stack measure-long width-full pad-left-400 pad-right-400">
+        {% if isOutdated %}
```
matthiasrohmer, line 31: Sorry, nit: better stick with camel_case 🙂

site/_includes/partials/post-banner.njk
```diff
@@ -0,0 +1,9 @@
+<div class="post-banner">
+  <div>
+    {% if new_available_content_url %}
+      {% set material_url = helpers.findByUrl(collections.all, new_available_content_url, locale) %}
+    {% endif %}
+
+    This content is outdated and no longer accurate. {% if material_url %}<a href={{ material_url.url }}>New material available</a>.{% endif %}
```
devnook, line 7: Sorry for not spotting this earlier - this strings need to be passed through i18n filter to make them translatable.

site/_scss/blocks/_post-banner.scss
```diff
@@ -0,0 +1,17 @@
+.post-banner {
+  background-color: #fff;
+  border-radius: 6px;
+  margin-bottom: 30px;
+  position: sticky;
+  top: 24px;
+  z-index: 9;
+
+  > div {
```
matthiasrohmer, line 9: Please use a distinct class name here, like `&__message`

matthiasrohmer (approved):

site/_scss/blocks/_post-banner.scss
matthiasrohmer: Yep, then add `--color-red-lightest-solid` to theme.scss

mamieorine: Dan have confirmed me last week that we can use `color-red-lightest` (with opacity) for the outdated banner.

site/_scss/blocks/_post-banner.scss
```diff
@@ -0,0 +1,23 @@
+.post-banner {
+  background-color: #fff;
+  border-radius: 6px;
+  margin-bottom: 30px;
+  position: sticky;
+  top: 24px;
+  z-index: 9;
+
+  .post-banner__content {
```
matthiasrohmer, line 9: This would produce a selector `.post-banner .post-banner__content`, which is both too specific and too long. Use `&__content` instead which produces the shorter `.post-banner__content` selector.

matthiasrohmer (changes requested):

matthiasrohmer (approved):

matthiasrohmer:
@rachelandrew, forwarding to you for handbook updates 🙂

rachelandrew (approved):
LGTM thank you!

PR #5032, opened 2023-01-25
Update content scripts documentation to reflect API.

### LINKED ISSUE #327: MV3 content scripts docs list unavailable aspects of the scripting API
**Describe the bug**
https://developer.chrome.com/docs/extensions/mv3/content_scripts/ lists a number of fields that don't exist on the scripting API.

https://developer.chrome.com/docs/extensions/mv3/content_scripts/#dynamic-declarative

This describes the dynamic content script API, which is not yet implemented (even on dev channel).  We should not document it.

This also applies to its references elsewhere in the page.

https://developer.chrome.com/docs/extensions/mv3/content_scripts/#programmatic

* The property is `files`, not `file`, and takes an array
* The `arguments` field does not yet exist and is not supported on any channel
* exclude matches and globs ( https://developer.chrome.com/docs/extensions/mv3/content_scripts/#matchAndGlob ) are listed under "Inject programmatically" in the header hierarchy, but are only applicable to statically-defined scripts

(This was a quick pass; there may be other issues)

oliverdunk:
A number of fixes to make sure the content scripts documentation reflects the latest MV3 API changes. This includes using the correct function names, removing some hidden content for features which have now shipped, and removing some examples which are not supported.

Fixes #327

### DIFF
--- a/site/en/docs/extensions/mv3/content_scripts/index.md
+++ b/site/en/docs/extensions/mv3/content_scripts/index.md
@@ -13,7 +13,7 @@ changes to them, and pass information to their parent extension.
 
 ## Understand content script capabilities {: #capabilities }
 
-Content scripts can access Chrome APIs used by their parent extension by exchanging [messages][2]. They can [access extension files][section-files] after declaring them as web-accessible resources.
+Content scripts can access Chrome APIs used by their parent extension by exchanging [messages][2]. They can [access extension files][section-files] after declaring them as [web-accessible resources][34].
 
 Additionally, content scripts can access the following chrome APIs directly:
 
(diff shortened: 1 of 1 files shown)


site/en/docs/extensions/mv3/content_scripts/index.md
```diff
@@ -353,10 +351,10 @@ The content script will be injected into a page if both of the following are tru
 Because the `matches` property is required, `exclude_matches`, `include_globs`, and `exclude_globs`
 can only be used to limit which pages will be affected.
 
-The following extension injects the content script into **https://www.nytimes.com/ health**
-but not into **https://www.nytimes.com/ business** .
+The following extension injects the content script into **https://www.nytimes.com/health**
+but not into **https://www.nytimes.com/business** .
```
jpmedley, line 355: The spaces were likely deliberately so that these do not create unwanted traffic for the New York Times and so that we don't have to make sure they always work. You should probably change them back, but I'm open to suggestion if you have another idea.

jpmedley (changes requested):
One small nit.

jpmedley (approved):

PR #5046, opened 2023-01-26
Fix a few items throwing off build stage.

### LINKED ISSUE #5045: Errors in Build Stage of CI
Our source files are accumulating errors that are causing the Build stage of our CI to not pass. (See the image below.) This means that only commiters with Admin access can merge pull requests. (That's how it appears to me.)

image

jpmedley:
Fixes #5045 

Here are some fixes that move us toward eliminating the errors noted in the issue above. There is more work to be done.


rachelandrew:
@jpmedley this PR currently has a build error itself. Let me know when that is fixed and I'll review it.

PR #5063, opened 2023-01-28
Fix  '}' expected.ts(1005) error

ahnayef:
- There were some unclosed "}" and ")"


AmySteam:
Hey @ahnayef, thank you for your PR! I'd like to give you a bit more information about how the tutorial is divided into steps.

In [step #4](https://developer.chrome.com/docs/extensions/mv3/getstarted/tut-focus-mode/#step-4), we start with the first part of background.js. The ellipsis shows that there's more code to come. We'll finish up with the second part in [step #5](https://developer.chrome.com/docs/extensions/mv3/getstarted/tut-focus-mode/#step-5), making sure everything is closed properly.

You can find the full [background.js](https://github.com/GoogleChrome/chrome-extensions-samples/blob/main/tutorials/focus-mode/background.js) code in the Github sample repository.

Hope this helps! 🙂

PR #5077, opened 2023-01-30
Misc updated to private prefetch proxy post for clarity

tunetheweb:

- Adds location and mime-type of PPP configuration file
- Adds info on how to disable
- Adds links

FYI @KenjiBaheux

### DIFF
--- a/site/en/blog/private-prefetch-proxy/index.md
+++ b/site/en/blog/private-prefetch-proxy/index.md
@@ -7,7 +7,7 @@ description: >
 subhead: >
   Speeding up Largest Contentful Paint (LCP) with cross-site prefetching.
 date: 2022-05-11
-updated: 2022-12-02
+updated: 2023-01-30
 authors:
   - katiehempenius
   - kenjibaheux
@@ -22,9 +22,7 @@ alt: >
 
 Starting with Chrome 103 for Android, Chrome will gradually roll out a private prefetch proxy feature to speed up out-going navigations from Google Search and other participating websites by 30% at the median. This private prefetch proxy feature allows the prefetching of cross-origin content without exposing user information to the destination website until the user navigates.
 
-Read on to learn about [how this feature works](#how),
-[how it can help significantly improve your sites' Largest Contentful Paint (LCP)](#owners),
-or [how referrer websites can help their users](#referrers) achieve their goals by speeding up cross-site navigations.
+Read on to learn about [how this feature works](#how), [how it can help significantly improve your sites' Largest Contentful Paint (LCP)](#owners), or [how referrer websites can help their users](#referrers) achieve their goals by speeding up cross-site navigations.
 
 ## How Private Prefetch Proxy works {: #how }
 
@@ -76,7 +74,7 @@ So, with that in mind, here is what we recommend:
 
 ### Traffic control {: #traffic }
 
-From past experiments, we know that this feature typically results in less than 2% extra requests for main resources (for example HTML documents). That said, if you are the cautious kind, you can use the traffic advice's fraction field to control how much Private Prefetch Proxy should let through. You can start with a small fraction such as 0.3 (that is, 30%), and gradually increase it to 1.0 (that is, 100%).
+From past experiments, we know that this feature typically results in less than 2% extra requests for main resources (for example HTML documents). That said, if you are the cautious kind, you can use the traffic advice's fraction field to control how much Private Prefetch Proxy should let through. You can start with a small fraction such as 0.3 (that is, 30%), and gradually increase it to 1.0 (that is, 100%) by adding the following JSON to in an `/.well-known/traffic-advice` file, which needs to be served with `application/trafficadvice+json` MIME type:
 
 ```json
 [{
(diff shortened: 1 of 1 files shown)


malchata (approved):
LGTM, with the caveat that I'm not an expert on private prefetch proxy.

tunetheweb:
Thanks @malchata , will wait on the API owners (who asked for this!) to review this before merging anyway.

site/en/blog/private-prefetch-proxy/index.md
```diff
@@ -76,7 +74,7 @@ So, with that in mind, here is what we recommend:
 
 ### Traffic control {: #traffic }
 
-From past experiments, we know that this feature typically results in less than 2% extra requests for main resources (for example HTML documents). That said, if you are the cautious kind, you can use the traffic advice's fraction field to control how much Private Prefetch Proxy should let through. You can start with a small fraction such as 0.3 (that is, 30%), and gradually increase it to 1.0 (that is, 100%).
+From past experiments, we know that this feature typically results in less than 2% extra requests for main resources (for example HTML documents). That said, if you are the cautious kind, you can use the traffic advice's fraction field to control how much Private Prefetch Proxy should let through. You can start with a small fraction such as 0.3 (that is, 30%), and gradually increase it to 1.0 (that is, 100%) by adding the following JSON to in an `/.well-known/traffic-advice` file, which needs to be served with `application/trafficadvice+json` MIME type:
```
buettner, line 77: "how much Private Prefetch proxy should let through" -> ".. how much *traffic the* Private Prefetch Proxy should let through."

"JSON to in an " -> "JSON to a ..."

"..served with `application..." -> ".. served with the..."

site/en/blog/private-prefetch-proxy/index.md
```diff
@@ -87,7 +85,16 @@ From past experiments, we know that this feature typically results in less than
 
 The `fraction` field gives you control over the prefetch traffic by specifying how much Private Prefetch Proxy should let through. The value is a float between 0.0 (no prefetch at all) and 1.0 (100% of the prefetch requests get through).
```
buettner, line 86: Up to you, but this sentence is a dupe from above and can probably be simplified by dropping the first sentence.

e.g., "The fraction field is a float between 0.0.."

site/en/blog/private-prefetch-proxy/index.md
```diff
@@ -87,7 +85,16 @@ From past experiments, we know that this feature typically results in less than
...
+It is also possible to disable this completely with the following configuration:
+
+```json
+[{
+  "user_agent": "prefetch-proxy",
+  "disallow": true
+}]
+```
+
+The `/.well-known/traffic-advice` file is not be checked for every private prefetch proxy request and will be cached as per the usual HTTP cache semantics. For more flexibility—for example, a sudden peak of heavy access—you may want to temporarily reject prefetch requests (`Sec-Purpose: prefetch;anonymous-client-ip`) with a 503 status code, and by setting the `Cache-Control: no-store` header on the response. You may also add the [`Retry-After`](https://tools.ietf.org/html/rfc7231#section-6.6.4) header to tell Chrome how long to wait before retrying prefetch requests.
```
buettner, line 97: "The /.well-known/traffic-advice file is fetched by the proxy, not the client, and cached at the proxy per the usual HTTP cache semantics. "

buettner (commented):
A few comments but otherwise LGTM. 

Thanks!

site/en/blog/private-prefetch-proxy/index.md
tunetheweb:
```suggestion
From past experiments, we know that this feature typically results in less than 2% extra requests for main resources (for example HTML documents). That said, if you are the cautious kind, you can use the traffic advice's fraction field to control how much traffic the Private Prefetch Proxy should let through. You can start with a small fraction such as 0.3 (that is, 30%), and gradually increase it to 1.0 (that is, 100%) by adding the following JSON to a `/.well-known/traffic-advice` file, which needs to be served with the `application/trafficadvice+json` MIME type:
```

site/en/blog/private-prefetch-proxy/index.md
tunetheweb:
```suggestion
The `fraction` field is a float between 0.0 (no prefetch at all) and 1.0 (100% of the prefetch requests get through).
```

site/en/blog/private-prefetch-proxy/index.md
tunetheweb:
```suggestion
The `/.well-known/traffic-advice` file is fetched by the proxy, not the client, and cached at the proxy per the usual HTTP cache semantics. For more flexibility—for example, a sudden peak of heavy access—you may want to temporarily reject prefetch requests (`Sec-Purpose: prefetch;anonymous-client-ip`) with a 503 status code, and by setting the `Cache-Control: no-store` header on the response. You may also add the [`Retry-After`](https://tools.ietf.org/html/rfc7231#section-6.6.4) header to tell Chrome how long to wait before retrying prefetch requests.
```

jeremyroman:
lgtm2

PR #5081, opened 2023-01-30
Link to lifetime enforcers for offscreen documents.

### LINKED ISSUE #5019: Document lifetime_enforcer and audio_lifetime_enforcer
Nit: Thanks for adding this, but can we link to https://source.chromium.org/chromium/chromium/src/+/main:extensions/browser/api/offscreen/audio_lifetime_enforcer.h and https://source.chromium.org/chromium/chromium/src/+/main:extensions/browser/api/offscreen/lifetime_enforcer_factories.cc. That at least gives an indication of the code we're referencing and a place to read more :)

_Originally posted by @oliverdunk in https://github.com/GoogleChrome/developer.chrome.com/pull/4765#discussion_r1082493505_

oliverdunk:
Links to the source for the mentioned lifetime enforcers. These links aren't ideal but felt better than giving a name and no indication as to what they refer to.

### DIFF
--- a/site/en/docs/extensions/reference/offscreen/index.md
+++ b/site/en/docs/extensions/reference/offscreen/index.md
@@ -23,7 +23,7 @@ Pages loaded as offscreen documents are handled differently from other types of 
 * An extension can only have one offscreen document open at a time. If the extension is running in split mode with an active incognito profile, both the normal and incognito profiles can each have one offscreen document. 
 
 ## Reasons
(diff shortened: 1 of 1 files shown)


jpmedley:
I'm still having trouble with this. Why don't we just say, something like 'offscreen documents have an unlimited lifetime unless it's doing audio playback. In those cases etc.' 

Also, does video playback work the same way as audio playback? Are there other exceptions to the general rule?

oliverdunk:
@jpmedley I think having links on these names is better than nothing, since it at least allows someone to dig further if they want to. I agree it would be better to document the actual behaviour though. If I remove the "fixes" in the description here so we can have another followup, are you ok with merging this?

jpmedley (approved):
Unless I'm missing something, it does not seem like the observable behavior is so complicated that someone would need to dig into it.

oliverdunk:
@jpmedley, I mostly agree. I just want to do this in the interim since I think a link is better than nothing and very trivial to add 😄

PR #5082, opened 2023-01-30
Use correct code tags

pepelsbey:

- Correct `css` code tags used instead of `js`

---

I’d also suggest either wrapping lists of properties into selectors with brackets (it’s already used in some cases) or removing `css` tags from these examples. They are not fully CSS and can’t be highlighted. For example:

```
color-mix(in srgb, black, white);
color-mix(in srgb-linear, black, white);
color-mix(in lch, black, white);
color-mix(in oklch, black, white);
color-mix(in lab, black, white);
color-mix(in oklab, black, white);
color-mix(in xyz, black, white);
```

### DIFF
--- a/site/en/blog/css-color-mix/index.md
+++ b/site/en/blog/css-color-mix/index.md
@@ -25,7 +25,7 @@ CSS preprocessors or
 channels.
 
 {% Compare 'worse', 'Before with SCSS' %}
-```js
+```css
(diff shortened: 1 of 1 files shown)


pepelsbey:
@argyleink FYI

argyleink (approved):
nice catch, ty!

AaronForinton (approved):

PR #5096, opened 2023-01-31
Clarify service worker timeout

jpmedley:
Fixes #SOME_ISSUE_NUMBER

-
-
-

### DIFF
--- a/site/en/blog/longer-esw-lifetimes/index.md
+++ b/site/en/blog/longer-esw-lifetimes/index.md
@@ -40,7 +40,7 @@ Unfortunately, this behavior did not apply to extension events. Extension events
 
 ## What's changed
 
-As of Chrome 110, all events reset the lidle timer and the idle timeout will not occur if there are pending events. In other words, assuming there are no unexpected interruptions, extension service workers will now typically stay alive as long as they are actively processing events. In addition, calls to extension specific Chrome APIs, such as `chrome.storage.local.get()`, will reset the idle timeout. Be aware that:
+As of Chrome 110, all events reset the idle timer and the idle timeout will not occur if there are pending events. In other words, assuming there are no unexpected interruptions, extension service workers will now typically stay alive as long as they are actively processing events. In addition, calls to extension specific Chrome APIs, such as `chrome.storage.local.get()`, will reset the idle timeout. Be aware that:
 
 * The service worker will terminate after 30 seconds if there are no events.
 * The service worker will terminate after 5 minutes if a single request, such as an event or API call, takes longer than 5 minutes to process (with the exception of a native messaging connection).
--- a/site/en/blog/longer-esw-lifetimes/index.md
+++ b/site/en/blog/longer-esw-lifetimes/index.md
@@ -18,7 +18,7 @@ alt: >
 ---
 
 {% Aside %}
-This post has been updated since we originally published it in January. We've added text indicating that calls to extension APIs reset the service worker's timeout clock, and clarified exactly when service worker timeouts will ocurr.
+This post has been updated since we originally published it in January. We've added text indicating that calls to extension APIs reset the service worker's timeout clock, and clarified exactly when service worker timeouts will occur.
 {% endAside %}
 
 Starting in Chrome 110 (in beta as of February 7, 2023), extension service workers stay alive as long as they're receiving events. This corrects a timing problem in the previous implementation of extension service workers. It was possible for timeouts to occur when new events were in the event queue and for the timeouts to truncate asynchronous work. This improvement removes the hard five-minute maximum lifetime for extension service workers. 
--- a/site/en/blog/longer-esw-lifetimes/index.md
+++ b/site/en/blog/longer-esw-lifetimes/index.md
@@ -36,7 +36,7 @@ Unfortunately, this behavior did not apply to extension events. Extension events
 
 ## What's changed
 
-As of Chrome 110, all events reset the lidle timer and the idle timeout will not occur if there are pending events. In other words, assuming there are no unexpected interruptions, extension service workers will now typically stay alive as long as they are actively processing events. Be aware that:
+As of Chrome 110, all events reset the lidle timer and the idle timeout will not occur if there are pending events. In other words, assuming there are no unexpected interruptions, extension service workers will now typically stay alive as long as they are actively processing events. In addition, calls to extension specific Chrome APIs, such as `chrome.storage.local.get(...)`, will reset the idle timeout. Be aware that:
 
 * The service worker will terminate after 30 seconds if there are no events.
 * The service worker will terminate after 5 minutes if a single request, such as an event or API call, takes longer than 5 minutes to process (with the exception of a native messaging connection).
--- a/site/en/blog/longer-esw-lifetimes/index.md
+++ b/site/en/blog/longer-esw-lifetimes/index.md
@@ -47,7 +47,7 @@ As of Chrome 110, all events reset the idle timer and the idle timeout will not 
 
 Some APIs like native messaging provide a strong keep-alive which cancel both of these timers.
 
-We are till working to ensure that extension service workers are terminated when possible, without shutting down long-running work. A resource-concious extension service workers should always yield when possible. Additionally, extensions should prepare for unexpected termination by persisting state. This ensures against unpredictable events like the browser being forcefully closed by the user.
+We are still working to ensure that extension service workers are terminated when possible, without shutting down long-running work. Resource-concious extension service workers should always yield when possible. Additionally, extensions should prepare for unexpected termination by persisting state. This guards against unpredictable events like the browser being forcefully closed by the user.
 
 <p>Photo by <a href="https://unsplash.com/@pguerreiro?utm_source=unsplash&utm_medium=referral&utm_content=creditCopyText">Paula Guerreiro</a> on <a href="https://unsplash.com/photos/W2atfIRHDIk?utm_source=unsplash&utm_medium=referral&utm_content=creditCopyText">Unsplash</a></p>
   
--- a/site/en/blog/longer-esw-lifetimes/index.md
+++ b/site/en/blog/longer-esw-lifetimes/index.md
@@ -45,7 +45,7 @@ As of Chrome 110, all events reset the idle timer and the idle timeout will not 
 * The service worker will terminate after 30 seconds if there are no events.
 * The service worker will terminate after 5 minutes if a single request, such as an event or API call, takes longer than 5 minutes to process (with the exception of a native messaging connection).
 
-You are still strongly encouraged to design your extensions to be event-based and guard against unexpected termination. Not only should you refrain from keeping your extensions alive indefinitely (which we consider an anti-pattern), you should test your extensions to ensure that they're not doing this unintentionally. Make sure your extensions yield when possible and prepare for termination by persisting extension states.
+We are till working to ensure that extension service workers are terminated when possible, without shutting down long-running work. A resource-concious extension service workers should always yield when possible. Additionally, extensions should prepare for unexpected termination by persisting state. This ensures against unpredictable events like the browser being forcefully closed by the user.
 
 <p>Photo by <a href="https://unsplash.com/@pguerreiro?utm_source=unsplash&utm_medium=referral&utm_content=creditCopyText">Paula Guerreiro</a> on <a href="https://unsplash.com/photos/W2atfIRHDIk?utm_source=unsplash&utm_medium=referral&utm_content=creditCopyText">Unsplash</a></p>
   
--- a/site/en/blog/longer-esw-lifetimes/index.md
+++ b/site/en/blog/longer-esw-lifetimes/index.md
@@ -6,7 +6,7 @@ description: >
 subhead: >
   Extension service workers can now stay alive as long as they're receiving events. This increases the reliability of extension services workers, but has a pitfall you should avoid.
 date: 2023-01-27
-updated: 2023-04-14
+updated: 2023-05-08
 authors:
   - joemedley
 tags:
@@ -18,7 +18,7 @@ alt: >
 ---
 
 {% Aside %}
-This post has been updated since we originally published it in January. TBD -summary of changes.
+This post has been updated since we originally published it in January. We've added text indicating that calls to extension APIs reset the service worker's timeout clock, and clarified exactly when service worker timeouts will ocurr.
 {% endAside %}
 
 Starting in Chrome 110 (in beta as of February 7, 2023), extension service workers stay alive as long as they're receiving events. This corrects a timing problem in the previous implementation of extension service workers. It was possible for timeouts to occur when new events were in the event queue and for the timeouts to truncate asynchronous work. This improvement removes the hard five-minute maximum lifetime for extension service workers. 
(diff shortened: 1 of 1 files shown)


site/en/blog/longer-esw-lifetimes/index.md
```diff
@@ -37,7 +38,7 @@ Unfortunately, this behavior did not apply to extension events. Extension events
 
 As of Chrome 110, all events reset the lidle timer and the idle timeout will not occur if there are pending events. In other words, assuming there are no unexpected interruptions, extension service workers will now stay alive as long as they are actively processing events.
 
-While this change removes the five minute limit on extension service worker lifetimes, you are still strongly encouraged to design your extensions to be event-based and guard against unexpected termination. Not only should you refrain from keeping your extensions alive indefinitely (which we consider an anti-pattern), you should test your extensions to ensure that they're not doing this unintentionally.
+Because the service worker will still terminate after five minutes if there are no events, you are still strongly encouraged to design your extensions to be event-based and guard against unexpected termination. Not only should you refrain from keeping your extensions alive indefinitely (which we consider an anti-pattern), you should test your extensions to ensure that they're not doing this unintentionally.
```
rdcronin, line 41: This still isn't quite right.
- The service worker will terminate after 30 seconds if there are no events
- The service worker will terminate after 5 minutes if a single request (event or API call) takes longer than 5 minutes to process

rdcronin (commented):
Thanks, Joe!  I think we still need to tweak this a bit.

site/en/blog/longer-esw-lifetimes/index.md
jpmedley: PTAL

site/en/blog/longer-esw-lifetimes/index.md
```diff
@@ -35,9 +36,12 @@ Unfortunately, this behavior did not apply to extension events. Extension events
...
 
-As of Chrome 110, all events reset the lidle timer and the idle timeout will not occur if there are pending events. In other words, assuming there are no unexpected interruptions, extension service workers will now stay alive as long as they are actively processing events.
+As of Chrome 110, all events reset the lidle timer and the idle timeout will not occur if there are pending events. In other words, assuming there are no unexpected interruptions, extension service workers will now stay alive as long as they are actively processing events. Be aware that:
 
-While this change removes the five minute limit on extension service worker lifetimes, you are still strongly encouraged to design your extensions to be event-based and guard against unexpected termination. Not only should you refrain from keeping your extensions alive indefinitely (which we consider an anti-pattern), you should test your extensions to ensure that they're not doing this unintentionally.
+* The service worker will terminate after 30 seconds if there are no events.
+* The service worker will terminate after 5 minutes if a single request (event or API call) takes longer than 5 minutes to process.
```
rdcronin, line 42:
```suggestion
* The service worker will terminate after 5 minutes if a single request, such as an event or API call, takes longer than 5 minutes to process (with the exception of a native messaging connection).
```

site/en/blog/longer-esw-lifetimes/index.md
```diff
@@ -35,9 +36,12 @@ Unfortunately, this behavior did not apply to extension events. Extension events
 
 ## What's changed
 
-As of Chrome 110, all events reset the lidle timer and the idle timeout will not occur if there are pending events. In other words, assuming there are no unexpected interruptions, extension service workers will now stay alive as long as they are actively processing events.
+As of Chrome 110, all events reset the lidle timer and the idle timeout will not occur if there are pending events. In other words, assuming there are no unexpected interruptions, extension service workers will now stay alive as long as they are actively processing events. Be aware that:
```
rdcronin, line 39:
```suggestion
As of Chrome 110, all events reset the lidle timer and the idle timeout will not occur if there are pending events. In other words, assuming there are no unexpected interruptions, extension service workers will now typically stay alive as long as they are actively processing events. Be aware that:
```

site/en/blog/longer-esw-lifetimes/index.md
```diff
@@ -35,9 +36,12 @@ Unfortunately, this behavior did not apply to extension events. Extension events
```
rdcronin, line 36: Github won't let me comment on line 33, but:

> New service worker events received before the idle timer or long running task timer expire would reset the timers and extend the service worker's lifetime.

This should read "... would reset the idle timer and..." vs "reset the timers and".  The long-running task timer is not reset.

site/en/blog/longer-esw-lifetimes/index.md
```diff
@@ -35,9 +36,12 @@ Unfortunately, this behavior did not apply to extension events. Extension events
...
+* The service worker will terminate after 30 seconds if there are no events.
+* The service worker will terminate after 5 minutes if a single request (event or API call) takes longer than 5 minutes to process.
+
+You are still strongly encouraged to design your extensions to be event-based and guard against unexpected termination. Not only should you refrain from keeping your extensions alive indefinitely (which we consider an anti-pattern), you should test your extensions to ensure that they're not doing this unintentionally.
 
 In a future release, we may change how specific events are processed to prevent keeping service workers alive indefinitely, but without shutting down long-running work. In the meantime, make sure your extensions yield when possible and prepare for termination by persisting extension states.
```
rdcronin, line 46: This line doesn't make sense to me.

> to prevent keeping service workers alive indefinitely

We don't mind this as long as it's legitimately useful work.

> but without shutting down long-running work

This seems directly contradictory to the long-running task timer, which explicitly _will_ shut down long-running work.

If you want to keep this sentence, maybe:

"Future releases of Chromium may prevent artificially extending service worker lifetimes."

That said, I we could even just remove this whole sentence, and just append the second to the previous paragraph:

"... you should test your extensions to ensure that they're not doing this unintentionally and make sure your extensions yield whenever possible..."

rdcronin (changes requested):

site/en/blog/longer-esw-lifetimes/index.md
jpmedley: Done.

site/en/blog/longer-esw-lifetimes/index.md
jpmedley: I took the second approach. The existing language was Simeon's. Not only do I prefer to stay away from remarks about the future, such promises are also contrary to Google's style guide.

jpmedley:
@rdcronin, back to you. I hope I got it this time.

jpmedley:
@rdcronin Can you please give this a final pass. Thanks.

site/en/blog/longer-esw-lifetimes/index.md
```diff
@@ -35,11 +36,12 @@ Unfortunately, this behavior did not apply to extension events. Extension events
...
+As of Chrome 110, all events reset the lidle timer and the idle timeout will not occur if there are pending events. In other words, assuming there are no unexpected interruptions, extension service workers will now typically stay alive as long as they are actively processing events. Be aware that:
 
-While this change removes the five minute limit on extension service worker lifetimes, you are still strongly encouraged to design your extensions to be event-based and guard against unexpected termination. Not only should you refrain from keeping your extensions alive indefinitely (which we consider an anti-pattern), you should test your extensions to ensure that they're not doing this unintentionally.
+* The service worker will terminate after 30 seconds if there are no events.
+* The service worker will terminate after 5 minutes if a single request, such as an event or API call, takes longer than 5 minutes to process (with the exception of a native messaging connection).
```
sebastianbenz, line 42: I assume this is only the case for pending async calls, correct? Or will synchronous operations be terminated as well?

//cc @rdcronin

site/en/blog/longer-esw-lifetimes/index.md
```diff
@@ -35,11 +36,12 @@ Unfortunately, this behavior did not apply to extension events. Extension events
...
 
-In a future release, we may change how specific events are processed to prevent keeping service workers alive indefinitely, but without shutting down long-running work. In the meantime, make sure your extensions yield when possible and prepare for termination by persisting extension states. 
+You are still strongly encouraged to design your extensions to be event-based and guard against unexpected termination. Not only should you refrain from keeping your extensions alive indefinitely (which we consider an anti-pattern), you should test your extensions to ensure that they're not doing this unintentionally. Make sure your extensions yield when possible and prepare for termination by persisting extension states.
```
sebastianbenz, line 44: This does not quite fit the rest of the flow: "You are still strongly encouraged to design your extensions to be event-based and guard against unexpected termination."

Reading the post I see nothing that would indicate that sw's are no longer event-based (as we don't explicitly mention hacking the lifecycle).

site/en/blog/longer-esw-lifetimes/index.md
```diff
@@ -35,11 +36,12 @@ Unfortunately, this behavior did not apply to extension events. Extension events
 
 ## What's changed
 
-As of Chrome 110, all events reset the lidle timer and the idle timeout will not occur if there are pending events. In other words, assuming there are no unexpected interruptions, extension service workers will now stay alive as long as they are actively processing events.
+As of Chrome 110, all events reset the lidle timer and the idle timeout will not occur if there are pending events. In other words, assuming there are no unexpected interruptions, extension service workers will now typically stay alive as long as they are actively processing events. Be aware that:
```
sebastianbenz, line 39: I think it'd be great to elaborate a bit more on this:

- why are we're making this change?
- what was the previous behavior? (this helps devs gauge how this might impact the behavior of their extension)

Another thing: can we either remove `typically` or make it more precise? left on its own it is a bit confusing (at least I stumbled over it).

sebastianbenz (commented):
Looking good! Glad to see this go out. Left a few comments.

site/en/blog/longer-esw-lifetimes/index.md
jpmedley: >why are we're making this change?

Because premature termination was causing errors in people's code. At the time we figured that if somebody wasn't  having that problem they wouldn't be reading this article. Frankly, I'd like to err on the side of getting this out.

>what was the previous behavior?

It's described in the background.

site/en/blog/longer-esw-lifetimes/index.md
jpmedley: I don't know. Maybe Devlin does.

site/en/blog/longer-esw-lifetimes/index.md
jpmedley: I remember no. Simeon strongly insisted on this line. We can get rid of it. @oliverdunk I'd be curious to know, if you have any opinions on this line.

site/en/blog/longer-esw-lifetimes/index.md
```diff
@@ -16,6 +17,10 @@ alt: >
   A picture of an hourglass.
 ---
 
+{% Aside %}
+This post has been updated since we originally published it in January. TBD -summary of changes.
```
jpmedley, line 21: @sebastianbenz I'll fill this out when the content is settled.

site/en/blog/longer-esw-lifetimes/index.md
oliverdunk: I think the two important things to communicate are:

- We may change this behaviour to be stricter in the future (though hopefully in a way that does not impact genuine extensions).
- Your extension may terminate unexpectedly even with these changes (for example, if the browser crashes).

How about something like:

> In a future release, we may continue to iterate on this behaviour to ensure extension service workers are terminated when possible, without shutting down long-running work. As part of building a resource conscious extension, extension service workers should always yield when possible. Additionally, extensions should prepare for unexpected termination by persisting state. This makes sure they are robust against unpredictable events like the browser being forcefully closed by the user.

I think that's a bit wordy but better in some sense?

jpmedley: I just submitted a change with a version of Oliver's suggestion.

site/en/blog/longer-esw-lifetimes/index.md
jpmedley: New wording from Oliver.

jpmedley: Oliver didn't like the new wording and he makes some compelling arguments:
* >It won't terminate after 30 seconds if you make certain API calls for example. So it would need to be "The service worker will terminate after 30 seconds if there are no events or API calls to extension APIs and there isn't a native messaging connection, regardless of whether you're running synchronous or asynchronous code." I realise we touch on that elsewhere in the article but I think we should make sure if people quote a specific bullet point it isn't wrong.
* >I feel like synchronous and asynchronous was sort of implied with how we had it before, and calling it out actually adds confusion because then it makes me stop to think about what those are.

oliverdunk: I'm not sure what the latest wording is but I think something like this would be fine:

> * The service worker will terminate after 30 seconds of inactivity (receiving an event or calling an extension API resets this timer).
> * The service worker will terminate after 5 minutes if a single request, such as an event or API call, takes longer than 5 minutes to process
>
> Some APIs like native messaging provide a strong keep-alive which cancel both of these timers.

And maybe change "be aware that" to "specifically" or something? Since it's not a bad thing.

sebastianbenz: +1 to Oliver's suggestions

jpmedley:
@sebastianbenz PTAL

site/en/blog/longer-esw-lifetimes/index.md
```diff
@@ -16,6 +17,10 @@ alt: >
   A picture of an hourglass.
 ---
 
+{% Aside %}
+This post has been updated since we originally published it in January. We've added text indicating that calls to extension APIs reset the service worker's timeout clock, and clarified exactly when service worker timeouts will ocurr.
```
oliverdunk, line 21: Hopefully this is the US spelling too:

```suggestion
This post has been updated since we originally published it in January. We've added text indicating that calls to extension APIs reset the service worker's timeout clock, and clarified exactly when service worker timeouts will occur.
```

site/en/blog/longer-esw-lifetimes/index.md
```diff
@@ -35,11 +40,14 @@ Unfortunately, this behavior did not apply to extension events. Extension events
...
 
-While this change removes the five minute limit on extension service worker lifetimes, you are still strongly encouraged to design your extensions to be event-based and guard against unexpected termination. Not only should you refrain from keeping your extensions alive indefinitely (which we consider an anti-pattern), you should test your extensions to ensure that they're not doing this unintentionally.
+Some APIs like native messaging provide a strong keep-alive which cancel both of these timers.
 
-In a future release, we may change how specific events are processed to prevent keeping service workers alive indefinitely, but without shutting down long-running work. In the meantime, make sure your extensions yield when possible and prepare for termination by persisting extension states. 
+We are till working to ensure that extension service workers are terminated when possible, without shutting down long-running work. A resource-concious extension service workers should always yield when possible. Additionally, extensions should prepare for unexpected termination by persisting state. This ensures against unpredictable events like the browser being forcefully closed by the user.
```
oliverdunk, line 50:
```suggestion
We are still working to ensure that extension service workers are terminated when possible, without shutting down long-running work. Resource-concious extension service workers should always yield when possible. Additionally, extensions should prepare for unexpected termination by persisting state. This guards against unpredictable events like the browser being forcefully closed by the user.
```

oliverdunk (approved):
Left a couple of nits but otherwise this looks good to me (would still love for Sebastian to approve as well) 😄

sebastianbenz (approved):
LGTM once Oliver's feedback has been addressed.

jpmedley:
I'm getting build issues I can't resolve. Opening a new PR.

PR #5111, opened 2023-02-01
Update Lighthouse `is-crawlable` audit doc

adamraine:
Closes https://github.com/GoogleChrome/lighthouse/issues/5329

Lighthouse *will* now look for crawler-specific directives and emit a warning if they are found: https://github.com/GoogleChrome/lighthouse/pull/14550

cc @connorjclark

### DIFF
--- a/site/en/docs/lighthouse/seo/is-crawlable/index.md
+++ b/site/en/docs/lighthouse/seo/is-crawlable/index.md
@@ -39,11 +39,12 @@ X-Robots-Tag: noindex
 You might also have `<meta>` elements that block specific crawlers, such as:
 
 ```html
-<meta name="AdsBot-Google" content="noindex"/>
+<meta name="Googlebot" content="noindex"/>
 ```
 
-Lighthouse doesn't check for crawler-specific directives like this, but they can
-still make your page harder to discover, so use them with caution.
+Lighthouse doesn't fail the audit for crawler-specific directives like this, but they can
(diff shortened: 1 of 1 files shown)


site/en/docs/lighthouse/seo/is-crawlable/index.md
```diff
@@ -39,11 +39,12 @@ X-Robots-Tag: noindex
 You might also have `<meta>` elements that block specific crawlers, such as:
 
 ```html
-<meta name="AdsBot-Google" content="noindex"/>
+<meta name="Googlebot" content="noindex"/>
 ```
 
-Lighthouse doesn't check for crawler-specific directives like this, but they can
-still make your page harder to discover, so use them with caution.
+Lighthouse doesn't fail the audit for crawler-specific directives like this, but they can
```
connorjclark, line 45: Shoudl we qualify this as "as of 10.0"? Must account for chrome stable running older LH longer than other channels; and to a lesser extent, historical LH versions.

connorjclark (approved):
+1, but merging should hold for our release.

site/en/docs/lighthouse/seo/is-crawlable/index.md
adamraine: I don't think we need a 10.0 disclaimer for something as minor as this

brendankenny (approved):

malchata (approved):
LGTM.

PR #5120, opened 2023-02-02
Test: sticky pull request comment on forks

matthiasrohmer:


matthiasrohmer:
Nope that doesn't work 🙂  would have been too easy...

PR #5128, opened 2023-02-02
Fix various staging flaws

matthiasrohmer:


.cloudbuild/stage.yaml
```diff
@@ -1,43 +1,74 @@
...
+  - id: 'build'
+    name: node:18
     env:
-      - 'BRANCH_NAME=$BRANCH_NAME'
-      - 'PR_NUMBER=$_PR_NUMBER'
-    secretEnv: ['GITHUB_TOKEN']
-    args: ['run', 'stage:github', 'determineDeploymentType']
+      - 'CI=true'
+      - 'NODE_OPTIONS="--max_old_space_size=8192"'
+      - 'ELEVENTY_IGNORE_NACL=true'
```
devnook, line 22: We didn't ignore this previously, did we?

gulp-tasks/stageGitHub/lib/updateStickyComment.js
```diff
@@ -0,0 +1,47 @@
...
+
+  if (!comment) {
+    await requestGitHubApi('POST', `issues/${prNumber}/comments`, {body});
+  } else {
+    await requestGitHubApi('PATCH', `issues/comments/${comment.id}`, {body});
+  }
+}
+
+(async () => {
+  await updateStickyComment(5128, 'Hello world, sticky?');
```
devnook, line 44: You'll freak out some folks with this:)

.cloudbuild/stage.yaml
devnook: also shoukld not influence prod build i think, and you run "npm run production" later on in gulp?

devnook (approved):
Left some small questions

gulp-tasks/stageGitHub/lib/updateStickyComment.js
matthiasrohmer: Oh damn 😅  at least it was hard coded to one of my PRs.

.cloudbuild/stage.yaml
matthiasrohmer: It does by setting `CI=true` (which is weird, but here we are). That's how it currently is for Netlify, and it takes ~1m from the build (less files to build, less files to upload).

I am a little bit torn about it: I think it's good to do this, to not have people (rightfully) complain that the new previews are slower to build (which they are not measured no a per page basis) on the other hand I actually liked having **full** builds. What's your stance?

devnook: we can ignore nacl for PRs because they are fully automated content, which is unlikely to get influenced by random PR.

matthiasrohmer:
Keeping this for Monday as probably a bunch of PRs need a merge with main to avoid confusion by "breaking" checks.

PR #5141, opened 2023-02-03
refactor: rename "gotchas" aside type to "important"

### LINKED ISSUE #3380: Update asides to match web.dev
I notice that the dcc and web.dev asides are not parallel, and should be.

- web.dev has an [important aside](https://web.dev/handbook/web-dev-components/#important-asides), where on dcc it's a [gotcha aside](https://developer.chrome.com/docs/handbook/components/#gotchas-asides). I suggest we use "Important" as it's more inclusive of other uses.
- The dcc codelab aside should also have the header changed to match web.dev, as [Try it](https://web.dev/handbook/web-dev-components/#codelab-asides). Updating this could also close https://github.com/GoogleChrome/developer.chrome.com/issues/2170, if we are opposed to a separate additional example aside. 

Further, web.dev also has a [Celebration](https://developer.chrome.com/docs/handbook/components/#gotchas-asides)

Blazzike:
Fixes #3380

- Add "important" Spanish translation
- Remove "gotchas" translations
- Use `.aside--important` class where `.aside--gotchas` was previously used
- Rewrite previous uses of "gotchas" type to "important" without modifying affected articles
- Update handbook gotchas demo to an important aside demo

### DIFF
--- a/site/_data/i18n/common.yml
+++ b/site/_data/i18n/common.yml
@@ -78,9 +78,23 @@ success:
   en: 'Success'
   es: 'Éxito'
 
+gotchas:
+  en: 'Gotchas'
+  es: 'Gotchas'
+  ja: '注意事項'
+  ko: '주의사항'
+  pt: 'Gotchas'
+  ru: 'Особенности'
+  zh: '注意事项'
+
 important:
   en: 'Important'
   es: 'Importante'
+  ja: '重要'
+  ko: '중요한'
+  pt: 'Importante'
+  ru: 'Важный'
+  zh: '重要的'
 
 key-term:
   en: 'Key Term'
--- a/site/_data/i18n/common.yml
+++ b/site/_data/i18n/common.yml
@@ -78,12 +78,9 @@ success:
   en: 'Success'
   es: 'Éxito'
 
-gotchas:
-  en: 'Gotchas'
-  es: 'Gotchas'
-
 important:
   en: 'Important'
+  es: 'Importante'
 
 key-term:
   en: 'Key Term'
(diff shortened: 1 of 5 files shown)


site/_shortcodes/Aside.js
```diff
@@ -37,6 +36,12 @@ function Aside(content, type = 'note') {
   // @ts-ignore: `this` has type of `any`
   const locale = this.ctx.locale || defaultLocale;
   let text;
+
+  // implement backwards compatibility for old "gotchas" type, rewrite to "important" type
+  if (type === 'gotchas') {
+    type = 'important';
+  }
```
matthiasrohmer, line 43: Good approach! Nevertheless a follow-up PR actually updating the using articles and removing this compat condition would be great to keep things tidy 🙂

site/_data/i18n/common.yml
```diff
@@ -78,12 +78,9 @@ success:
   en: 'Success'
   es: 'Éxito'
 
-gotchas:
-  en: 'Gotchas'
-  es: 'Gotchas'
-
 important:
   en: 'Important'
+  es: 'Importante'
```
matthiasrohmer, line 83: Single words are safe to be translated by Google Translate.

```suggestion
es: Importante
ja: 重要
ko: 중요한
pt: Importante
ru: Важный
zh: 重要的
```

site/_shortcodes/Aside.js
Blazzike: hey @matthiasrohmer cheers, happy to do that in this one if you like!

matthiasrohmer (approved):

PR #5149, opened 2023-02-03
feat: expanded landing page macros

### LINKED ISSUE #3234: Expanded landing page macros
The new landing page designs for Workbox are terrific, and I'd like to use those macros for Privacy Sandbox! However, we've got some additional needs around ordering and style that weren't captured in the Figma process.

I've worked out some possible examples in https://github.com/GoogleChrome/developer.chrome.com/pull/3233

**Tl;dr** No one comes to the Privacy Sandbox docs to see everything. How can we make this a more pleasant and organized to their needs experience?

I'm hoping to achieve:

1. **High-level, use case based navigation**. The Privacy Sandbox navigation is unique among dcc doc sets, as a reader will most likely care about some sets of docs and completely disregard the rest. There is a set of introductory material (including a glossary and FAQs), and the rest of the materials are grouped by use case
   - Strengthen cross-site privacy boundaries
   - Show relevant content and ads,
   - Measure digital ads
   - Prevent covert tracking
   - Fight spam and fraud on the web

This matches the expectations of [privacysandbox.com](https://privacysandbox.com/intl/en_us/open-web/#how-works-on-web-hero), which is a point of entry for many of our users. These use cases will soon have expanded documentation which gives more context ([go/k-visions](http://goto.google.com/k-visions)).

### Specific design module needs

1. **Introducing use cases**. It would be useful, after the introduction materials, to list the use cases. Ideally, this module would not take too much visual real estate, but would serve to point users towards the list of relevant materials. This could be carried through iconography as jump links.
2. **“Learn more” about the headlines**. I'm writing up documentation about the use cases, the history and Privacy Sandbox proposals, which should have high prominence in a section.
3. **Highlighting a short list of important docs in a larger list**. Some APIs have a longer list of documentation, such as Attribution Reporting. It would be useful to highlight the top-most important content, which is relevant to all of our docs visitors (technical practitioners who implement the APIS, as well as leadership and regulators). But, we also want to make it easy for technical practitioners to skip what they already know (overview materials) and get right to the advanced materials. 
In the context of current design modules:
Top-4 style landing section, with an additional links and blog posts section
4. **Single use case with multiple ordered sets**. For example, the “Show relevant content use case” has two active, in-progress API proposals with several docs: Topics and FLEDGE. Both have multiple docs which should be read in order. And while these could be there own boxes of content, they both fit with the same use case.

simgobb:
Fixes #3234 

- removed some vertical padding from the main hero banner
- created "Explore use cases component"
- added "Learn More" option to use case blocks
- added "More Documents" option to use case blocks
- added option for "Important" documents
- added samples of the above in the handbook


johnrich85 (approved):

site/_data/i18n/landing.yml
```diff
@@ -0,0 +1,5 @@
+landing_sections:
```
matthiasrohmer, line 1: Nit: as the filename already sets the scope, that we are in landing I think we can simply use `sections` as a key here.

matthiasrohmer (approved):
- Depending on the icons used, links can jump.
image

This looks about right. I am not particularly a fan of numbering arguments (moreDocuments1, moreDocuments2) but I see this follow the already existing scheme for useCases.

While reviewing this I stumbled over some things that just don't look right on the landing pages in general, which I'll file an issue for.

matthiasrohmer (changes requested):
Created https://github.com/GoogleChrome/developer.chrome.com/issues/5197 as a follow-up which we can take care of next sprint.

site/_data/i18n/landing.yml
simgobb: Yep, my bad, moved it from common.yml after review and didn't change the names

matthiasrohmer (approved):
Thanks, @simgobb!

PR #5152, opened 2023-02-04
DevTools 111: Breakpoints UX redesign

sofiayem:
Documented new UX for breakpoints.
Added the missing logpoints section.

Publish on March 7.


site/en/docs/devtools/javascript/breakpoints/index.md
```diff
@@ -157,27 +211,28 @@ all mouse events.
...
+{% Img src="image/NJdAV9UgKuN8AhoaPBquL7giZQo1/aFFgOPi3outuMb32WO3k.png", alt="Creating an event listener breakpoint.", width="800", height="726" %}
 
-**Figure 7**: Creating an event listener breakpoint for `deviceorientation`
+This example shows how to create an event listener breakpoint for `deviceorientation`.
 
 ## Exception breakpoints {: #exceptions }
 
 Use exception breakpoints when you want to pause on the line of code that's throwing a caught or
-uncaught exception.
+uncaught exception. You can pause on both such exceptions independently.
```
ktran, line 221: This is actually currently only true for non-node debug sessions. The reason is that for node LTS, which still has an outdated version of v8, we cannot yet support that. We're planning to make these independent switches available as soon as node LTE will switch versions, and have an updated v8. See crbug: https://crbug.com/1382762

The RFC's FAQ shows how it looks like for Node applications right now: https://github.com/ChromeDevTools/rfcs/discussions/3

Maybe this is worth mentioning?

ktran (approved):
Looks great, thanks Sofia! Just one question.

site/en/docs/devtools/javascript/breakpoints/index.md
sofiayem: Added a note about it. Thanks!

ktran:
Looks great, thanks :)!

rachelandrew (approved):

site/en/docs/devtools/javascript/breakpoints/index.md
```diff
@@ -59,37 +60,90 @@ investigate, but you want to pause only when some other condition is true.
...
+1.  Open the **Sources** tab.
 2.  Open the file containing the line of code you want to break on.
 3.  Go to the line of code.
 4.  To the left of the line of code is the line number column. Right-click it.
 5.  Select **Add conditional breakpoint**. A dialog displays underneath the line of code.
 6.  Enter your condition in the dialog.
-7.  Press Enter to activate the breakpoint. An orange icon appears on top of the line number column.
+7.  Press <kbd>Enter</kbd> to activate the breakpoint. An orange icon with a question mark appears on top of the line number column.
+
+{% Img src="image/NJdAV9UgKuN8AhoaPBquL7giZQo1/wbb7VQOAG1GOHvBqqjdt.png", alt="A conditional line-of-code breakpoint.", width="800", height="579" %}
```
jecfish, line 71: The walkthrough should include an example of condition or usecase to help beginner understand better. 
Probably add some example in step 6, or explain your screenshot on why you want to do that in real-life.

Currently the code example only show in image, which is not accessible.

site/en/docs/devtools/javascript/breakpoints/index.md
```diff
@@ -59,37 +60,90 @@ investigate, but you want to pause only when some other condition is true.
...
+To set a logpoint:
+
+1.  Open the **Sources** tab.
+1.  Open the file containing the line of code you want to break on.
+1.  Go to the line of code.
+1.  To the left of the line of code is the line number column. Right-click it.
+1.  Select **Add logpoint**. A dialog displays underneath the line of code.
+1.  Enter your log message in the dialog.
+    
+    You can separate strings, variable, and objects with commas. Wrap strings in quotes (`" "`), objects in curly brackets (`{ }`), and type variable names as they are.
```
jecfish, line 88: > Wrap strings in quotes (`" "`), objects in curly brackets (`{ }`), and type variable names as they are

There are a few thoughts are:

1. Erm... I afraid this might give user wrong impression: this line indicate they must wrap object in curly brackets which is not the case, and string can be wrap in single quotes as well. 
2. I am wondering if we can replace this to something "accept javascript expression"
3. User can also log an expression / condition. For example, here is a few valid logs `"test " + variableName, "test".length > 3, "test".toUpperCase()"`

Given `variableName = 'haha'` This example will log `testhaha, true, TEST` in the console.

I would like to see us expand the section here to cover more use-cases.

jecfish: @ktran WDYT?

site/en/docs/devtools/javascript/breakpoints/index.md
```diff
@@ -59,37 +60,90 @@ investigate, but you want to pause only when some other condition is true.
...
+{% Video src="video/NJdAV9UgKuN8AhoaPBquL7giZQo1/UslK4pKSMUfuzuzE29gx.mp4", muted="true", controls="true", class="screenshot" %}
 
-**Figure 2**: A conditional line-of-code breakpoint set on line **32**
+This video shows how to collapse groups and disable or enable breakpoints one by one or by groups. When you disable a breakpoint, the **Sources** panel makes its marker next to the line number transparent.
 
-### Manage line-of-code breakpoints {: #manage-loc }
+Groups have context menus. Right-click a group and choose:
 
-Use the **Breakpoints** pane to disable or remove line-of-code breakpoints from a single location.
+{% Img src="image/NJdAV9UgKuN8AhoaPBquL7giZQo1/SdWkkrJ10XJ7IrsSluZv.png", alt="The context menu of a group.", width="800", height="749" %}
```
jecfish, line 116: highlight breakpoint pane in the image, making it clear to user.

site/en/docs/devtools/javascript/breakpoints/index.md
```diff
@@ -104,9 +158,9 @@ To set a DOM change breakpoint:
 4.  Hover over **Break on** then select **Subtree modifications**, **Attribute modifications**, or
     **Node removal**.
 
-{% Img src="image/admin/PlPfY25q2BTOrc69IVDj.png", alt="The context menu for creating a DOM change breakpoint.", width="800", height="577" %}
+{% Img src="image/NJdAV9UgKuN8AhoaPBquL7giZQo1/XsXVifCHA03yWaSJHeUR.png", alt="The context menu for creating a DOM change breakpoint.", width="800", height="756" %}
 
-**Figure 5**: The context menu for creating a DOM change breakpoint
+This example shows the context menu for creating a DOM change breakpoint.
```
jecfish, line 163: Not related to breakpoint pane changes, but this session missing the guidance on where to manage this breakpoint. - Elements panel > DOM breakpoints pane, Sources panel > DOM Breakpoints side pane.

site/en/docs/devtools/javascript/breakpoints/index.md
```diff
@@ -140,10 +194,10 @@ To set an XHR breakpoint:
     anywhere in an XHR's request URL.
 5.  Press Enter to confirm.
 
-{% Img src="image/admin/BQXfCgOqQp7usvLOSySx.png", alt="Creating an XHR breakpoint.", width="800", height="698" %}
+{% Img src="image/NJdAV9UgKuN8AhoaPBquL7giZQo1/AAL4i6FKWOaIEdreyLjE.png", alt="Creating an XHR breakpoint.", width="800", height="756" %}
```
jecfish, line 197: Not about this image but comments for the walkthrough above. This section title already mention "XHR/Fetch breakpoint", so i think we can 

1. remove the "extra notes"

```
{% Aside %}
**Note:** This feature also works with [Fetch][10] requests.
{% endAside %}
```

2. Replace  all wording of`XHR breakpoint` to `XHR/ fetch breakpoint`

site/en/docs/devtools/javascript/breakpoints/index.md
```diff
@@ -157,27 +211,33 @@ all mouse events.
 3.  Check one of these categories to pause whenever any event from that category is fired, or expand
     the category and check a specific event.
 
-{% Img src="image/admin/c40PakNjXuOPIn21VySn.png", alt="Creating an event listener breakpoint.", width="800", height="543" %}
+{% Img src="image/NJdAV9UgKuN8AhoaPBquL7giZQo1/aFFgOPi3outuMb32WO3k.png", alt="Creating an event listener breakpoint.", width="800", height="726" %}
```
jecfish, line 214: Somewhere in this section mention the event listener breakpoint can be found in Elements panel > Event Listeners too

jecfish (changes requested):

site/en/docs/devtools/javascript/breakpoints/index.md
ktran: I agree that it is a bit misleading wrt to objects. I think it's a great idea to add more examples here! 

Actually, using `'x is' + x` is also valid. What happens with log points is that whatever you enter is wrapped into a `console.log`, so maybe adding this info (or something along these lines) will make it more apparent to developers on what syntax they can use here.

Adding @szuend , in case Simon has some input on this.

szuend: Maybe we can keep the initial instructions relatively simple, as they are now: "You can separate strings, variables, and objects with commas." But add a technical note or a footnote that explains that any JavaScript expression, or comma-separated list of JavaScript expressions is valid since it's wrapped as-is into a `console.log`.

site/en/docs/devtools/javascript/breakpoints/index.md
sofiayem: Ummm... Settings such a breakpoint in Sources doesn't populate the Elements > Event Listener list. Maybe because it's listeners, not breakpoints?

site/en/docs/devtools/javascript/breakpoints/index.md
sofiayem: I apologise to have hallucinated a misleading description. Thank you for catching that. Fixed.

site/en/docs/devtools/javascript/breakpoints/index.md
```diff
@@ -83,17 +83,12 @@ To set a logpoint:
 1.  Go to the line of code.
 1.  To the left of the line of code is the line number column. Right-click it.
 1.  Select **Add logpoint**. A dialog displays underneath the line of code.
-1.  Enter your log message in the dialog.
-    
-    You can separate strings, variable, and objects with commas. Wrap strings in quotes (`" "`), objects in curly brackets (`{ }`), and type variable names as they are.
-
-    For example: `"I'm a string.", variableName, {object}`.
-
+1.  Enter your log message in the dialog. You can use the same syntax as you would with a [`console.log(message)`](https://developer.mozilla.org/docs/Web/API/Console/log) call.
```
jecfish, line 86: perhaps, point to our documentation of `console.log` instead of MDN.

site/en/docs/devtools/javascript/breakpoints/index.md
jecfish: Oh yes, it Elements > Event Listener list only list the listeners, not breakpoints.

site/en/docs/devtools/javascript/breakpoints/index.md
```diff
@@ -56,7 +56,7 @@ console.log('c');
 ### Conditional line-of-code breakpoints {: #conditional-loc }
 
 Use a conditional line-of-code breakpoint when you know the exact region of code that you need to
-investigate, but you want to pause only when some other condition is true.
+investigate, but you want to pause only when some other condition is true. This can help you catch edge cases with variables.
```
jecfish, line 59: Conditional breakpoint is useful to not annoy yourself with irrelevant break during debugging. It helps developers focus on just stopping at the condition relevant to the current debugging, especially in a loop. 

Tho it's helpful for catching edge cases but not so much on that.

site/en/docs/devtools/javascript/breakpoints/index.md
jecfish: I think the the example prior to this change is helpful. I suggest instead of removing them all, rephrase them properly and use it here (e.g. `"a string", object, variable, variable.length > 1`).

jecfish (changes requested):

site/en/docs/devtools/javascript/breakpoints/index.md
sofiayem: Our console.log doc lacks syntax explanation: https://developer.chrome.com/docs/devtools/console/api/#log. Something to fix later.

jecfish (approved):

PR #5165, opened 2023-02-06
Update index.md

IanStanion-google:
Assessed page for deprecated MV2 information.


oliverdunk (approved):
LGTM. I pushed a couple of changes (I was going to comment but I couldn't comment on unchanged lines). I think there's an argument for including a version number somewhere but with how it was written it sounded like an outdated doc - Chrome 56 has been out for a long time and many other APIs are likely newer than it.

jpmedley (approved):

PR #5198, opened 2023-02-08
Do not autoplay videos for prefers-reduce-motion users

tunetheweb:

- Same as https://github.com/GoogleChrome/web.dev/pull/9406 but for this site


devnook (changes requested):
Thanks @tunetheweb ! If you want this to run on every page, you'd need to plug it in in main.js: https://github.com/GoogleChrome/developer.chrome.com/blob/main/site/_js/main.js

tunetheweb:
> Thanks @tunetheweb ! If you want this to run on every page, you'd need to plug it in in main.js: https://github.com/GoogleChrome/developer.chrome.com/blob/main/site/_js/main.js

🤦‍♂️ - committed that file this time!

PR #5200, opened 2023-02-08
Fugu API Showcase: Add image error handling

tomayac:

- Sometimes screenshots can't be created due to timeouts or because the sites block Puppeteer. This PR adds a fallback image in such cases.

Screenshot 2023-02-08 at 14 28 44


tomayac:
Unfortunately `i18n()` can only be called from the server, but with the invisible `<div>` trick I can pre-populate the text and set it dynamically in the SVG.

tomayac:
Example link to test this: https://deploy-preview-5200--developer-chrome-com.netlify.app/fugu-showcase/#color-picker-threejs.

matthiasrohmer (approved):

PR #5203, opened 2023-02-08
Fugu API Showcase: Add Web Share for Fugu Showcase

### LINKED ISSUE #5184: Add sharing with the Web Share API (including app screenshot) on supporting browsers on /fugu-showcase
Add sharing with the Web Share API (including app screenshot) on supporting browsers on /fugu-showcase.
Raised as part of https://github.com/GoogleChrome/developer.chrome.com/pull/5159

tomayac:
Fixes #5184 

- Add **Share app** action for Fugu Showcase on supporting browsers.
- Move **Copy link** action to lower right corner.

This is how it looks like:

- Supporting browser:
   Screenshot 2023-02-09 at 16 13 25
   Screenshot 2023-02-09 at 16 17 25
- Non-supporting browser:
   Screenshot 2023-02-09 at 16 14 37


site/_includes/partials/showcase-card.njk
```diff
@@ -38,8 +38,9 @@
         {{ 'i18n.common.view_source' | i18n(locale) }}
       </a>
     {% endif %}
-    <a class="display-flex align-center color-primary type--small fugu-app-link" href="/fugu-showcase/#{{ id }}">
-      #
+    <a class="display-none align-center color-primary type--small web-share" href="#">
```
matthiasrohmer, line 41: This should probably be a `button`.

site/_js/fugu-showcase.js
```diff
@@ -86,6 +88,102 @@ import './web-components/enhanced-select';
     new URL(document.location.href).searchParams.get('api')?.split(',') || [];
   onSearch();
 
+  if ('share' in navigator && 'canShare' in navigator) {
+    const imageToPNG = async blob => {
```
matthiasrohmer, line 92: Gave this a quick test on iOS. While I like using the capabilities of the Web Share API I think I'd vote to remove this. WhatsApp seems to ignore the image anyway, iMessage takes it but this introduces weirdness for users as the links and the image would appear as distinct entries in the "Shared with you entry". Also usually the app sites have their own og:image set anyway, which would probably be a better fit than our own screenshot too.

site/en/fugu-showcase/index.njk
```diff
@@ -63,6 +63,16 @@ alt: Blowfish swarm swimming in the ocean.
...
+
+<div class="display-none" id="i_just_found_the_app">
+  {{ 'i18n.fugu_showcase.i_just_found_the_app' | i18n(locale) }}
+</div>
+<div class="display-none" id="it_uses_these_fugu_apis">
+  {{ 'i18n.fugu_showcase.it_uses_these_fugu_apis' | i18n(locale) }}
+</div>
+<div class="display-none" id="via_the_fugu_showcase">
+  {{ 'i18n.fugu_showcase.via_the_fugu_showcase' | i18n(locale) }}
+</div>
```
matthiasrohmer, line 75: I'd vote to use a `<script id="fugu-showcase-i18n" type="application/json"></script>` or something instead.

matthiasrohmer (commented):
Thanks that's a good addition! Left some comments inline, but also I am not too sure how compatible this is with https://github.com/GoogleChrome/developer.chrome.com/pull/5242 as with this PR there is essentially no space anymore.

What do you think about moving the share action down to the bottom left corner of the card? I think that would also make sense from an UX perspective as the primary actions are "Launch app" and "View source code".

However the new sharing icon does not fit the others:

Screen Shot 2023-02-13 at 11 57 06

site/_js/fugu-showcase.js
tomayac: It works super well on Twitter, which is our primary share target I guess. I vouch for keeping it.

site/en/fugu-showcase/index.njk
tomayac: Will do.

site/_includes/partials/showcase-card.njk
tomayac: @matthiasrohmer Fixed, but I fail to style it. I need to essentially apply…

```css
background-color: transparent;
border: none;
````

…but I don't know how with this funny SCSS thingy we use. I tried `class="border-none background-transparent"`, but it didn't get picked up. Help wanted.

tomayac:
@matthiasrohmer See https://github.com/GoogleChrome/developer.chrome.com/pull/5203#discussion_r1104432785 for the styling issue.
 
Screenshot 2023-02-13 at 13 54 40

We may want to remove the border at the bottom for the second row of buttons.

site/_js/fugu-showcase.js
matthiasrohmer: Well, looks like I was biased when testing 😄 sounds good to me, since you could manually edit what's shared on iMessage.

matthiasrohmer:
@tomayac, updated the card stylings and fixed the icon sizings, PTAL:

image

tomayac:
> @tomayac, updated the card stylings and fixed the icon sizings, PTAL:

Works for me and clearly differentiates primary actions from secondary actions.

matthiasrohmer (approved):
Sorry for hijacking this, noticed a bug in the staging workflow that I was finally able to fix 🙂 Approved and merged!

PR #5208, opened 2023-02-08
recognize falsy values in .env ignoring variables

connorjclark:
This is a minor improvement, but it would have saved me a few minutes of confusion (not yet being familiar with how to build the site).

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,9 @@ ELEVENTY_IGNORE_WORKBOX=true
 ELEVENTY_IGNORE_BLOG=true
 ```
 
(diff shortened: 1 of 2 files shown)


matthiasrohmer (approved):
Thanks for the improvement! I'd appreciate if you could drop a quick note why you were using the flags - were you just tired waiting for the build? How long does it take for you?

connorjclark:
Honestly I didn't try the whole build at all - I was just changing some Lighthouse docs and during setup I saw this note about limiting the build and did it, figuring it'd save me some time given how huge the site is.

PR #5213, opened 2023-02-08
update Lighthouse perf ToC and scoring for 10.0 release

brendankenny:
In the imminent Lighthouse 10 release, TTI is bumped from the main scored metrics and score weight are adjusted accordingly.

Also makes a "retired metrics" section so the docs are still accessible but clearer that they're no longer part of the current tool.


malchata:
I'll get a review on this in the morning, @brendankenny!

malchata (approved):
LGTM.

PR #5214, opened 2023-02-09
Fix bug in social media links caused by unnecessary URL

lmssiehdev:
This pull request addresses a bug in the display of certain social media links. Currently, the links were showing up with an unnecessary URL, such as "https://twitter.com/https://twitter.com/username", instead of the intended format "https://twitter.com/username".

- Remove the unnecessary URL from affected social media links.

image

https://developer.chrome.com/articles/hidden-until-found/

### DIFF
--- a/site/_data/authorsData.json
+++ b/site/_data/authorsData.json
@@ -300,7 +300,7 @@
   },
   "justinschuh": {
     "country": "US",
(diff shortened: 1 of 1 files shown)


matthiasrohmer (approved):
Thank you for fixing this!

PR #5226, opened 2023-02-09
separate glossary entries b/265335974

nancymic2:
Fixes b/265335974

Take glossary entries out of main glossary index page, put each into its own partial, and add only partials refs to the glossary index.md

staging: https://pr-5226-static-dot-dcc-staging.uc.r.appspot.com/docs/privacy-sandbox/glossary/
live page: https://developer.chrome.com/docs/privacy-sandbox/glossary/


heyawhite (approved):
I know that was incredibly tedious, and I'm so excited about it. Thank you, Nancy! You rock!

PR #5229, opened 2023-02-09
Add new methods for chrome.action to 'What's new'.

jpmedley:
These should have been published in 108 and 109. That's why they're done in a weird way.


IanStanion-google (approved):

IanStanion-google:
@jpmedley LGTM!

PR #5237, opened 2023-02-10
Setup dependabot

matthiasrohmer:
This sets up dependabot to auto-create PRs for 11ty, webdev-infra and @mdn/browser-compat-data on a daily basis.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,8 +87,9 @@
     "unistore": "^3.5.2",
     "uslug": "^1.0.4",
     "web-vitals": "^3.0.4",
-    "webdev-infra": "1.0.42",
-    "xss": "^1.0.14"
+    "webdev-infra": "1.0.44",
+    "xss": "^1.0.14",
+    "@mdn/browser-compat-data": "5.2.33"
   },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -87,8 +87,9 @@
     "unistore": "^3.5.2",
     "uslug": "^1.0.4",
     "web-vitals": "^3.0.4",
-    "webdev-infra": "1.0.42",
-    "xss": "^1.0.14"
+    "webdev-infra": "1.0.44",
+    "xss": "^1.0.14",
+    "@mdn/browser-compat-data": "5.2.33"
```
andreban, line 92: Probably a silly question - but I expected `@mdn/browser-compat-data` to already exist as a dependency, so double checking if this WAI.

andreban: Discussed offline.

andreban (approved):
LGTM

package.json
matthiasrohmer: It is a dependency of webdev-infra, with a version range. To make dependabot pick it up it needs to be specified in the main package.json as well.

PR #5242, opened 2023-02-10
[Fugu API Showcase] Add Electron.js label

tomayac:

- Add Electron.js label 

Screenshot 2023-02-10 at 16 48 12

### DIFF
--- a/site/_includes/partials/showcase-card.njk
+++ b/site/_includes/partials/showcase-card.njk
@@ -29,7 +29,11 @@
     {% if item.appURL %}
       <a class="display-flex align-center color-primary type--small" href="{{ item.appURL }}" rel="noopener" target="_blank">
         {{ icon('external-link') }}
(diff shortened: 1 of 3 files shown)


site/_includes/partials/showcase-card.njk
```diff
@@ -32,6 +32,9 @@
         {{ 'i18n.common.launch_app' | i18n(locale) }}
```
matthiasrohmer, line 32: I wonder if the CTA should also change to "Download app" in this case...? WDYT?

tomayac: Good point. Will make it so.

tomayac:
"Download app" when Electron:

Screenshot 2023-02-13 at 12 49 56

matthiasrohmer (approved):
Thanks, Tom 🙂

PR #5243, opened 2023-02-10
Improve documentation on offscreen document lifetime enforcement.

### LINKED ISSUE #5019: Document lifetime_enforcer and audio_lifetime_enforcer

### LINKED ISSUE #5095: Define lifetime enforcement in terms of observable behavior
>Unless I'm missing something, it does not seem like the observable behavior is so complicated that someone would need to dig into it.

_Originally posted by @jpmedley in https://github.com/GoogleChrome/developer.chrome.com/pull/5081#pullrequestreview-1277661538_
           
Links to source code should always be an extra, never the primary means of conveying information.

oliverdunk:
Explicitly document the lifetime enforcement for an offscreen document based on the specified reason.

Fixes #5019
Fixes #5095

### DIFF
--- a/site/en/docs/extensions/reference/offscreen/index.md
+++ b/site/en/docs/extensions/reference/offscreen/index.md
@@ -23,7 +23,12 @@ Pages loaded as offscreen documents are handled differently from other types of 
 * An extension can only have one offscreen document open at a time. If the extension is running in split mode with an active incognito profile, both the normal and incognito profiles can each have one offscreen document. 
 
 ## Reasons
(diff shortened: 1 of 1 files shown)


jpmedley (approved):

jpmedley:
@oliverdunk I don't know if I mentioned this. Submitters merge their own PRs. The writer usually knows better when something needs to go out. This guards against merging at the wrong time.

PR #5251, opened 2023-02-11
Update soft navs doc for new flag

tunetheweb:
There's a [new flag needed on performance observers](https://github.com/WICG/soft-navigations/issues/11#issuecomment-1423909430), and looks like it's landed in Canary already as got a compliant it's not working anymore.

So updating doc to include that.

FYI @yoavweiss

### DIFF
--- a/site/en/blog/soft-navigations-experiment/index.md
+++ b/site/en/blog/soft-navigations-experiment/index.md
@@ -8,7 +8,7 @@ authors:
  - tunetheweb
  - yoavweiss
 date: 2023-02-01
-#updated: 2023-02-01
+updated: 2023-02-11
 hero: image/W3z1f5ZkBJSgL1V1IfloTIctbIF3/ZJLD5QJvoFcWXkqF8m5P.jpg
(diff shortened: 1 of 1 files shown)


malchata (approved):
LGTM!

site/en/blog/soft-navigations-experiment/index.md
```diff
@@ -70,7 +70,17 @@ Site owners can choose to include the origin trial on their pages for all, or fo
 
 ## How can I measure Core Web Vitals per soft navigation?
 
-Some considerations need to be taken into account when attempting to measure Core Web Vitals for soft navigations.
+To includes soft navigations you need to include `includeSoftNavigationObservations: true` in your performance observer's `observe` call:
```
yoavweiss, line 73: Should we be explicit about this being the case starting from Chrome 112?

tunetheweb: I considered this, but since there's no harm of setting this now I'm fine with keeping it simple.

PR #5262, opened 2023-02-13
[Fugu API Showace] Add RSS feed

### LINKED ISSUE #5258: [Fugu Showcase] Add RSS feed
Add RSS feed

tomayac:
Fixes #5258

- Add RSS feed.

Screenshot 2023-02-13 at 17 41 20

Screenshot 2023-02-13 at 17 40 24


site/en/fugu-showcase/index.njk
```diff
@@ -54,6 +54,10 @@ alt: Blowfish swarm swimming in the ocean.
           {% endfor %}
         </select>
       </enhanced-select>
+      <a href="./feed.xml" class="material-button button-filled button-round display-inline-flex color-bg bg-cyan-medium">
```
tomayac, line 57:
```suggestion
      <a href="/fugu-showcase/feed.xml" class="material-button button-filled button-round display-inline-flex color-bg bg-cyan-medium">
```

tomayac:
This is how the feed renders in [Feedly](https://feedly.com/):

Screenshot 2023-02-14 at 13 19 09

matthiasrohmer (commented):
Only glanced at it so far, but before I draw wrong assumptions, can you help me understand why you choose to add the XML building to the external data flow instead of adding a simple template like /site/en/fugu-showcase/feed.xml and go from there...? But I think I am missing something.

devnook:
Hi @tomayac - thanks for adding the RSS capabilities to the Fugu page!
In order to make it compatible with the rest of the site (for consistency and maintainability, and to avoid future brittleness of the code), we'd need to make it use more 11-ty driven, declarative approach though.

The steps I'd recommend to follow:
1. Pre-process and fetch additional data in external/build/fugu-showcase.js. Add the resulting data to data.json.
2. Add site/en/fugu-showcase/feed.njk template, similar to https://github.com/GoogleChrome/developer.chrome.com/blob/main/site/feed.njk
3. Populate the feed.njk template with variables from data.json (same as site/en/fugu-showcase/index.njk)

Let me know if you need help with this!

tomayac:
@devnook I have now made the requested changes and it's working perfectly fine (https://deploy-preview-5262--developer-chrome-com.netlify.app/fugu-showcase/feed.xml). I think I'm doing something wrong with the `pagination` frontmatter, since I need to set the `size` to a big number so it doesn't actually paginate. I guess there's a better way that you know… `:-)` PTAL!

site/en/fugu-showcase/feed.njk
```diff
@@ -0,0 +1,69 @@
+---
+permalink: /fugu-showcase/feed.xml
+disable_algolia: true
+pagination:
+  data: fuguShowcase.fuguItems
+  size: 10000000000
```
tomayac, line 6: There's probably a better way and I'm just holding Eleventy wrong.

site/_data/i18n/fugu_showcase.yml
```diff
@@ -4,6 +4,9 @@ select_api:
 search_apps:
   en: Search apps
 
+rss_feed:
+  en: RSS feed
```
devnook, line 8: It's useful enough to have it in common.yaml

tomayac: Done in https://github.com/GoogleChrome/developer.chrome.com/pull/5262/commits/22b1fa01bb5955b73185c4e3e8c9c8537f888897.

site/_data/fuguShowcase.js
```diff
@@ -20,25 +20,47 @@
 
 const path = require('path');
 const fs = require('fs');
+const {default: fetch} = require('node-fetch');
 
-module.exports = () => {
+async function getContentLength(url) {
+  try {
```
devnook, line 26: This should happen when you build fugu-showcase.json file, in build-exernal. Fetching all images in _data will cause the images to be fetched for _every_ build of dcc, and this is too heavy.

tomayac: I have outsourced this now directly into https://googlechromelabs.github.io/fugu-showcase/data/data.json where I have this information available already: https://github.com/GoogleChromeLabs/fugu-showcase/commit/d98f8ad3b228e846ea1baa31d230b0a7ebf9ff41#diff-d6299b5bb2afc0956126a0adb12fd89829e01a4cc025e7277a092228b96f32e4.

site/_data/fuguShowcase.js
```diff
@@ -21,24 +21,28 @@
...
   const fuguItems = JSON.parse(fs.readFileSync(fuguShowcaseFile, 'utf-8'));
 
   const availableAPIs = new Set();
-  fuguItems.forEach(item => {
-    item.usedAPIs.forEach(api => {
-      availableAPIs.add(api.name);
-    });
-  });
-
+  await Promise.all(
```
devnook, line 32: I think this does not need to be async anymore?

devnook (commented):
Hi @tomayac ! I fixed the pagination. I moved the feed to feeds/fugu-showcase.xml to make it compatible with other feeds on the site. Is it ok?

Apart from the last comment about async/await, it's good to go.

site/_data/fuguShowcase.js
tomayac: Oh, right. That's a left-over. Cleaning it up.

tomayac:
I fixed the link on the page, too. It seems to work all fine now. Woohoo! Thanks for the pagination fix. This was just too simple. I wasn't aware the variable is just exposed like that…

devnook (approved):

PR #5273, opened 2023-02-14
Fix `enhanced-select` on iOS and improve styling

### LINKED ISSUE #5186: <enhanced-select> shows long double-line option name like 2 options
List items in the enhanced-select use line-height instead of padding to space items vertically which makes items that span over 2 lines (becasue of long option names) look like they were two items, while they are actually one.

E.g. PointerLock and unadjustedMovement in /fugu-showcase:
https://deploy-preview-5159--developer-chrome-com.netlify.app/fugu-showcase

Screenshot 2023-02-07 at 12 32 28

### LINKED ISSUE #5236: `<enhanced-select>` component is broken on Safari (`Version 16.3 (18614.4.6.1.5)`)
The `<enhanced-select>` component is also completely broken on Safari (`Version 16.3 (18614.4.6.1.5)`):

- Safari:
   Screenshot 2023-02-09 at 16 33 00
- Chrome:
   Screenshot 2023-02-09 at 16 33 10

_Originally posted by @tomayac in https://github.com/GoogleChrome/developer.chrome.com/issues/5186#issuecomment-1424387851_

matthiasrohmer:
Fixes #5186 and fixes #5236.

Technically there is a [polyfill](https://www.npmjs.com/package/element-internals-polyfill) for `ElementInternals` but I think it's not really worth it bringing it in, since we are not using the element in form contexts.

### DIFF
--- a/site/_js/web-components/enhanced-select.js
+++ b/site/_js/web-components/enhanced-select.js
@@ -35,8 +35,12 @@ export class EnhancedSelect extends BaseElement {
   constructor() {
     super();
 
-    // @ts-ignore
-    this.internals = this.attachInternals();
+    try {
+      // @ts-ignore
+      this.internals = this.attachInternals();
+    } catch (e) {
+      console.warn('ElementInternals not supported');
+    }
 
     this.handleLabelClick = this.handleLabelClick.bind(this);
     this.handleLabelKeydown = this.handleLabelKeydown.bind(this);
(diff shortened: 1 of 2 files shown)


matthiasrohmer:
@tomayac, sending for you to review as Ewa is OOO and the Fugu Showcase is the only page currently using that element 🙂

tomayac (approved):
LGTM, approval

Works as expected in Safari now.

As you're at it, may I suggest adding `:defined` styling to [hide the custom element until it's defined](https://developer.mozilla.org/en-US/docs/Web/CSS/:defined#hiding_elements_until_they_are_defined), since there's a short moment where the non-final version shows, which doesn't look great.

Screenshot 2023-02-14 at 13 42 50

matthiasrohmer:
Ah very good suggestion 😄 it also bothered me when fixing those two bugs, but I didn't think about fixing it. Now done!

Issue #3959, opened 2022-10-10
Implement playlist component for Landing Pages


devnook:
![Image](https://user-images.githubusercontent.com/1914261/194893630-c6695f31-a550-4c86-b6d1-14d12026fe51.png)

https://www.figma.com/file/koVhYNpf2SXKuqm2yJ8qkj/Workbox?node-id=413%3A29491

simgobb:
03.02 / Submitted Data Fetching Plan doc. Suggested that the data fetching would happen through an external data flow as opposed to the solution implemented originally in the PR. 

CC @mollietokyo

mollietokyo:
Estimate = 3 days remaining (SPR-3)

simgobb:
The external data flow fetching the data has been approved and merged.

The component part needs extra work as it was originally thought of as a shortcode, but to keep things consistent with the rest of the landing page components, it needs to be refactored in a macro.

To keep the macro simple, a lot of the logic will need to be moved in the data-fetching part of the PR. 

CC @mollietokyo @matthiasrohmer

matthiasrohmer:
Won't fix, due to the pending move to the new CMS.

PR #5314, opened 2023-02-16
You tube external data flow

simgobb:
Adds External data flow for YouTube Playlists. 

Needed for issue #3959 and component in PR #5027


external/build/youtube-playlist.js
```diff
@@ -0,0 +1,181 @@
...
+
+require('dotenv').config();
+
+const fs = require('fs');
+const path = require('path');
+const {google} = require('googleapis');
+
+// Set an interval of time that makes this data stale and converts it in ms
+const FETCH_INTERVAL_HOURS = 4;
+const FETCH_INTERVAL_MILLISECONDS = FETCH_INTERVAL_HOURS * 3600000;
```
matthiasrohmer, line 13: Since `ms` is a peer dependency already (via `gulp` and a few others), let's install it as a direct dependency and use it here to make things more readable.

external/build/youtube-playlist.js
```diff
@@ -0,0 +1,181 @@
...
+
+// Set all the config constansts needed for the API calls
+const API_KEY = process.env.YOUTUBE_API_KEY;
+const PART = 'snippet';
+const CHANNELS = 'UCnUYZLuoy1rq1aVMwx4aTzw';
```
matthiasrohmer, line 18: Add a comment/link which channel this is

matthiasrohmer (commented):
Left some more comments, mostly style related.

external/build/youtube-playlist.js
```diff
@@ -0,0 +1,204 @@
+/**
```
matthiasrohmer, line 1: Missing license header.

external/build/youtube-playlist.js
```diff
@@ -0,0 +1,204 @@
...
+
+/**
+ * Target directory for the fetched data
+ */
+const targetFile = path.join(__dirname, '../data/youtube-playlist.json');
+const currentTimestamp = Date.now();
+
+const result = {};
+
+const youtube = google.youtube('v3');
```
matthiasrohmer, line 37: I think I'd merge this with the require above, so `const youtube = require('googleapis').google.youtube('v3')` - what a weird way to structure a package anyway.

external/build/youtube-playlist.js
```diff
@@ -0,0 +1,204 @@
...
+    const currentDataRaw = await fs.readFile(targetFile, {encoding: 'utf8'});
+    const currentData = JSON.parse(currentDataRaw);
+
+    dataAge = currentTimestamp - currentData.timestamp;
+  } catch (error) {
+    console.warn(
+      'No existing data found or existing data malformatted. Fetching from YouTube...'
+    );
+  }
+  if (dataAge > FETCH_INTERVAL || !dataAge) {
```
matthiasrohmer, line 63: The following block is fairly small but it's usually good practice to inverse the condition and early exit, so

```js
if (dataAge && dataAge < FETCH_INTERVAL) {
  console.log(`YouTube data is ${ms(dataAge)} old, no need to fetch.`);
  return;
}
```

external/build/youtube-playlist.js
```diff
@@ -0,0 +1,204 @@
...
+
+/**
+ * Fetches all the meta data for the playlist and then calls the function
+ * needed for fetching the single videos data
+ * @param {string} id The YouTube channel id
+ * @return {promise} A promise that resolves in the YouTube playlist data
+ * neeeded to display the playlist component
+ */
+
+async function getPlaylistData(id) {
```
matthiasrohmer, line 101: Rename `id` to `channelId` to have it obvious.

external/build/youtube-playlist.js
```diff
@@ -0,0 +1,204 @@
...
+    part: PART,
+    channelId: id,
+    auth: API_KEY,
+    maxResults: MAX_RESULTS,
+  });
+
+  const playlists = response?.data?.items;
+
+  if (!playlists || playlists.length === 0) {
+    console.warn('No playlist found');
```
matthiasrohmer, line 113: Similar to above, to make things easier to read and don't require the `else` indentation, instead early return.

external/build/youtube-playlist.js
```diff
@@ -0,0 +1,204 @@
...
+    part: PART,
+    playlistId: id,
+    auth: API_KEY,
+    maxResults: MAX_RESULTS,
+  });
+
+  const videos = response.data.items;
+
+  if (!videos || videos.length === 0) {
+    throw new Error('No playlist videos found');
```
matthiasrohmer, line 160: Early return.

external/build/youtube-playlist.js
```diff
@@ -0,0 +1,204 @@
...
+  const response = await youtube.channels.list({
+    part: PART,
+    id: id,
+    auth: API_KEY,
+  });
+
+  const channels = response.data.items;
+
+  if (!channels || channels.length === 0) {
+    throw new Error('No channel found');
```
matthiasrohmer, line 193: Is that an error really? I think I would simply return early, that could also save a try/catch further up 🙂

matthiasrohmer (commented):
We are almost there, just some slight structure comments 🙂

external/build/youtube-playlist.js
```diff
@@ -0,0 +1,213 @@
...
+    dataAge = currentTimestamp - currentData.timestamp;
+  } catch (error) {
+    console.warn(
+      'No existing data found or existing data malformatted. Fetching from YouTube...'
+    );
+  }
+  if (dataAge < FETCH_INTERVAL && dataAge) {
+    console.log(`YouTube data is ${ms(dataAge)} old, no need to fetch.`);
+    return;
+  }
```
matthiasrohmer, line 79: Sorry I actually changed my mind here. The current behaviour would require us to run `sync-external` before `build-external`, which might not be the best solution. Instead we could do the following: use `fetch` to try to fetch https://www.googleapis.com/storage/v1/b/external-dcc-data/o/youtube-playlist.json (currently 404s) and guesstimate wether a refetch is needed from there.

Look at https://www.googleapis.com/storage/v1/b/external-dcc-data/o/chrome-release.json, there is a `updated` key which is exactly what we want.

Instead of `fetch` try to use @google-cloud/storage, which already is a dependency anyway.

external/build/youtube-playlist.js
```diff
@@ -0,0 +1,213 @@
...
+}
+
+/**
+ * Fetches all the channel meta data the component will need to render correctly the playlist.
+ * @param {array} id The YouTube channel id
+ * @return {promise} A promise that resolves in the YouTube Channel data needed
+ * to display the playlist component
+ */
+
+async function getChannelData(id) {
```
matthiasrohmer, line 192: Nit, no new line after JSDoc comments.

external/build/youtube-playlist.js
```diff
@@ -0,0 +1,213 @@
...
+      thumbnail: video?.snippet?.thumbnails?.medium?.url,
+    });
+  }
+
+  return data;
+}
+
+/**
+ * Fetches all the channel meta data the component will need to render correctly the playlist.
+ * @param {array} id The YouTube channel id
```
matthiasrohmer, line 187: Is it an `array` or rather a `string`?

matthiasrohmer (commented):
Sorry for holding this so long. I was thinking about the waterfall we have been creating here for too long and now proposed an alternative solution.

external/build/youtube-playlist.js
simgobb: I've created a function that using @google-cloud/storage, tries to fetch the data from the bucket, check the timestamp (if exists) and returns a boolean to indicate if the data is stale or not. 

Let me know if that's the direction you had in mind!

external/build/youtube-playlist.js
```diff
@@ -0,0 +1,238 @@
...
+    stream.on('error', error => {
+      console.error('Error reading the file', error);
+      return true;
+    });
+
+    return true;
+  } catch (error) {
+    console.error('Error reading file, refetching from YouTube', error);
+    return true;
+  }
```
matthiasrohmer, line 198: This can be heavily simplified 🙂  you can use `getMetadata` instead. See https://github.com/googleapis/nodejs-storage/blob/0501127f14deb7403676c93fcb53eb302fe618b7/samples/getMetadata.js#L45-L67

This saves us from handling the stream, parsing the JSON, ...

package.json
```diff
@@ -64,6 +64,7 @@
     "express": "^4.17.1",
     "glob": "^7.1.7",
     "google-auth-library": "^8.5.2",
+    "googleapis": "^110.0.0",
```
matthiasrohmer, line 67: Please pin newly introduced dependencies (remove the `^`) to not rely on package-lock.json shenanigans.

matthiasrohmer (approved):

PR #5327, opened 2023-02-16
[lighthouse-10] fix PSI launch date and alt text

brendankenny:
c/p error on the PSI rollout date for Lighthouse 10, and I missed an alt text for the last image


malchata (approved):
LGTM!

PR #5328, opened 2023-02-16
[DO NOT MERGE] Test meet PS team with chrome events

heyawhite:


rachelandrew:
See my comments here https://github.com/GoogleChrome/developer.chrome.com/issues/5333#issuecomment-1434700042 this approach is going to cause code and content duplication, which we want to avoid. I think there's likely a more elegant (and useful to other teams) way to achieve this.

heyawhite:
This PR existed as a demonstration, rather than active plan! Will close.

PR #5374, opened 2023-02-20
Fix a small typo

berkay-ozkan:
I believe there is a minuscule typo [right here](https://developer.chrome.com/docs/extensions/mv3/getstarted/tut-reading-time/#:~:text=page%20or%20other-,extension%27s,-content%20scripts.). Changed "...their host page or other extension's content scripts." to "...their host page or other extensions' content scripts."


AaronForinton (approved):
LGTM, thank you.

PR #5377, opened 2023-02-20
fix: show non-article/blog/docs pages in search

### LINKED ISSUE #5309: Landing pages aren't appearing in search
I've had someone raise by email that landing pages aren't appearing in search. To reproduce search for "Fugu" or "Capabilities". I'd expect that to return the Capabilities landing page at https://developer.chrome.com/capabilities/

Blazzike:
Fixes #5309 

- Adds a new "other" category to search results which contains previously missing results
- Refactors/cleans up existing code

Screenshot:
Screenshot 2023-02-20 at 12 39 38

### DIFF
--- a/site/_js/web-components/search-box.js
+++ b/site/_js/web-components/search-box.js
@@ -70,7 +70,7 @@ export class SearchBox extends BaseElement {
     this._active = false;
     this.buttonLabel = 'open search';
     this.docsLabel = 'Documentation';
-    this.otherLabel = 'Other';
+    this.overviewLabel = 'Overview';
     this.articlesLabel = 'Articles';
     this.blogLabel = 'Blog';
     this.locale = 'en';
@@ -341,6 +341,7 @@ export class SearchBox extends BaseElement {
       };
 
       this.categorisedResults = {
+        [this.overviewLabel]: [],
         [this.docsLabel]: mutableResults.filterMutate(r => r.type === 'doc'),
         [this.articlesLabel]: mutableResults.filterMutate(
           r => r.type === 'article'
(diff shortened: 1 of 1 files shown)


matthiasrohmer (changes requested):
Gave this a go with this config:

```js
const client = algoliasearch('0PPZV3EY55', '09e331f9c299902dbc8ea18da6df29b1');
```

and also discussed this with Rachel who brought this up. Currently this change would add landing pages to the bottom of the search result list, which isn't really matching their importance as they are usually a good match for someone looking for a specific keyword like *Capabilities* as it is giving a good introduction.

So instead of introducing `Other` as a category let's introduce a `Overview` category which comes first and holds landing pages. Another bit I noticed is that with the current change, pages can appear twice: for example searching for "Privacy Sandbox" makes the page "The Privacy Sandbox" appear in both "Documentation" and "Other" which seems off. Especially because it's with documentation once, and once without.

Blazzike:
> Gave this a go with this config:
> 
> ```js
> const client = algoliasearch('0PPZV3EY55', '09e331f9c299902dbc8ea18da6df29b1');
> ```
> 
> and also discussed this with Rachel who brought this up. Currently this change would add landing pages to the bottom of the search result list, which isn't really matching their importance as they are usually a good match for someone looking for a specific keyword like _Capabilities_ as it is giving a good introduction.
> 
> So instead of introducing `Other` as a category let's introduce a `Overview` category which comes first and holds landing pages. Another bit I noticed is that with the current change, pages can appear twice: for example searching for "Privacy Sandbox" makes the page "The Privacy Sandbox" appear in both "Documentation" and "Other" which seems off. Especially because it's with documentation once, and once without.

I've fixed the naming & position, however, the duplicate result is more of an indexing issue. The results aren't duplicate so to speak, one of them just ends up redirecting to the other (/docs/privacy-sandbox/ -> /privacy-sandbox/) not sure if that's really in scope to be fixed by this PR?

matthiasrohmer (approved):
Good point, let's not derail this 🙂  Now works as intended, thanks!

PR #5406, opened 2023-02-22
Add a page listing all known Recorder extensions

OrKoN:
Fixes https://crbug.com/1413168

We propose adding a simple standalone page listing all known Recorder extensions in the alphabetical order. The page will be linked to from the Recorder panel UI. Currently, there is no way to find all Recorder extensions on the web store so this page would serve as a workaround until it becomes possible. It will be linked using a URL shortener so the link can be updated without changes to the product code.


OrKoN:
For the list of example extensions embedded into the Recorder reference page, I suggest we keep it with the source code references and link to the full list maintained on the page as well.

OrKoN:
I am not sure why the new page does not show up in the menu. Perhaps someone could advise on that.

jecfish:
Over to Sofia on the docs review and add to our menu

site/en/docs/devtools/recorder/extensions/index.md
```diff
@@ -1,40 +1,62 @@
...
+This page lists knows Chrome extensions for the DevTools **Recorder**.
+
 ## Export extensions
 
 {% Aside %}
 **Note**: This feature is available from Chrome version 104.
 {% endAside %}
 
-Export extensions allow customizing the code generated by the Recorder panel:
+Export extensions let you to customize the code generated by the **Recorder** panel:
```
OrKoN, line 23: I don't think we need extra text for every extension. I feel it is redundant since all extensions do the same thing. Any other opinions?

sofiayem:
> I am not sure why the new page does not show up in the menu. Perhaps someone could advise on that.

That's because it wasn't added to the table of contents. Fixed that and removed the list from the reference doc.
Thanks! LGTM

sofiayem (approved):

site/en/docs/devtools/recorder/extensions/index.md
sofiayem: I just saw this comment. I dunno. We can kinda promote third-party projects by providing more links. I reorg'd the list into a glossary, it read better.

sofiayem: WDYT?

OrKoN: That sounds fine by me but I have two concerns as a user:

1) I can read detailed descriptions on the web store page if I am interested in a new tool. If I already know the tools or use specific tools, then the information is redundant. Additional text content requires more time to read through the list to see if your favourite tools are supported.
2) maintaining the text description would be more work. And if we the vendors don't like them anymore or want to update slogans slogans for whatever reason, it will require a manual update unlike the webstore that offers the self-service UI.

OrKoN: I like how the glossary looks visually though! so LGTM.

OrKoN:
Thanks @sofiayem! @jecfish I think we can land it. If you could create a shortener link, I can open a CL to add it to the UI.

Issue #4195, opened 2022-11-03
TOC for supernested docs


heyawhite:
**Is your feature request related to a problem? Please describe.**
The FLEDGE API now has substantial documentation which needs to live on DCC (and not MDN). These docs will be merged in https://github.com/GoogleChrome/developer.chrome.com/pull/4172

The growing list of API references are exposed in a section at the bottom of the article, rather than in the side navigation.

![All FLEDGE API references](https://user-images.githubusercontent.com/2766281/199742173-0064fbe9-85f4-43e2-80b5-e2e0956954b4.png)

The side navigation already has three levels of docs, and these would fit in a fourth level of nesting. I believe that would be a bad user experience, and look messy.

#### Current navigation
![Privacy Sandbox nested navigation](https://user-images.githubusercontent.com/2766281/199741625-7e8f7af4-a251-4ec3-87e0-5204946147ba.png)

1. **CMA commitment and Privacy Sandbox use case**.
  1. Proposal overview (in the proposal stage, no additional docs needed.
  2. **Description of an API use case**
    1. Proposal overview (in the experiment phase)
    2. Developer API guide
    3. Experiment & participate
    4. Updates
  3. **Description of an API use case**
    1. Proposal overview (in the experiment phase)
    2. Developer API guide
    3. Experiment & participate
    4. Updates

**Describe the solution you'd like**
The current navigation suffices for all of the top level docs. In practice, the full navigation structure works as follows.

1. **CMA commitment and Privacy Sandbox use case**.
   1. Proposal overview (in the proposal stage, no additional docs needed.
   2. **Description of an API use case**
      1. Proposal overview (in the experiment phase)
      2. Developer API guide
          1. API method 1
          2. API method 2
          3. [...]
          4. API method x
      3. Experiment & participate
      4. Updates
   3. **Description of an API use case**
      1. Proposal overview (in the experiment phase)
      2. Developer API guide
      3. Experiment & participate
      4. Updates

When we get to a Developer API guide, it would be terrific if sidebar TOC could shift to focus on the second, third and fourth level. 
If you're looking at an API method for FLEDGE, you don't currently need to care about an API method for a different use case or even a different CMA commitment.

1. **Description of an API use case**
   1. Proposal overview (in the experiment phase)
   2. Developer API guide
      1. API method 1
      2. API method 2
      3. [...]
      4. API method x
   3. Experiment & participate
   4. Updates

matthiasrohmer:
[...]

PR #5438, opened 2023-02-24
Fix supernest on landing page

heyawhite:
Fixes re-opened #4195

Adds 'nested-column' style for anything that lives in a nested column.

The design could still be improved but is no longer a squished version of a document.


heyawhite:
@matthiasrohmer there may be a more elegant solution than this, but this is definitely a small and fast solution.

matthiasrohmer (approved):
Thanks a lot @heyawhite for fixing this!

PR #5462, opened 2023-02-27
Run npm init before creating temporary package with workbox dependencies for build-external script

### LINKED ISSUE #5461: build-external script fails
build-external script fails with the following message:

`Error: failed to parse typedoc: The --name option was not specified, and package.json does not have a name field. Defaulting project name to "Documentation".`

devnook:
Fixes #5461

It looks like the new TS version does not like the fact, that the temporary package create from Workbox dependencies did not have a name.

I added an npm init call to fix it. It works locally, we need to merge to main to test on prod.


matthiasrohmer (approved):
Let's give it a try!

PR #5482, opened 2023-02-28
Fix Shared Storage frequency cap link

heyawhite:
Update broken URL


heyawhite:
I noticed the URLs for the sample code were no longer right, so I've updated the code and URLs accordingly. Let me know if you have any concerns.

PR #5492, opened 2023-03-01
Update storage.session API quota.

oliverdunk:
Updates our documentation to reflect that the storage.session quota has been increased to 10 MB.

### DIFF
--- a/site/en/docs/extensions/mv3/known-issues/index.md
+++ b/site/en/docs/extensions/mv3/known-issues/index.md
@@ -56,7 +56,7 @@ This section lists major features that will be added to the Manifest V3 platform
 
 **Summary:** When it was introduced, the `session` storage area had an intentionally conservative maximum quota of 1 MB. This limit is being increased to 10 MB.
 
-**Estimated timeline**: This will be available in Chrome 112.
+**Estimated timeline**: This will be available in Chrome 112. Chrome 112 beta is planned around March 9, with a stable release around the beginning of April.
 
 ## Bugs {: #bugs }
 
--- a/site/en/docs/extensions/mv3/known-issues/index.md
+++ b/site/en/docs/extensions/mv3/known-issues/index.md
@@ -54,9 +54,9 @@ This section lists major features that will be added to the Manifest V3 platform
 
 ### Increased quota for session storage in the Storage API {: #increased-session-storage-quota }
 
-**Summary:** When it was introduced, the `session` storage area had an intentionally conservative maximum quota of 1 MB. We are planning to increase this limit, but have not yet settled on a new value.
+**Summary:** When it was introduced, the `session` storage area had an intentionally conservative maximum quota of 1 MB. This limit is being increased to 10 MB.
 
-**Estimated timeline**: Targeting Canary support in the first quarter of 2023.
+**Estimated timeline**: This will be available in Chrome 112.
 
 ## Bugs {: #bugs }
 
--- a/site/en/docs/extensions/reference/storage/index.md
+++ b/site/en/docs/extensions/reference/storage/index.md
@@ -44,7 +44,13 @@ Local and sync storage areas should not store confidential user data because the
 {% endAside %}
 
 [storage.session][prop-session]
-: Holds data in memory for the duration of a browser session. By default, it's not exposed to content scripts, but this behavior can be changed by setting [`chrome.storage.session.setAccessLevel()`][method-access-level]. The quota limitation is 1 MB approx. Consider using it to store global variables across service worker runs.
+: Holds data in memory for the duration of a browser session. By default, it's not exposed to content scripts, but this behavior can be changed by setting [`chrome.storage.session.setAccessLevel()`][method-access-level]. The quota limitation is approximately 10 MB. Consider using it to store global variables across service worker runs.
+
+{% Aside 'warning' %}
+
+Before Chrome 112, the quota was approximately 1 MB.
+
+{% endAside %}
 
 [storage.managed][prop-managed]
 : Administrator can use a [schema][manifest-storage] and enterprise policies to configure a supporting extension's settings in a managed environment. This storage area is read-only.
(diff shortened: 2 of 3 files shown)


site/en/docs/extensions/mv3/known-issues/index.md
```diff
@@ -54,9 +54,9 @@ This section lists major features that will be added to the Manifest V3 platform
 
 ### Increased quota for session storage in the Storage API {: #increased-session-storage-quota }
 
-**Summary:** When it was introduced, the `session` storage area had an intentionally conservative maximum quota of 1 MB. We are planning to increase this limit, but have not yet settled on a new value.
+**Summary:** When it was introduced, the `session` storage area had an intentionally conservative maximum quota of 1 MB. This limit is being increased to 10 MB.
 
-**Estimated timeline**: Targeting Canary support in the first quarter of 2023.
+**Estimated timeline**: This will be available in Chrome 112.
```
jpmedley, line 59:
```suggestion
**Estimated timeline**: This will be available in Chrome 112. Chrome 112 beta is planned around March 9, with a stable release around the beginning of April.
```

jpmedley: The reason I'm avoiding definite language for the release dates is that they're not guaranteed. The release time for stable is always on a Tuesday, but the time varies. The release time for beta is always two days later at around 11:00 Pacific. Delays are not unheard of. I once saw a beta that didn't make it out until the following Monday. https://chromiumdash.appspot.com/schedule

site/en/docs/extensions/reference/storage/index.md
```diff
@@ -44,7 +44,13 @@ Local and sync storage areas should not store confidential user data because the
...
 
 [storage.session][prop-session]
-: Holds data in memory for the duration of a browser session. By default, it's not exposed to content scripts, but this behavior can be changed by setting [`chrome.storage.session.setAccessLevel()`][method-access-level]. The quota limitation is 1 MB approx. Consider using it to store global variables across service worker runs.
+: Holds data in memory for the duration of a browser session. By default, it's not exposed to content scripts, but this behavior can be changed by setting [`chrome.storage.session.setAccessLevel()`][method-access-level]. The quota limitation is approximately 10 MB. Consider using it to store global variables across service worker runs.
+
+{% Aside 'warning' %}
+
+Before Chrome 112, the quota was approximately 1 MB.
+
+{% endAside %}
```
jpmedley, line 53:
```suggestion
{% Aside 'warning' %}
Before Chrome 112, the quota was approximately 1 MB.
{% endAside %}
```

jpmedley: I actually don't remember if the widget honors that spacing. The point is that the widget controls it, and we need to be consistent with the rest of the site.

site/en/docs/extensions/whatsnew/index.md
```diff
@@ -16,6 +16,12 @@ tags:
 Check this page often to learn about changes to the Chrome extensions platform, its documentation,
 and related policy or other changes. You'll find other notices posted on the [Extensions Google Group](https://groups.google.com/a/chromium.org/g/chromium-extensions). The [Extensions News](/tags/extensions-news/) tag lists articles about many of the topics listed here. (It even has [an RSS feed](/feeds/extensions-news.xml).)
 
+### Chrome 112: Increased storage.session quota {: #m112-storage-session-quota}
+
+March 1, 2023
+
+From Chrome 112, the quota for the [`storage.session`](/docs/extensions/reference/storage/#property-session) API has been increased to approximately 10 MB. This was discussed in the Web Extensions Community Group: [https://github.com/w3c/webextensions/issues/350](https://github.com/w3c/webextensions/issues/350)
```
jpmedley, line 23:
```suggestion
From Chrome 112, the quota for the [`storage.session`](/docs/extensions/reference/storage/#property-session) property has been increased to approximately 10 MB. This was discussed in the Web Extensions Community Group: [https://github.com/w3c/webextensions/issues/350](https://github.com/w3c/webextensions/issues/350)
```

jpmedley: This is a best practice to use definite, specific, concrete language.

jpmedley: >This was discussed in the Web Extensions Community Group

Was there a resolution to the discussion? I see that other vendors are adding to storage, but I can't find a PR to update the spec. The future we're headed toward is what happens on the rest of the DevRel team: everything is discussed in the standards committee so we only call it out when there's a resolution or some reason to draw attention to a specific discussion.

jpmedley (changes requested):

site/en/docs/extensions/whatsnew/index.md
oliverdunk: Yes, increasing the limit to 10 MB was the resolution in this case. Unfortunately there isn't a spec for WebExtensions - the goal of the community group in the short term is just to align as best as possible. I wanted to call this out specifically because it's a nice example of us using the group to guide our direction.

site/en/docs/extensions/reference/storage/index.md
oliverdunk: I actually copied this from a few lines above! But I see we don't make this mistake anywhere else so I've fixed it in both places.

site/en/docs/extensions/whatsnew/index.md
jpmedley: Then I think the statement needs to be more positive, something along the lines of 'vendors agreed to increase storage.' (I'm just trying to give you s sense. Word smith it your way.) One word of caution: if any of the vendors who were present at the meeting didn't agree, don't use phrasing that implies that they did.

jpmedley (approved):

site/en/docs/extensions/whatsnew/index.md
oliverdunk: That makes sense. Thanks for the advice and approval!

PR #5513, opened 2023-03-02
Fix linkedin icon URL

beaufortfrancois:
@matthiasrohmer Please review.


matthiasrohmer (approved):
Ugh... and I even noticed and praised you removed the URL in the authors data 😄 Sorry that I let it slip!

PR #5524, opened 2023-03-03
Whats new in Chrome 111

tropicadri:


site/en/blog/new-in-chrome-111/index.md
```diff
@@ -0,0 +1,155 @@
...
+```
+
+
+When `.startViewTransition()` is called, the API captures the current state of the page.
+
+Once that's complete, the `callback` passed to `.startViewTransition()` is called. That's where the DOM is changed. Then, the API captures the new state of the page.
+
+Note that the API is launched for Single-Page Apps (SPAs) but support for other models is being implemented too.
+
+There are many details to this API,[this article](/docs/web-platform/view-transitions/) contains samples and details to create your own custom transitions.
```
rachelandrew, line 66:
```suggestion
There are many details to this API, learn more [in our article containing samples and details](/docs/web-platform/view-transitions/), or explore the [View Transitions documentation on MDN](https://developer.mozilla.org/en-US/docs/Web/API/View_Transitions_API).
```

rachelandrew (approved):
Good to go, I just made one suggestion as we now have docs on MDN for View Transitions.

rachelandrew (approved):

PR #5535, opened 2023-03-03
[CRX] Update offscreen reference doc

sebastianbenz:
Explain how to check for an existing offscreen document.

### DIFF
--- a/site/en/docs/extensions/reference/offscreen/index.md
+++ b/site/en/docs/extensions/reference/offscreen/index.md
@@ -46,7 +46,7 @@ chrome.offscreen.createDocument({
 chrome.offscreen.closeDocument()
 ```
 
-The following example shows how to check for existing offscreen documents. The `hasOffscreenDocument` function uses [clients.matchAll()](https://developer.mozilla.org/docs/Web/API/Clients/matchAll) to find existing offscreen documents. Please note: the function assumes that your extension uses a single offscreen document. 
+The following example shows how to check for existing offscreen documents. The `hasOffscreenDocument()` function uses [clients.matchAll()](https://developer.mozilla.org/docs/Web/API/Clients/matchAll) to find existing offscreen documents. Please note: the function assumes that your extension uses a single offscreen document. 
 
 ```js
 async function hasOffscreenDocument(path) {
--- a/site/en/docs/extensions/reference/offscreen/index.md
+++ b/site/en/docs/extensions/reference/offscreen/index.md
@@ -50,7 +50,7 @@ The following example shows how to check for existing offscreen documents. The `
 
 ```js
 async function hasOffscreenDocument(path) {
-  // Check all windows controlled by the service worker if one 
+  // Check all windows controlled by the service worker to see if one 
   // of them is the offscreen document with the given path
   const offscreenUrl = chrome.runtime.getURL(path);
   const matchedClients = await clients.matchAll();
--- a/site/en/docs/extensions/reference/offscreen/index.md
+++ b/site/en/docs/extensions/reference/offscreen/index.md
@@ -34,7 +34,7 @@ Reasons, listed [below][offscreen-reason], are set upon document creation to det
 
 ## Example
 
-Use `chrome.offscreen.createDocument(...)` and `chrome.offscreen.closeDocument()` for creating and closing an offscreen document. Only a single Document can be open at a time. 
+Use `chrome.offscreen.createDocument()` and `chrome.offscreen.closeDocument()` for creating and closing an offscreen document. Only a single Document can be open at a time. 
 
 ```js
 chrome.offscreen.createDocument({
--- a/site/en/docs/extensions/reference/offscreen/index.md
+++ b/site/en/docs/extensions/reference/offscreen/index.md
@@ -32,9 +32,9 @@ Reasons, listed [below][offscreen-reason], are set upon document creation to det
 | AUDIO_PLAYBACK    | Closed after 30 seconds without audio playing. |
 | All other reasons | Not currently limited                          |
 
-## Examples
+## Example
 
-The following methods create and close an offscreen document. Only a single Document can be open at a time. 
+Use `chrome.offscree.createDocument(...)` and `chrome.offscreen.closeDocument()` for creating and closing an offscreen document. Only a single Document can be open at a time. 
 
 ```js
 chrome.offscreen.createDocument({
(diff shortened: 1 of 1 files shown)


site/en/docs/extensions/reference/offscreen/index.md
```diff
@@ -32,9 +32,9 @@ Reasons, listed [below][offscreen-reason], are set upon document creation to det
 | AUDIO_PLAYBACK    | Closed after 30 seconds without audio playing. |
 | All other reasons | Not currently limited                          |
 
-## Examples
+## Example
 
-The following methods create and close an offscreen document. Only a single Document can be open at a time. 
+Use `chrome.offscree.createDocument(...)` and `chrome.offscreen.closeDocument()` for creating and closing an offscreen document. Only a single Document can be open at a time.
```
oliverdunk, line 37:
```suggestion
Use `chrome.offscreen.createDocument(...)` and `chrome.offscreen.closeDocument()` for creating and closing an offscreen document. Only a single Document can be open at a time. 
```

site/en/docs/extensions/reference/offscreen/index.md
```diff
@@ -46,11 +46,50 @@ chrome.offscreen.createDocument({
 chrome.offscreen.closeDocument()
 ```
 
-For a complete extension, see the [offscreen-clipboard demo][gh-offscreen-clipboard] on GitHub.
+The following example shows how to check for existing offscreen documents. The `hasOffscreenDocument` method uses [clients.matchAll()](https://developer.mozilla.org/docs/Web/API/Clients/matchAll) to find existing offscreen documents.
```
oliverdunk, line 49:
```suggestion
The following example shows how to check for existing offscreen documents. The `hasOffscreenDocument` function uses [clients.matchAll()](https://developer.mozilla.org/docs/Web/API/Clients/matchAll) to find existing offscreen documents.
```

site/en/docs/extensions/reference/offscreen/index.md
```diff
@@ -46,11 +46,50 @@ chrome.offscreen.createDocument({
...
 
-[api-runtime]: /docs/extensions/reference/runtime/
-[api-windows]: /docs/extensions/reference/windows/
-[doc-manifest]: /docs/extensions/mv3/manifest/
-[gh-offscreen-clipboard]: https://github.com/GoogleChrome/chrome-extensions-samples/tree/main/functional-samples/cookbook.offscreen-clipboard-write
-[offscreen-reason]: /docs/extensions/reference/offscreen/#type-Reason
+```js
+async function hasOffscreenDocument(path) {
+  // Check all windows controlled by the service worker if one 
+  // of them is the offscreen document with the given path
```
oliverdunk, line 54: Maybe something like this?

```
// Check to see if there is a context associated with the service worker matching the expected path.
```

I'm reluctant to say `check all windows` because I think clients can also be workers etc.

site/en/docs/extensions/reference/offscreen/index.md
```diff
@@ -46,11 +46,50 @@ chrome.offscreen.createDocument({
...
+  for (const client of matchedClients) {
+    if (client.url === offscreenUrl) {
+      return true;
+    }
+  }
+  return false;
+}
+```
+
+Before trying to create a new offscreen document, use `hasOffscreenDocument` to make sure that there is no existing offscreen document, as demonstrated in the following example.
```
oliverdunk, line 66: Do you think we need to clarify that this wouldn't work if you had two different paths for different use cases?

oliverdunk (changes requested):
Thanks for documenting this! Left a couple of thoughts but overall I think this is a good thing to have. May be worth checking with engineering to be sure we intend for this to keep working and are happy with it as the suggested approach?

site/en/docs/extensions/reference/offscreen/index.md
sebastianbenz: You can filter by type of client and the default is `window` https://developer.mozilla.org/en-US/docs/Web/API/Clients/matchAll

oliverdunk: Oh, TIL!

site/en/docs/extensions/reference/offscreen/index.md
sebastianbenz: fixed

site/en/docs/extensions/reference/offscreen/index.md
sebastianbenz: good catch!

site/en/docs/extensions/reference/offscreen/index.md
sebastianbenz: I was wondering about this as well, the problem is that when you're using different docs this become quickly very complex and there is no good default approach for handling this. I've added a sentence mentioning that we assume only a single document.

sebastianbenz:
Thanks for the review! I've asked Devlin, but haven't heard back - which I take as an OK. The problem is that there is no other good out of the box alternative.

site/en/docs/extensions/reference/offscreen/index.md
```diff
@@ -32,9 +32,9 @@ Reasons, listed [below][offscreen-reason], are set upon document creation to det
 | AUDIO_PLAYBACK    | Closed after 30 seconds without audio playing. |
 | All other reasons | Not currently limited                          |
 
-## Examples
+## Example
 
-The following methods create and close an offscreen document. Only a single Document can be open at a time. 
+Use `chrome.offscreen.createDocument(...)` and `chrome.offscreen.closeDocument()` for creating and closing an offscreen document. Only a single Document can be open at a time.
```
jpmedley, line 37:
```suggestion
Use `chrome.offscreen.createDocument()` and `chrome.offscreen.closeDocument()` for creating and closing an offscreen document. Only a single Document can be open at a time. 
```

site/en/docs/extensions/reference/offscreen/index.md
```diff
@@ -46,11 +46,50 @@ chrome.offscreen.createDocument({
 chrome.offscreen.closeDocument()
 ```
 
-For a complete extension, see the [offscreen-clipboard demo][gh-offscreen-clipboard] on GitHub.
+The following example shows how to check for existing offscreen documents. The `hasOffscreenDocument` function uses [clients.matchAll()](https://developer.mozilla.org/docs/Web/API/Clients/matchAll) to find existing offscreen documents. Please note: the function assumes that your extension uses a single offscreen document.
```
jpmedley, line 49:
```suggestion
The following example shows how to check for existing offscreen documents. The `hasOffscreenDocument()` function uses [clients.matchAll()](https://developer.mozilla.org/docs/Web/API/Clients/matchAll) to find existing offscreen documents. Please note: the function assumes that your extension uses a single offscreen document. 
```

site/en/docs/extensions/reference/offscreen/index.md
```diff
@@ -46,11 +46,50 @@ chrome.offscreen.createDocument({
...
+  for (const client of matchedClients) {
+    if (client.url === offscreenUrl) {
+      return true;
+    }
+  }
+  return false;
+}
+```
+
+Before trying to create a new offscreen document, use `hasOffscreenDocument` to make sure that there is no existing offscreen document, as demonstrated in the following example.
```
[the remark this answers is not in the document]
jpmedley: >use > call

This is an application of the best practice to use specific and concrete language.

sebastianbenz:
@jpmedley thanks for the review! Committed your suggestions.

jpmedley (approved):

PR #5538, opened 2023-03-03
Hosting extensions OS update

dcoollx:
It is not clear that this is an impossibility except for in extremely limited circumstances, such as Linux users and enterprise users.

- updated this article to explicitly say that self-hosting is not an options.

### DIFF
--- a/site/en/docs/extensions/mv3/hosting/index.md
+++ b/site/en/docs/extensions/mv3/hosting/index.md
@@ -45,8 +45,7 @@ and automatically updates them without user intervention.
 development process.
 
 {% Aside %}
-Windows/MacOs users *can not* install packed extensions not distributed by Chrome, except in an enterprise environment
-Linux users can manually install packed extensions that are not distributed or signed by Chrome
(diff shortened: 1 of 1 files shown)


site/en/docs/extensions/mv3/hosting/index.md
```diff
@@ -45,6 +45,7 @@ and automatically updates them without user intervention.
 development process.
 
 {% Aside %}
+Windows/MacOs users *can not* install packed extensions not distributed by Chrome, except in an enterprise environment
 Linux users can manually install packed extensions that are not distributed or signed by Chrome
```
AmySteam, line 49:
```suggestion
Windows/macOS users can only install self-hosted extensions through enterprise policies. However, Linux users can manually install packed extensions that are not distributed or signed by Chrome
```

AmySteam (changes requested):
@dcoollx Thanks for this PR. I tweaked the wording for clarity.

AmySteam (approved):

PR #5552, opened 2023-03-06
Adds a new component: tag-pill-list

johnrich85:

When the users applies filters on the events page, their selections need to render as tag-pills, as per the design below. This PR introduces a new data driven tag-pill-list component which renders the tags.

image

### DIFF
--- a/site/_js/web-components/tag-pill-list.js
+++ b/site/_js/web-components/tag-pill-list.js
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022 Google LLC
+ * Copyright 2023 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/site/_js/web-components/tag-pill-list.js
+++ b/site/_js/web-components/tag-pill-list.js
@@ -39,7 +39,7 @@ export class TagPillList extends BaseElement {
     const items = this.items.map(
       item => html`
         <span
-          class="surface color-blue-medium hairline rounded-lg tag-pill type--label display-inline-flex align-center"
+          class="surface hairline rounded-lg tag-pill type--label display-inline-flex align-center"
           data-key="${item.key}"
           data-value="${item.value}"
         >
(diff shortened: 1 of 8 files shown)


site/_js/web-components/tag-pill-list.js
```diff
@@ -0,0 +1,55 @@
...
+
+  render() {
+    const items = this.items.map(
+      item => html`
+        <span
+          class="surface hairline rounded-lg tag-pill type--label display-inline-flex align-center"
+          data-key="${item.key}"
+          data-value="${item.value}"
+        >
+          ${item.value} ${this.icon?.cloneNode(true)}
```
devnook, line 46: I don't see any interactivity in this element. Why is it a web component and not just markup?
We aim to ship JS only if needed. It looks like here it is used to render some static HTML - we surely don't need Lit nor web components for that?

devnook (changes requested):

site/_js/web-components/tag-pill-list.js
johnrich85: So I'm using it for reactivity in this instance. If we don't use a WC then we'll need to render/append to the dom using JS anyway but happy to refactor to that if that's the preference

devnook: I see, to make the list of pills dynamic? Can you make the example in the docs reactive then?
Also, I think the 'close' icon should do something.

site/_js/web-components/tag-pill-list.js
```diff
@@ -0,0 +1,65 @@
...
+      items: {type: Array, reflect: true},
+    };
+  }
+
+  _handleClick(item) {
+    const event = new CustomEvent('remove-pill', {
+      detail: item,
+    });
+
+    this.dispatchEvent(event);
```
devnook, line 44: I think this should just remove the pill, rather than fire an event that has to be handled outside of the component. After all, 1. removing the pill influences the internal state of the component at hand. 2. Pill-removing logic would need to be reimplemented on each page that wants to use the pill-list.

johnrich85: Granted re #2 but isn't this standard practice with one-way binding? If we mutate the prop internally parent components wouldn't be aware of the change...

devnook: Are there parent components usually? This is not a react app.

devnook (approved):

Issue #5436, opened 2023-02-24
Split main.css on template basis for quicker purging


matthiasrohmer:
The site has three main CSS build targets: `main.css`, `events-main.css`, `chrome-main.css`. While the later two are specialized stylesheets for the events (/meet-the-team) and the Chrome 100 page (/100) the `main.css` file powers the styling for all other pages of the site.

As it weighs over 200KB unminified and uncompressed this amount of CSS slows down some parts of the build pipeline, the CSS purging process specifically which removes unused CSS from pages whose execution time is dependent on the amount of CSS given in.

Therefore we should try to further split up `main.css` into specific stylesheets for our content types. Stylesheets that *could make sense* are `landing-page.scss`, `fugu-showcase.scss`, `homepage.scss`, `article.scss` (powering both blog articles and docs). Please check those assumptions before getting started.

Those specific stylesheets should only include the blocks/components and utilities that are actually required or can be used on the matching templates.

simgobb:
After a quick document proposal, the idea is to split `main.css` into the following: 

- `home.css`
- `fugu.css`
- `landing.css`
- `overview.css`
- `single-post.css`

To make the implementation and testing easier, it's been decided to split this issue into smaller ones, one for each of the new stylesheets.

mollietokyo:
spoke with Matthias, he's waiting on internal updates prior to marking this one as 'done'

matthiasrohmer:
Fixed up #5554 and merged. We now got distinct CSS files per template/page type

PR #5554, opened 2023-03-06
5436 home css

simgobb:
Linked to #5436 

- Created a stylesheet specific for the homepage
- Created a stylesheet with the blocks needed on every page
- Moved the main.css import from the base layout to the specific ones
- Imported the homepage styles in the homepage template


site/_scss/home.scss
```diff
@@ -0,0 +1,36 @@
...
+@import 'globals/mixins';
+@import 'globals/gorko';
+@import 'globals/extends';
+@import 'globals/common-blocks';
+
+@import 'fonts';
+@import 'theme';
+@import 'reset';
+@import 'global';
+@import 'typography';
```
matthiasrohmer, line 16: Sorry I am now going to ask for quite an operation here: we'll most likely use this block across all templates, no? So I would move _fonts.scss, _theme.scss, _reset.scss, and _typography.scss into the globals directory to and then add this block to _global.scss and then only import this here. I think this would make things clearer.

site/_scss/globals/_common-blocks.scss
```diff
@@ -0,0 +1,19 @@
+// Common Blocks
```
matthiasrohmer, line 1: Let's create a new directory block-groups and put this inside there, so `block-groups/_common.scss` and expand the comment to explain why this selection of blocks is considered common.

matthiasrohmer (commented):
Finally got to this, sorry. This looks most likely correct, just some style comments 🙂

matthiasrohmer (approved):

PR #5564, opened 2023-03-07
Added "Copy Code Snippet" Button

### LINKED ISSUE #5296: [Feature Request] Copy Code Snippet Button
## **Is your feature request related to a problem? Please describe.**

The code snippet/examples given in the guides are fantastic. It would be super useful to have a way to copy the code snippets with a single button, rather than manually selecting the contents.

## **Describe the solution you'd like**

A simple "copy to clipboard" square icon-button on the top right of code snippets (like in markdown files on GitHub) would be great. It's a popular feature on several sites so it'll be convenient to have it here as well, and shouldn't be too difficult to implement.

## Relevant Screenshot

| snippet |
|:-:|
| Simple boilerplate code snippets like these are great, and there should be an easy way to copy the contents |

MaAnCoSa:
Fixes #5296

- A button was added in the top right corner of all code snippets (specifically, every 'pre' block in the page).
- The script that creates the button was added to the _shortcode directory.
- The shortcode script is added at the end of the body in the base.njk layout. That way all 'pre' blocks on the page have already been created by the time the scripts runs, pulling each block and adding the button.

Signed the Individual CLA as MaAnCoSa (Manuel Andrés Cota Santeliz).

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -41,6 +41,7 @@ const {LanguageList} = require('./site/_shortcodes/LanguageList');
 const {Partial} = require('./site/_shortcodes/Partial');
 const {ChromeDate} = require('./site/_shortcodes/ChromeDate');
 const {BrowserCompat} = require('webdev-infra/shortcodes/BrowserCompat');
+const {CopyCodeButton} = require('./site/_shortcodes/CopyCodeButton');
 
 // Transforms
 const {domTransformer} = require('./site/_transforms/dom-transformer-pool');
@@ -155,6 +156,7 @@ module.exports = eleventyConfig => {
   eleventyConfig.addShortcode('BrowserCompat', BrowserCompat);
   eleventyConfig.addNunjucksAsyncShortcode('Partial', Partial);
   eleventyConfig.addShortcode('ChromeDate', ChromeDate);
+  eleventyConfig.addShortcode('CopyCodeButton', CopyCodeButton);
 
   // Empty shortcodes. They are added for backward compatibility with web.dev.
   // They will not render any html, but will prevent the build from failing.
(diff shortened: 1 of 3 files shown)


MaAnCoSa:
@matthiasrohmer can you review this PR please?

matthiasrohmer (changes requested):
Hi @MaAnCoSa, thanks for working on this. Unfortunately this approach will not work.

Using a shortcode here is an additional layer that is not needed. You could have also directly pasted the script into base.njk, or used a simple `{% include %}`. But that's the wrong place for client-side logic anyway.

This should be done as a web component in site/_js/web-components. Have a look at the web component for web.dev, which uses a similar architecture: https://github.com/GoogleChrome/web.dev/blob/main/src/lib/components/CopyCode/index.js

This web component would then need to be injected into code blocks by altering the markdown rendering behaviour. Again, this is how it is done on web.dev: https://github.com/GoogleChrome/web.dev/blob/d99e7ad4eb1ce64d1079112b9d7ed24243524cb9/src/site/_plugins/markdown.js#L52-L57

Some feedback on your existing work though: you are using inline CSS but for this project CSS for components lives in site/_scss/blocks. There are some console.log debug statements left, which shouldn't go into production.

Instead of `onclick` always make sure to use `addEventListener` instead, to not overwrite existing handlers and being able to detach them.

You are also using three hard coded URLs to icons8.com, which is not ideal. That's probably not covered by their terms but also all our icons are usually SVGs (to have them sharp across all devices) and are inlined to the HTML document.

developer.chrome.com is not an easy site to get started with and the knowledge you gain from working on it is very specific and doesn't apply to a lot of projects, so you might probably have more fun working on another project 🙂 Especially since we can basically copy over what we did for web.dev.

Once again thanks for your contribution, I appreciate the work you put into this, but in this state it won't work for the production site.

PR #5584, opened 2023-03-08
Add noise metrics to noiselab page

cucuandreea:
https://deploy-preview-5584--developer-chrome-com.netlify.app/en/docs/privacy-sandbox/summary-reports/design-decisions/

-Added noise metrics section

### DIFF
--- a/site/en/docs/privacy-sandbox/summary-reports/design-decisions/index.md
+++ b/site/en/docs/privacy-sandbox/summary-reports/design-decisions/index.md
@@ -682,8 +682,10 @@ In Noise Lab, this is then multiplied by 100 to give a percentage.
 
 RMSPE_T is a bit more complex to grasp than APE. However, it has a few advantages that make it in some cases more suitable than APE for analyzing noise in summary reports:
 
-* **It's more stable.** *"T"* is a threshold: it's used to give less weight, in the RMSPE_T calculation, to buckets that have less conversions and are therefore more sensitive to noise due to their small size. With T, the metric does not spike on buckets with few conversions. If T is equal to 5, a noise value as small as 1 on a bucket with 0 conversion will not be displayed as way over 100%. Instead, it will be capped at 20%, which is equivalent to 1/5, as T is equal to 5. By giving less weight to buckets that are more sensitive to noise due to their small size, this metric is more stable — and therefore makes it easier to compare two simulations.
-* **It allows for easy aggregation.** Knowing RMSPE_t of multiple buckets, together with their true counts, allows one to compute the RMSPSE_t of their sum via a simple formula. This also allows one to easily optimize for RMSPE_t of these combined values. (While aggregation is possible for APE, the formula is quite complicated since it involves the absolute value of sum of Laplace noises, which makes it hard to optimize.) 
+* **RMSPE_T is more stable.** "T" is a threshold. "T" is used to give less weight in the RMSPE_T calculation to buckets that have less conversions and are therefore more sensitive to noise due to their small size. With T, the metric does not spike on buckets with few conversions. If T is equal to 5, a noise value as small as 1 on a bucket with 0 conversions will not be displayed as way over 100%. Instead, it will be capped at 20%, which is equivalent to 1/5, as T is equal to 5. By giving less weight to smaller buckets which are therefore more sensitive to noise, this metric is more stable, and therefore makes it easier to compare two simulations.
+* **RMSPE_T allows for easy aggregation.** Knowing the RMSPE_T of multiple buckets, together with their true counts, allows you to compute the RMSPE_T of their sum. This also allows you to optimize for RMSPE_T for these combined values.
+
+While aggregation is possible for APE, the formula is quite complicated as it involves the absolute value of sum of Laplace noises. This makes APE harder to optimize.
 
 
 ##### Code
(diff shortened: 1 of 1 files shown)


site/en/docs/privacy-sandbox/summary-reports/design-decisions/index.md
```diff
@@ -578,6 +578,170 @@ Click on the buttons in the top menu to toggle between the two modes (_#1. in th
...
+automatically by Noise Lab, to allow individuals to "hide in the crowd," or in
+other words, protects these limited individuals' privacy with a larger amount
+of added noise.
+
+A low noise percentage indicates that the data setup was designed in such
+a way that already allows individuals to "hide in the crowd." This means the
+buckets contain contributions from a sufficient number of events to ensure that
+individual user privacy is protected.
+
+This statement holds true for both APE and RMSPE_T.
```
heyawhite, line 602:
```suggestion
This statement holds true for both the [APE](#ape) (average percentage error) and [RMSPE_T](#rmspe) (root-mean-square percentage error with a threshold).
```

site/en/docs/privacy-sandbox/summary-reports/design-decisions/index.md
```diff
@@ -581,6 +581,171 @@ Click on the buttons in the top menu to toggle between the two modes (_#1. in th
...
+
+{% Aside %}
+What does it mean if RMSPE_T is above 100%?
+The interpretation of  RMSPE_T over 100% is similar to that of an APE over 100% LINK.
+
+RMSPE (Root Mean Square Error) is a common metricused by data scientists. 
+RMSPE_T is a variation of RMSPE. RMSPE_T differs from RMSPE in two ways:
+
+1. It uses a percentage error similar to noise-to-signal rate, like noise/true-value. 
+2. It assumes the existence of minimal signal: the percentage looks like noise/max(true, T).
```
heyawhite, line 664:
```suggestion
1. RMSPE_T uses a percentage error similar to noise-to-signal rate, like noise/true-value. 
2. RMSPE_T assumes the existence of minimal signal: the percentage looks like noise/max(true, T). 
```

This assumes "it" is RMSPE_T.

site/en/docs/privacy-sandbox/summary-reports/design-decisions/index.md
```diff
@@ -581,6 +581,171 @@ Click on the buttons in the top menu to toggle between the two modes (_#1. in th
...
+
+* **It's more stable.** *"T"* is a threshold: it's used to give less weight, in the RMSPE_T calculation, to buckets that have less conversions and are therefore more sensitive to noise due to their small size. With T, the metric does not spike on buckets with few conversions. If T is equal to 5, a noise value as small as 1 on a bucket with 0 conversion will not be displayed as way over 100%. Instead, it will be capped at 20%, which is equivalent to 1/5, as T is equal to 5. By giving less weight to buckets that are more sensitive to noise due to their small size, this metric is more stable — and therefore makes it easier to compare two simulations.
+* **It allows for easy aggregation.** Knowing RMSPE_t of multiple buckets, together with their true counts, allows one to compute the RMSPSE_t of their sum via a simple formula. This also allows one to easily optimize for RMSPE_t of these combined values. (While aggregation is possible for APE, the formula is quite complicated since it involves the absolute value of sum of Laplace noises, which makes it hard to optimize.)
```
heyawhite, line 686:
```suggestion
* **RMSPE_T is more stable.** "T" is a threshold. "T" is used to give less weight in the RMSPE_T calculation to buckets that have less conversions and are therefore more sensitive to noise due to their small size. With T, the metric does not spike on buckets with few conversions. If T is equal to 5, a noise value as small as 1 on a bucket with 0 conversions will not be displayed as way over 100%. Instead, it will be capped at 20%, which is equivalent to 1/5, as T is equal to 5. By giving less weight to smaller buckets which are therefore more sensitive to noise, this metric is more stable, and therefore makes it easier to compare two simulations.
* **RMSPE_T allows for easy aggregation.** Knowing the RMSPE_T of multiple buckets, together with their true counts, allows you to compute the RMSPE_T of their sum. This also allows you to optimize for RMSPE_T for these combined values.

While aggregation is possible for APE, the formula is quite complicated as it involves the absolute value of sum of Laplace noises. This makes APE harder to optimize.
```

site/en/docs/privacy-sandbox/summary-reports/design-decisions/index.md
```diff
@@ -581,6 +581,171 @@ Click on the buttons in the top menu to toggle between the two modes (_#1. in th
...
+
+bucket_2 = noise: 20, trueSummaryValue: 100
+
+bucket_3 = noise: 20, trueSummaryValue: 200
+
+APE = (0.1 + 0.2 + 0.1) / 3 = **13%**
+
+RMSPE_T = sqrt( ( (10/max(5,100))^2  + (20/max(5,100))^2 + (20/max(5,200))^2) / 3) 
+
+      =  sqrt( (0.01 + 0.04 + 0.01) / 3) =  0.14 =  **14%**
```
heyawhite, line 709: This looks the same as `RMSPE_T = sqrt( ( (10/max(5,100))^2  + (20/max(5,100))^2 + (20/max(5,200))^2) / 3) =  sqrt( (0.01 + 0.04 + 0.01) / 3) =  0.14 =  **14%** `

Should these equations be on separate lines or be on one line?

heyawhite (commented):
I've pushed some copy/style nits, but the rest of the feedback/suggestions requires clarification.

site/en/docs/privacy-sandbox/summary-reports/design-decisions/index.md
cucuandreea: I just added  a new line because I didn't like where it was breaking the line, but normally it is the same equation and should be one line only.

heyawhite (approved):

PR #5591, opened 2023-03-09
Explain that Native Messaging is not available from content scripts

ericlaw1979:
We don't seem to document this anywhere obvious, requiring users to either find the StackOverflow answer or read the Chromium source code.

https://chromium.googlesource.com/chromium/src/+/main/chrome/common/extensions/api/_features.md#Feature-Contexts

https://source.chromium.org/chromium/chromium/src/+/main:extensions/common/api/_api_features.json

  "runtime.connectNative": {
    "dependencies": ["permission:nativeMessaging"],
    "contexts": ["blessed_extension"]
  },

  "runtime.sendNativeMessage": {
    "dependencies": ["permission:nativeMessaging"],
    "contexts": ["blessed_extension"]
  },

The blessed_extension context refers to a JavaScript context running from an extension process. These are typically the most secure JavaScript contexts, as it reduces the likelihood that a compromised web page renderer will have access to secure APIs.

The content_script context refers to a JavaScript context for an extension content script. Since content scripts share a process with (and run on the same content as) web pages, these are considered very insecure contexts. Very few features should be exposed to these contexts.

### DIFF
--- a/site/en/docs/extensions/mv3/nativeMessaging/index.md
+++ b/site/en/docs/extensions/mv3/nativeMessaging/index.md
@@ -120,8 +120,8 @@ messaging. The main difference is that [`runtime.connectNative`][connect-native]
 To use these methods, the "nativeMessaging" permission must be [declared][declare-permissions] in your
 extensions's manifest file.
 
-These methods are not available inside content scripts, only inside your extension's pages and ServiceWorker. If you
-wish to communicate from a content script to the native application, send the message to your background script page
+These methods are not available inside content scripts, only inside your extension's pages and service worker. If you
+wish to communicate from a content script to the native application, send the message to your background context
 or service worker, and have it pass it along to the native application.
 
 The following example creates a [`runtime.Port`][port] object that's connected to native messaging host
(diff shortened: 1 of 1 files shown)


site/en/docs/extensions/mv3/nativeMessaging/index.md
```diff
@@ -116,9 +116,14 @@ Sending and receiving messages to and from a native application is very similar
 messaging. The main difference is that [`runtime.connectNative`][connect-native] is used instead of
 [`runtime.connect`][connect], and [`runtime.sendNativeMessage`][send-native-message] is used instead of
 [`runtime.sendMessage`][send-message].
+
 To use these methods, the "nativeMessaging" permission must be [declared][declare-permissions] in your
 extensions's manifest file.
 
+These methods are not available inside content scripts, only inside your extension's pages and ServiceWorker. If you
+wish to communicate from a content script to the native application, send the message to your background script page
```
oliverdunk, line 124: Could we replace "ServiceWorker" with "service worker", and "background script page" with "background context"? I think that lines up nicer with our usual phrasing.

oliverdunk (changes requested):
Left one comment but overall I think this is a nice improvement! Thanks for the contribution :)

site/en/docs/extensions/mv3/nativeMessaging/index.md
ericlaw1979: done

oliverdunk (approved):
LGTM, thanks again ❤️

PR #5629, opened 2023-03-10
Update index.md

nberg79:
Proposed additions highlighting SSP actions regarding UACH Phase 6.


heyawhite:
Hi @nberg79!

As this is a blog post and not documentation, we need a new blog post. It's critical to keep older posts to preserve content history. A new post ensures it is published at the appropriate time in the feed, *and* that it has review by our technical writers.

Please file a bug request at go/sppi-words and add that content to a doc we can take a look.

heyawhite:
@mihajlija FYI, I've added to the content calendar

PR #5664, opened 2023-03-14
Adding Q&A for FLEDGE

Svenmay:
Adding specific question & answer for "How can a user block an ad when it's shown as a result of a FLEDGE auction?"

Fixes b/263325220


heyawhite:
As the FAQ page no longer exists in the original format, I've integrated this suggestion to the FLEDGE overview.

heyawhite:
@Svenmay is this no longer a necessary change? Can you give more context as to why this was closed?

heyawhite:
@Svenmay I could have been clearer! Thanks for clarifying :) Will merge now

PR #5667, opened 2023-03-14
Find broken URLs using webdev-infra's UrlCrawl

### LINKED ISSUE #46: Typo in docs [https://developer.chrome.com/docs/extensions/#modernize-your-extensions]
Same Content is Referenced Twice [Here](https://developer.chrome.com/docs/extensions/#modernize-your-extensions) With Title **Migrating from background pages to service workers**

Blazzike:
Fixes #46

- Uses to-be-updated webdev-infra's UrlCrawl feature to find broken HREFs via Github Action


.github/workflows/check.yml
```diff
@@ -225,3 +225,55 @@ jobs:
...
+              - 'site/_plugins/**.js'
+              - 'site/_scss/**.scss'
+              - 'site/_shortcodes/**.js'
+              - 'site/_transforms/**.js'
+              - 'site/_utils/**.js'
+              - 'site/en/content-types/**'
+
+      - name: Cloning repository
+        uses: actions/checkout@v3
+        if: ${{ steps.changes.outputs.frontend == 'true' }}
```
matthiasrohmer, line 253: We'd likely check links if the pages changed, so site/**/*.md, no?

.github/workflows/check.yml
```diff
@@ -225,3 +225,55 @@ jobs:
         env:
           PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
         run: npm run snapshots
+
+
+  check-links:
```
matthiasrohmer, line 230: Since the file is already named `check`, probably just `links`.

package.json
```diff
@@ -39,7 +39,8 @@
     "version-check": "npm run sync-external && node version-check.js",
     "build-external": "node ./external/build-external.js",
     "maybe-sync-external": "node ./external/maybe-sync-external.js",
-    "sync-external": "node ./external/sync-external.js"
+    "sync-external": "node ./external/sync-external.js",
+    "check-links": "node ./gulp-tasks/check/links.js"
```
matthiasrohmer, line 43: Rather export the task from the main gulpfile, so one is able to just run `npx gulp checkLinks`. In an effort to reduce the number of NPM scripts.

gulp-tasks/check/links.js
```diff
@@ -0,0 +1,14 @@
+const {UrlCrawl} = require('webdev-infra/utils/url-crawl');
+const handlers = require('../../server/handlers');
+
+new UrlCrawl({
```
matthiasrohmer, line 4: I think for this to be a proper gulp task you would need to wrap this in an (async) function and then export it from here. Also move to just gulp-tasks/checkLinks.js, since we only have one check task so far.

PR #5683, opened 2023-03-15
clarify options example file names

mjunder:
The current document doesn't explictly state what to name the example HTML or JS files, or where to place them. A reader coming from the [Development Basics](https://developer.chrome.com/docs/extensions/mv3/getstarted/development-basics/#structure) page might assume that script files should go into a "scripts" directory, or might simply name the files incorrectly. This change clarifies the naming and file placement, without altering the example file code.

### DIFF
--- a/site/en/docs/extensions/mv3/options/index.md
+++ b/site/en/docs/extensions/mv3/options/index.md
@@ -43,7 +43,7 @@ alt="Context Menu Options page", width="357", height="222" %}
 
 ## Write the options page {: #write_page }
 
-Below is an example options page.
+Below is an example options page. Save it as `options.html` in your extension's root folder.
 
 ```html
 <!DOCTYPE html>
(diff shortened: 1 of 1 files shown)


site/en/docs/extensions/mv3/options/index.md
```diff
@@ -72,7 +72,7 @@ Favorite color:
 </html>
 ```
 
-Save a user's preferred options across devices by using the [storage.sync][1] API.
+Below is an example options script. Save it as `options.js` in the same folder as `options.html`. It lets you save a user's preferred options across devices by using the [storage.sync][1] API.
```
AmySteam, line 75:
```suggestion
Below is an example options script. Save it as `options.js` in the same folder as `options.html`. This saves the user's preferred options across devices using the [storage.sync][1] API.
```

AmySteam (changes requested):
Thanks for this PR @mjunder! LGTM 👍 🙂 

Just requesting a minor tweak.

AmySteam (approved):

PR #5695, opened 2023-03-15
Add link to migration guide

jpmedley:
Link to new migration guide.


jpmedley:
@AmySteam, what do you think?

AmySteam (approved):
@jpmedley Good idea! 💯

PR #5702, opened 2023-03-16
Update web_accessible_resources index.md

HotariTobu:
According to

https://developer.chrome.com/docs/extensions/mv3/manifest/web_accessible_resources/

`web_accessible_resources`'s `matches` is:

> An array of strings, each containing a match pattern that specifies which sites can access this set of resources. Only the origin is used to match URLs. Origins include subdomain matching. Paths are ignored.

However, the description on the site differs from the actual behavior of Chrome so I adjusted to it.


oliverdunk (approved):
LGTM 👍 . Arguably we should clarify that only `/*` is supported, but I think this is better than what we had.

PR #5716, opened 2023-03-17
[Launch handler article] Update article now that the API has shipped

### LINKED ISSUE #5710: [launch-handler] Article suggests feature is feature-detectable when it isn't
In https://developer.chrome.com/docs/web-platform/launch-handler/#is-the-launch-handler-api-supported, there is a snippet of code that detects whether the launch handler API is supported:

```js
if ('launchQueue' in window && 'targetURL' in LaunchParams.prototype) {
  // The Launch Handler API is supported.
}
```

However, this is only detecting the existence of `launchQueue`, which may be available on browsers that don't support the Launch Handler API (e.g. it's also used by File Handlers). Notably, Launch Handler is only available on ChromeOS whereas all Chromium browsers expose `launchQueue` (so actually, this snippet incorrectly returns true for the majority of users).

Launch handler is not actually feature detectable. As a manifest feature, it's not something developers can detect ahead of time, you just set it in your manifest and if it works, it works.

**Suggestion**: Remove this section (do not imply that Launch Handler is feature-detectable), but incorporate the null checks into the usage code snippet, as best practice in case the APIs don't exist:

```js
if ('launchQueue' in window) {
  launchQueue.setConsumer((launchParams) => {
    if (launchParams.targetURL) {
      const songID = extractSongId(launchParams.targetURL);
      if (songID) {
        playSong(songID);
      }
    }
  });
}
```

tomayac:
Fixes #5710

- Remove OT leftovers.
- Slight restructuring.
- Add spec link.

### DIFF
--- a/site/en/docs/web-platform/launch-handler/index.md
+++ b/site/en/docs/web-platform/launch-handler/index.md
@@ -7,7 +7,7 @@ description: >
 subhead: >
   Control how your app is launched.
 date: 2021-12-14
-updated: 2022-08-04
+updated: 2023-03-17
 authors:
   - thomassteiner
 ---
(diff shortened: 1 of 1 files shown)


tomayac:
FYI @mgiuca. Thanks for the ping about this.

rachelandrew (approved):

PR #5719, opened 2023-03-17
Correct typo in "flags"

rowan-m:
Minuscule change.


heyawhite (approved):
"flahs" sounds better, but I guess you're right.

rowan-m:
_I was Josh Safdie’s muse when he wrote Uncuh Jahmz_

PR #5750, opened 2023-03-21
Add examples to printing APIs.

jpmedley:
Fixes #SOME_ISSUE_NUMBER

-
-
-

### DIFF
--- a/site/en/docs/extensions/reference/printing/index.md
+++ b/site/en/docs/extensions/reference/printing/index.md
@@ -62,18 +62,20 @@ console.log(`Default printer: ${defaultPrinter.name}.\n\t${defaultPrinter.descri
 ### submitJob()
 
 The `submitJob()` method requires three things.
+
 * A `ticket` structure specifying which capabilities of the printer are to be used. If the user needs to select from available capabilities, you can retrieve them for a specific printer using `getPrinterInfo()`.
 * A `SubmitJobRequest` structure, which specifies the printer to use, and the file or date to print. This structure contains a reference to the `ticket` structure. 
 * A blob of the file or data to print.
 
-This is a simplified version of the printing example. Notice that the `ticket` is attached to the `SubmitJobRequest` structure (line 6) and that the data to print is converted to a blob (line 8). 
+This is a simplified version of the printing example. Notice that the `ticket` is attached to the `SubmitJobRequest` structure (line 8) and that the data to print is converted to a blob (line 10). Getting the ID of the printer (line 1) is more complicated [in the sample](https://github.com/GoogleChrome/chrome-extensions-samples/tree/main/api-samples/printing)than is shown here. 
 
-```javascript/6,8
-const ticket = getPrinterTicket(printerId);
+```javascript/0,7,9
+const defaultPrinter = getDefaultPrinter();
+const ticket = getPrinterTicket(defaultPrinter);
 const arrayBuffer = getPrintData();
 const submitJobRequest = {
   job: {
-    printerId: printerId,
+    printerId: defaultPrinter,
     title: 'test job',
     ticket: ticket,
     contentType: 'application/pdf',
--- a/site/en/docs/extensions/reference/printing/index.md
+++ b/site/en/docs/extensions/reference/printing/index.md
@@ -4,7 +4,7 @@ api: printing
 
 ## Manifest {: #manifest }
 
-All `chrome.printing` methods and events require you to declare the "printing" permission in the [extension manifest](/docs/extensions/mv3/manifest/). For example:
+All `chrome.printing` methods and events require you to declare the `"printing"` permission in the [extension manifest](/docs/extensions/mv3/manifest/). For example:
 
 ```json
 {
--- a/site/en/docs/extensions/reference/printing/index.md
+++ b/site/en/docs/extensions/reference/printing/index.md
@@ -66,9 +66,9 @@ The `submitJob()` method requires three things.
 * A `SubmitJobRequest` structure, which specifies the printer to use, and the file or date to print. This structure contains a reference to the `ticket` structure. 
 * A blob of the file or data to print.
 
-This is a simplified version of the printing example. Notice that the `ticket` is attached to the `SubmitJobRequest` structure (line 8) and that the data to print is converted to a blob (line 8). 
+This is a simplified version of the printing example. Notice that the `ticket` is attached to the `SubmitJobRequest` structure (line 6) and that the data to print is converted to a blob (line 8). 
 
-```javascript
+```javascript/6,8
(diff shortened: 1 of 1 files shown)


site/en/docs/extensions/reference/printing/index.md
```diff
@@ -2,4 +2,95 @@
...
+}
+```
+
+## Examples
+
+The examples below demonstrate using each of the methods in the printing namespace. This code is copied from or based on the [printing](https://github.com/GoogleChrome/chrome-extensions-samples/tree/main/api-samples/printing) in the Extension samples database.
+
+### cancelJob()
+
+This example uses the `onJobStatusChanged` handler to hide a 'cancel' button when the `jobStatus` is neither `PENDING` or `IN_PROGRESS`. Note that on some networks or when a Chromebook is connected directly to the printer, these states may pass too quickly for the cancel button to be visible long enough to be called. This is greatly simplified the printing example.
```
AmySteam, line 26:
```suggestion
This example uses the `onJobStatusChanged` handler to hide a 'cancel' button when the `jobStatus` is neither `PENDING` or `IN_PROGRESS`. Note that on some networks or when a Chromebook is connected directly to the printer, these states may pass too quickly for the cancel button to be visible long enough to be called. This is greatly simplified printing example.
```

site/en/docs/extensions/reference/printing/index.md
```diff
@@ -2,4 +2,95 @@
...
+
```
AmySteam, line 27: Should we add [labels](https://developer.chrome.com/docs/handbook/components/#labels) to the code samples?

site/en/docs/extensions/reference/printing/index.md
```diff
@@ -2,4 +2,95 @@
...
+```
+
+### submitJob()
+
+The `submitJob()` method requires three things.
+* A `ticket` structure specifying which capabilities of the printer are to be used. If the user needs to select from available capabilities, you can retrieve them for a specific printer using `getPrinterInfo()`.
+* A `SubmitJobRequest` structure, which specifies the printer to use, and the file or date to print. This structure contains a reference to the `ticket` structure. 
+* A blob of the file or data to print.
+
+This is a simplified version of the printing example. Notice that the `ticket` is attached to the `SubmitJobRequest` structure (line 8) and that the data to print is converted to a blob (line 8).
```
AmySteam, line 69: > Notice that the `ticket` is attached to the `SubmitJobRequest` structure 
> (line 8) and that the data to print is converted to a blob (line 8). 

Can we highlight these in the code sample?

AmySteam (changes requested):
@jpmedley Looks good 👍 . I made a couple of suggestions. 

I don't have a Chromebook so I cannot test that the code works as expected.

I would request an additional tech review from @oliverdunk or @sebastianbenz

site/en/docs/extensions/reference/printing/index.md
```diff
@@ -2,4 +2,95 @@
...
+  const printerInfo = await chrome.printing.getPrinterInfo(printer.id);
+  return printerInfo.isDefault;
+}
+console.log(`Default printer: ${defaultPrinter.name}.\n\t${defaultPrinter.description}`);
+```
+
+### submitJob()
+
+The `submitJob()` method requires three things.
+* A `ticket` structure specifying which capabilities of the printer are to be used. If the user needs to select from available capabilities, you can retrieve them for a specific printer using `getPrinterInfo()`.
```
AmySteam, line 65: > a specific printer using `getPrinterInfo()`.
How does this work? Should we add a link?

Also in the code below says:
> const ticket = getPrinterTicket(printerId);

site/en/docs/extensions/reference/printing/index.md
```diff
@@ -2,4 +2,95 @@
...
+  "name": "My extension",
+  ...
+  "permissions": [
+    "printing"
+  ],
+  ...
+}
+```
+
+## Examples
```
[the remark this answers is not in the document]
jpmedley: I'm committing this, but I need to check something. I think this is just an override. We get an anchor regardless.

site/en/docs/extensions/reference/printing/index.md
```diff
@@ -2,4 +2,95 @@
...
+
+The examples below demonstrate using each of the methods in the printing namespace. This code is copied from or based on the [printing](https://github.com/GoogleChrome/chrome-extensions-samples/tree/main/api-samples/printing) in the Extension samples database.
+
+### cancelJob()
```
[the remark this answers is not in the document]
jpmedley: Why complicate it?

site/en/docs/extensions/reference/printing/index.md
```diff
@@ -2,4 +2,95 @@
...
+    chrome.printing.cancelJob(jobId).then((response) => {
+      if (response !== undefined) {
+        console.log(response.status);
+      }
+      if (chrome.runtime.lastError !== undefined) {
+        console.log(chrome.runtime.lastError.message);
+      }
+    });
+  });
+  if (status !== "PENDDING" && status !== "IN_PROGRESS") {
```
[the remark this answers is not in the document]
jpmedley: Yep.

site/en/docs/extensions/reference/printing/index.md
jpmedley: The label would be 'printers.js', which would need to be explained. I realize that I could say that about sw, content, and popup, but those are standard and it solves itself over time. 'printers.js' is completely unique to this page, so I'm not sure what the label gets us. If you actually go to the sample, it's pretty obvious where to look, even if you didn't know ahead of time.

(I've been thinking for a few days that we should have a meeting about this. Maybe we don't need labels everywhere, but we should have a common heuristic.)

site/en/docs/extensions/reference/printing/index.md
jpmedley: Weird. I thought I had. Done.

site/en/docs/extensions/reference/printing/index.md
jpmedley: I was trying to avoid making this overly complicated. I think I threaded the needle better. Please take a look.

AmySteam (commented):
Thanks @jpmedley.

I'm going to pass this on to a DRE, since I couldn't really verify that the code works without having access to ChromeOS.

 @sebastianbenz can you give this a quick technical review?

jpmedley:
@AmySteam I'm working with an engineer in ChromeOS. He has already verified this.

AmySteam (approved):

jpmedley:
Thank you. There will be more in the future from the ChromeOS folks. I'll make it clear when the content of a PR has been reviewed already.

PR #5773, opened 2023-03-23
Update Private Network Access blog post.

letitz:
Updates the timeline to reflect the latest deprecation trial extension, as well as mention the permission prompt that @iVanlIsh is working on.

### DIFF
--- a/site/en/blog/private-network-access-update/index.md
+++ b/site/en/blog/private-network-access-update/index.md
@@ -11,6 +11,9 @@ hero: image/YLflGBAPWecgtKJLqCJHSzHqe2J2/dwtN0NkxkBmIz1EyhzAm.jpg
 alt: A private sign in German
 tags:
   - chrome-94
+  - chrome-109
+  - chrome-113
+  - chrome-116
   - chrome-117
   - security
 ---
@@ -21,7 +24,7 @@ tags:
   occur until Chrome 117.
 
 - **January 19, 2023**: The timeline has been updated, and deprecation
-   will not occur until Chrome 113.
+   will not occur until Chrome 114.
 
 - **August 12, 2022**: The timeline has been updated, and deprecation
    will not occur until Chrome 109.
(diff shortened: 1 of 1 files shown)


site/en/blog/private-network-access-update/index.md
```diff
@@ -148,15 +155,15 @@ still mention the earlier milestone.
 
 This deprecation is accompanied by a deprecation trial, allowing web developers
 whose websites make use of the deprecated feature to continue using it until
-Chrome 113 by registering for tokens. See [below](#register-deprecation-trial)
+Chrome 117 by registering for tokens. See [below](#register-deprecation-trial)
```
johnathan79717, line 158: This reads to me like they can still use the deprecation feature in Chrome 117.
I looked up the definition of "until" https://ell.stackexchange.com/questions/33340/is-until-inclusive-or-exclusive.  It seems that it's usually inclusive when used positively and exclusive when used negatively.  This is probably why "deprecation will not occur until Chrome 117." sounds like 117 is not included but "continue using it until Chrome 117" sounds like 117 is included.  It is also consistent with how we used it before since in Chrome 113 the deprecation trial was still enabled.
I'm not a native English speaker or a professional UX writer though.

letitz: Ah, interesting! Thanks for pointing that out, TIL. I'll amend to consistently say 116 then, I think that will be less confusing. Note that the previous "timeline" section mistakenly said that the trial would end with the release of 116, though.

johnathan79717: You're right!  So it was consistent but wrong in a more obvious way.  I'm not entirely sure what I said was right though, so feel free to consult someone else 😆

letitz:
Maybe @heyawhite can review this?

heyawhite:
@letitz Unfortunately, this is outside of my coverage area. Please open a bug at go/sppi-words and our team will assign the right technical writer to review.

agektmr:
I can do tomorrow.

agektmr (approved):
LGTM

agektmr:
@heyawhite can you merge please?

anusmitaray (approved):

PR #5781, opened 2023-03-23
5649 deprecation calendar external data flow

### LINKED ISSUE #5649: Implement the external data flow for Chrome deprecations
Follow-up to https://github.com/GoogleChrome/developer.chrome.com/issues/5210, blocked by #5646.

Extend (and probably rename) the existing `chrome-release` external data flow, to fetch the data needed for the the Deprecations Calendar.

An example implementation of what's needed can be found here: https://chrome-status-lite.deno.dev/deprecations (Source: https://github.com/PaulKinlan/ChromeStatusLite/blob/main/src/routes/deprecations.ts). The existing flow is using the same API, thought querying it not that sophisticated, which might make sense given the amount of data we use from it.

simgobb:
Fixes #5649 

- create an external data flow to fetch all the data needed for the deprecation calendar


external/build/deprecation-calendar.js
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * @return {promise} A promise that resolves in an array containing
+ * all the deprecations
+ */
+async function fetchDeprecations() {
+  let deprecations = [];
+  try {
+    const data = await fetch(DEPRECATION_ENDPOINT);
+
+    const deprecationsText = (await data.text()).substring(5);
```
matthiasrohmer, line 21: Add a comment why the substring is needed.

external/build/deprecation-calendar.js
```diff
@@ -0,0 +1,116 @@
...
+  } catch (error) {
+    console.error('Error fetching the channels', error);
+  }
+
+  return channels;
+}
+
+/**
+ * Starting from an array of deprecations, removes the ones that have
+ * no Chrome version associated and them orders them chronologically
```
matthiasrohmer, line 60: Nit: then

matthiasrohmer (commented):
Only had a quick look. This looks about right. To your question how to deal with related articles, I would go with the same approach as for the YouTube stuff, creating a data helper in site/_data that pulls the fetched data from this job as well as a manually maintained yaml that links deprecations to a list of articles. Maybe by deprecation id...?

external/build/deprecation-calendar.js
```diff
@@ -0,0 +1,121 @@
...
+
+const DEPRECATION_ENDPOINT =
+  'https://chromestatus.com/api/v0/features?q=feature_type=3';
+
+const CHANNELS_ENDPOINT = 'https://chromestatus.com/api/v0/channels';
+
+/**
+ * Fetches all the deprecations from a list of features
+ *
+ * @return {promise} A promise that resolves in an array containing
```
matthiasrohmer, line 13: A better type would be `Promise<any[]>`

external/build/deprecation-calendar.js
```diff
@@ -0,0 +1,121 @@
...
+ * all the deprecations
+ */
+async function fetchDeprecations() {
+  let deprecations = [];
+  try {
+    const data = await fetch(DEPRECATION_ENDPOINT);
+
+    const deprecationsText = (await data.text()).substring(5); // Remove first 5 chars to have a correctly formatted JSON
```
matthiasrohmer, line 21: Style nit: move comment above line, but also reuse `let data` for this block instead of defining multiple one-off `const`s.

external/build/deprecation-calendar.js
```diff
@@ -0,0 +1,121 @@
...
+ * @param {number  | undefined} end newest version needed
+ * @return {promise} A promise that resolves in an array containing
+ * all the channel infos needed
+ */
+async function fetchChannels(start, end) {
+  let channels = [];
+  let queryString = '';
+  if (start !== undefined && end !== undefined) {
+    queryString = `?start=${start}&end=${end}`;
+  }
```
matthiasrohmer, line 44: Use the `URLSearchParams` API instead.

external/build/deprecation-calendar.js
```diff
@@ -0,0 +1,121 @@
...
+
+  try {
+    const data = await fetch(`${CHANNELS_ENDPOINT}${queryString}`);
+
+    const channelsText = (await data.text()).substring(5); // Remove first 5 chars to have a correctly formatted JSON
+    channels = JSON.parse(channelsText);
+  } catch (error) {
+    console.error('Error fetching the channels', error);
+  }
```
matthiasrohmer, line 53: Consider refactoring this into a reusable `apiFetch` (or anything) method to deduplicate L19-25

external/build/deprecation-calendar.js
```diff
@@ -0,0 +1,121 @@
...
+ * @return {number}
+ */
+function deprecationsSort(a, b) {
+  if (!b.browsers.chrome.desktop) return -1;
+  if (!a.browsers.chrome.desktop) return 1;
+
+  return (
+    parseInt(b.browsers.chrome.desktop) - parseInt(a.browsers.chrome.desktop)
+  );
+}
```
matthiasrohmer, line 89: Refactor to take the full array instead, easier to extend.

external/build/deprecation-calendar.js
```diff
@@ -0,0 +1,121 @@
...
+      id: deprecation.id,
+      version: deprecation.browsers.chrome.desktop,
+      name: deprecation.name,
+      date: date,
+    };
+
+    return formattedDeprecation;
+  });
+
+  const targetFile = path.join(__dirname, '../data/deprecation-calendar.json');
```
matthiasrohmer, line 113: Use `chromeDeprecations.json` as a name, I think it's more descriptive, while at the same time not limiting the data use to the calendar page.

site/_data/deprecation-calendar/deprecation-articles.yml
```diff
@@ -0,0 +1,2 @@
+- deprecation_id: 6239658726391808
```
matthiasrohmer, line 1: Since you got `deprecations` (mind the s 😉 ) in the folder title, no need to duplicate it in the file name.

site/_data/deprecationCalendar.js
```diff
@@ -0,0 +1,73 @@
+const path = require('path');
```
matthiasrohmer, line 1: Move into `site/_data/deprecation-calendar/index.js` - not too sure if that works, but it would be neat 😄

site/_data/deprecationCalendar.js
```diff
@@ -0,0 +1,73 @@
...
+const fs = require('fs').promises;
+const yaml = require('js-yaml');
+
+/**
+ * Pulls a list of deprecataion items
```
matthiasrohmer, line 6: Nit: `deprecation`

site/_data/deprecationCalendar.js
```diff
@@ -0,0 +1,73 @@
...
+}
+
+/**
+ * Pulls articles related to deprecations from the deprecation-articles.yml file
+ *
+ * @return {promise} A promise that resolves in an array containing all the articles
+ */
+async function getArticles() {
+  let articles;
+  const articlesFile = path.join(
```
matthiasrohmer, line 33: Misleading variable name, that's only the path, as this will always be the same consider moving it to the top of the file like `const RELATED_ARTICLES_PATH = ...`.

site/_data/deprecation-calendar/deprecation-articles.yml
```diff
@@ -0,0 +1,2 @@
...
+  article: "/articles/declarative-shadow-dom/"
```
matthiasrohmer, line 2: As it's a 1:1 relationship, consider using the id as a key, this will allow you to simply lookup articles by id further down, instead of needing to do `articles.filter` everytime you want to look one up.

site/_data/deprecationCalendar.js
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * Adds the articles to the correct deprecation
+ *
+ * @param {object[]} deprecations
+ * @param {object[]} articles
+ * @return {object[]} An array of deprecations with their related articles
+ */
+function addArticles(deprecations, articles) {
+  deprecations.forEach(deprecation => {
+    const linkedArticle = articles.filter(
```
matthiasrohmer, line 58: See comment in the articles.yaml - essentially you wrote a nested for-loop here (filter inside a forEach callback), which is always good to avoid.

site/_data/deprecationCalendar.js
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Adds the articles to the correct deprecation
+ *
+ * @param {object[]} deprecations
+ * @param {object[]} articles
+ * @return {object[]} An array of deprecations with their related articles
+ */
+function addArticles(deprecations, articles) {
+  deprecations.forEach(deprecation => {
```
matthiasrohmer, line 57: Always prefer proper `for ... of` loops instead of `forEach` as this always the looped iterable interface to easily change without relying on methods provided by the iterable (array in this case).

site/_data/deprecationCalendar.js
simgobb: Uhm - tried this and technically works, the issue is that then the data gets formatted in a slightly different way. Instead of having just {myData} I would get something like { articles: {articlesData}, index: {myData}.

Which happy to go that way, I suppose it makes it cleaner on file system, but a tiny bit less clean when the data is needed.

@matthiasrohmer

matthiasrohmer (approved):

PR #5830, opened 2023-03-28
Tighten up the intro to the migration guide

jpmedley:
Almost everything in intro other than the list was stated at least twice. This PR fixes that.


jpmedley:
@IanStanion-google This is super small. Can you please look at this real quick?

PR #5883, opened 2023-04-02
Fix anchor typo

daidr:

- Fixed an anchor typo in https://developer.chrome.com/docs/extensions/reference/runtime

image


AaronForinton (approved):
LGTM, thanks for spotting and correcting this.

PR #5923, opened 2023-04-06
Update featured post

beaufortfrancois:


rachelandrew:
@beaufortfrancois we don't want to do this right now as otherwise the same post will display as the latest blog, just under the featured box.

Once something else is in the latest blog spot we can merge this.

beaufortfrancois:
Maybe after Chrome 111 beta blog post is out later today?

rachelandrew:
yeah that should end up in the latest blog spot so I can do it then.

rachelandrew (approved):

Issue #4847, opened 2023-01-11
Fix and investigate express.js runtime errors


matthiasrohmer:
We have a bunch of long-standing runtime errors we should address by either fixing them or at least properly catch them:

```
[RangeNotSatisfiableError: Range Not Satisfiable](console.cloud.google.com/errors/detail/CIC4ypbP8JDXSA;time=P1D?project=******)
270/workspace/node_modules/express/node_modules/send/index.js(SendStream.erro
```
Over 1980 times in the last 30 days, the one thrown most often

```
[PreconditionFailedError: Precondition Failed](console.cloud.google.com/errors/detail/CKbZqLae6vyg6QE;time=P30D?project=******)
270/workspace/node_modules/express/node_modules/send/index.js(SendStream.erro
```
60 times in the last 30 days

```
TypeError [ERR_INVALID_URL]: Invalid URL: ////%09/example.com

at .NodeError ( [internal/errors.js:322](console.cloud.google.com/debug?referrer=fromlog&file=internal%2Ferrors.js&line=322&project=******) )
at .onParseError ( [internal/url.js:270](console.cloud.google.com/debug?referrer=fromlog&file=internal%2Furl.js&line=270&project=******) )
at .parse
at .URL ( [internal/url.js:346](console.cloud.google.com/debug?referrer=fromlog&file=internal%2Furl.js&line=346&project=******) )
at .domainRedirectHandler ( [/workspace/server/unknown-domain.js:61](console.cloud.google.com/debug?referrer=fromlog&file=%2Fworkspace%2Fserver%2Funknown-domain.js&line=61&project=******) )
```
60 times in the last 30 days (so potentially related to the one above) but also ~30 times more with other URL variants.

matthiasrohmer:
There is also

```
AssertionError: Path "extensions/download_images//images" is invalid

at .AssertIsValid ( [/base/data/home/apps/s~chrome-apps-doc/3-72-0.430732081695229562/path_util.py:29](https://pantheon.corp.google.com/debug?referrer=fromlog&file=%2Fbase%2Fdata%2Fhome%2Fapps%2Fs~chrome-apps-doc%2F3-72-0.430732081695229562%2Fpath_util.py&line=29&appModule=default&appVersion=3-72-0&project=chrome-apps-doc) )
[...]

PR #5927, opened 2023-04-06
Add express.js handler for internal server errors

matthiasrohmer:
Fixes https://github.com/GoogleChrome/developer.chrome.com/issues/4847.

This adds a new handler for internal server errors, for the following reasons:

- To serve a proper error page if things go wrong instead of showing the default App Engine error page, which suggests an infrastructure problem by reading "Something went wrong. Please try again in 30s". This will get even more important once we have dynamic rendering
- To properly catch, log and categorize known errors. For example we have over 15k occurrences of `RangeNotSatisfiable`, which we can't really debug as we don't have information where range requests are sent to. Another known error is when people try to find attack vectors and append weird stuff to our domain (like https://developers.chrome.com//https%3A//example.com/..%2F)
- To bring down the amount of logged errors to properly alert if things go *really* wrong.


devnook (approved):
Please update the file location

Issue #71, opened 2020-12-17
LHCI complains about small tap target on /docs/ page


robdodson:
@robdodson commented on [Mon Dec 07 2020](https://github.com/GoogleChrome/dcc-private/issues/476)

samthor:
LHCI doesn't complain any more.

PR #5975, opened 2023-04-11
feat: create rss feeds per author

### LINKED ISSUE #5859: Create RSS feeds per author
The list of articles available on the author pages like https://developer.chrome.com/authors/paulkinlan/ also needs to be surfaced in individual RSS feeds per author in order to use that content on web.dev and incorporate it into the author page there.

mamieorine:
Fixes #5859 

According to [#71](https://github.com/GoogleChrome/webdev-infra/issues/71), the d.d.c author RSS feeds should be surfaced on web.dev as well as the web.dev author feeds should be shown on the author page on d.d.c. We currently have an RSS feed per author for web.dev (e.g. https://web.dev/authors/paulkinlan/feed.xml) but don’t have one for d.d.c. yet. Therefore, we need to create RSS feeds per author in d.d.c. Then we can add those feed.xml paths into the `external` key.

---

- Create RSS feeds per author
   - Please see: https://deploy-preview-5975--developer-chrome-com.netlify.app/authors/paulkinlan/feed.xml
- Display a RSS icon on the author profile page
   - Please see: https://deploy-preview-5975--developer-chrome-com.netlify.app/authors/paulkinlan/


site/_collections/feeds.js
```diff
@@ -106,6 +109,21 @@ module.exports = collection => {
         tagsFeeds[tag].items.push(post);
       }
     }
+
+    const authors = [post.data.authors ?? []].flat();
+    for (const author of authors) {
+      // If author feed does not exist in feeds yet, create FeedsCollectionItem
+      if (!authorsFeeds[author]) {
```
matthiasrohmer, line 116: This condition is only necessary because `authors` might contain duplicate authors, no? You could instead think about converting `authors` to a `Set` which would remove duplicates and save unnecessary loops.

matthiasrohmer (commented):
Nice, easier than I thought. Just one comment.

site/_collections/feeds.js
mamieorine: Yes, this condition is for checking the author key whether exists in the `authorsFeeds` to avoid adding duplicate authors.

However, I don't think we can remove this condition. According to the current logic, we loop over posts, then list all authors for each post, and loop over the authors to add the current post into the author key in the  `authorsFeeds`. I think we need to keep `!authorsFeeds[author]` condition to check whether we have this author in the `authorsFeeds`. If so just add a current post into the `item`. If not, create a new author key in `authorsFeeds` which is the same code as in that condition. 

The result of the authorsFeeds will be like this: 
```
{
  permalink: '/authors/paulkinlan/feed.xml',
  title: 'Paul Kinlan',
  url: '/authors/paulkinlan',
  items: [
    {
      template: [Template],
      inputPath: './site/en/blog/insider-dec-22/index.md',
      fileSlug: 'insider-dec-22',
      filePathStem: '/en/blog/insider-dec-22/index',
      data: [Object],
      date: 2022-12-15T00:00:00.000Z,
      outputPath: 'dist/en/blog/insider-dec-22/index.html',
      url: '/en/blog/insider-dec-22/',
      checkTemplateContent: true,
      templateContent: [Getter/Setter]
    },
    {
      template: [Template],
      inputPath: './site/en/blog/insider-oct-2022/index.md',
      fileSlug: 'insider-oct-2022',
      filePathStem: '/en/blog/insider-oct-2022/index',
      data: [Object],
      date: 2022-10-05T00:00:00.000Z,
      outputPath: 'dist/en/blog/insider-oct-2022/index.html',
      url: '/en/blog/insider-oct-2022/',
      checkTemplateContent: true,
      templateContent: [Getter/Setter]
    },
    ...
  ]
}
```

Please feel free to let me know if I am wrong about this 😊

matthiasrohmer: Sorry Maya, I should have expanded the code a little more for full context. I had the wrong assumptions, what you originally had was perfectly fine 🙂

matthiasrohmer (approved):

PR #5994, opened 2023-04-13
[WordPress Playground] Update Pendant theme link to WordPress.org instead of WordPress.com

adamziel:
The Pendant theme exists in the open-source theme directory on WordPress.org and I've accidentally linked to the directory on WordPress.com. Let's update it to link to the .org site.

cc @tomayac


tomayac (approved):
LGTM

PR #6004, opened 2023-04-14
Misc updated to Prerender docs

tunetheweb:
Misc changes to the Prerender post to reflect some changes on this API:

- Add details on prefetching and that it can be a good first step
- Add details on why to prefer speculation rules for prefetching over `<link rel="prefetch">`
- Add a `Speculation rules restrictions and future enhancements` header for this section and move it down a bit
- Add an "update" about how same-site cross-origin is now supported with the `Supports-Loading-Mode: credentialed-prerender` HTTP header
- Add an update that cross-origin prerenders are not supported but may be in future with the `Supports-Loading-Mode: uncredentialed-prerender`
- Add an update on the [experiments](https://github.com/WICG/nav-speculation/blob/main/chrome-2023q1-experiment-overview.md) currently in origin trial
- Correct speculation rules capitalisation ("Speculation Rules API" when talking about the API and "speculation rules" otherwise)

FYI: @nhiroki @KenjiBaheux @domenic @jeremyroman


site/en/blog/prerender-pages/index.md
```diff
@@ -179,7 +194,25 @@ Multiple speculation rules can also be added to the same page, and they append t
 </script>
 ```
 
-#### Detecting of Speculation Rules API support
+### Speculation rules restrictions and future enhancements
+
+By default prerender is restricted to same-origin pages, opened within the same tab, but we are working to reduce those restrictions.
```
domenic, line 199: Maybe this should be "same-site" now, since per the next aside, that's the current boundary?

site/en/blog/prerender-pages/index.md
```diff
@@ -214,13 +247,13 @@ if (HTMLScriptElement.supports &&
...
 
-#### Speculation Rules and Content Security Policy
+### Speculation rules and Content Security Policy
 
-As Speculation Rules use a `<script>` element, even though they only contain JSON, they need to be included in the `script-src` [Content-Security-Policy](https://web.dev/csp/) if the site uses this—either using a hash or nonce. In future, a new `inline-speculation-rules` source will be supported in `script-src` allowing all `<script type="speculationrules">` elements on a page to be supported without the need of hashes or nonces.
+As speculation rules use a `<script>` element, even though they only contain JSON, they need to be included in the `script-src` [Content-Security-Policy](https://web.dev/csp/) if the site uses this—either using a hash or nonce. In future, a new `inline-speculation-rules` source will be supported in `script-src` allowing all `<script type="speculationrules">` elements on a page to be supported without the need of hashes or nonces.
```
domenic, line 256: This is implemented from Chrome 110, so it's no longer really a future extension: https://chromestatus.com/feature/5182859125456896 .

Also it might be worth noting that you can use `strict-dynamic` and insert speculation rules from hashed/nonced scripts. (We had a bug but it's fix now!!!)

domenic (commented):
Generally looking great!

site/en/blog/prerender-pages/index.md
domenic: I guess that's true. But we're not planning to change the "default" in that sense, so then the last half of the sentence feels a bit off.

tunetheweb: Ah gotcha.

WDYT about this?:

> At present speculation rules are restricted to pages opened within the same tab, but we are working to reduce that restrictions. By default prerender is restricted to same-origin pages.

tunetheweb: Updated to that suggestion in https://github.com/GoogleChrome/developer.chrome.com/pull/6004/commits/6af7f216ec227ff6dae496c6b46e6396f00ffc7f

site/en/blog/prerender-pages/index.md
tunetheweb: Not quite fully implemented IMHO but made that clearer in https://github.com/GoogleChrome/developer.chrome.com/pull/6004/commits/6af7f216ec227ff6dae496c6b46e6396f00ffc7f

site/en/blog/prerender-pages/index.md
domenic: LGTM!

site/en/blog/prerender-pages/index.md
domenic: Hmm, it should work on speculation rules included in the initial HTML...

malchata (approved):

PR #6038, opened 2023-04-18
Defined "offscreen document"

IanStanion-google:
Added a definition for the term "offscreen document" in the reference page.

### DIFF
--- a/site/en/docs/extensions/reference/offscreen/index.md
+++ b/site/en/docs/extensions/reference/offscreen/index.md
@@ -4,6 +4,8 @@ api: offscreen
 
 ## Manifest
 
+Offscreen documents are invisible webpages with their own DOM that can be created and managed using the `offscreen` API. Requesting this permission allows the extension to use DOM APIs without obtrusively opening new windows or tabs that interrupt the user experience. 
+
 You must declare the `"offscreen"` permission in the [extension manifest][doc-manifest] to use the Offscreen API. For example:
 
 ```json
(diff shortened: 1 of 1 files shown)


site/en/docs/extensions/reference/offscreen/index.md
```diff
@@ -4,6 +4,8 @@ api: offscreen
 
 ## Manifest
 
+Offscreen documents are invisible webpages with their own DOM that can be created and managed using the `offscreen` API. Requesting this permission allows the extension to use DOM APIs without obtrusively opening new windows or tabs that interrupt the user experience.
```
oliverdunk, line 7: Thoughts on moving this above the manifest header?

oliverdunk (changes requested):
Left one comment, otherwise LGTM 👍

site/en/docs/extensions/reference/offscreen/index.md
jpmedley: I agree. This is a basic offscreen definition. the place for it is here: https://source.chromium.org/chromium/chromium/src/+/main:extensions/common/api/offscreen.idl;l=5

jpmedley: @IanStanion-google, as requested, here's the Gerrit PR: https://chromium-review.git.corp.google.com/c/chromium/src/+/4440738

It doesn't include your second sentence. I think it's more appropriate here.

jpmedley (approved):

oliverdunk:
@IanStanion-google I think we can merge whenever :) Two things beforehand:

- Could we move this above the header still? e.g in this PR move your line above the line above it?
- We probably don't want to duplicate Joe's text, but I think the first line you had is still useful to have here too if you can reword it somehow.

oliverdunk (approved):

PR #6041, opened 2023-04-18
Only send search requests for queries longer than n characters

ShayPunter:
Fixes https://github.com/GoogleChrome/webdev-infra/issues/76

- Send requests to Algolia if search query is 3 or more characters
-- number calculated based on hits from popular search queries and length of keyword
- Add a message to search results to inform the user that they need to enter 3 characters for search suggestions
- Remove unused import striptags from search-box.js

n has been worked out in the following manner:

Sort the keyword spreadsheet by count and average the length of the top 10 keywords (output: 5 (rounded))
Sort the keyword spreadsheet by nbhits, exclude queries with just one char, average the length of the top 10 keywords (output: 2 (rounded))
Average both the count and nbhits keywords, which produced 4 (rounded), take into account that there are also a lot of acronyms that are 3 words, the final output will be 3 characters

## How to test
To test this change, spin up a local development environment, checkout `76_only_send_seach_queries_that_are_longer_than_n_characters`, build and run the application and then go to the search bar and begin searching, if two characters have been searched, it will display a message saying to type at least 3 characters for search results to return


site/_js/web-components/search-box.js
```diff
@@ -29,6 +29,15 @@ import {activateSearch, deactivateSearch} from '../actions/search';
...
 const index = client.initIndex('prod_developer_chrome');
 
+const blockedQueries = [
+  /add? bloc?k/,
+  /d(a|o)w?nl?o?/,
+  /^p(a|e)r?r(a|e)?/,
+  /automate be/,
+  /vpn/,
+  /xxx/,
+];
```
devnook, line 39: I'd make a blacklist in a separate yaml file, easily editable without touching the code.

devnook (approved):

site/_js/web-components/search-box.js
matthiasrohmer: Held back from doing so as it's client-side code and our Roll Up config doesn't support transforming YAML files, but also because after thinking about it I think having it in one place is more intuitive as we don't have a place for client-side configs yet.

matthiasrohmer (approved):

PR #6113, opened 2023-04-25
Meet the team updates barry

tunetheweb:

- Updates some events I'm involved in
- Fixes the "Load more" button on the [meet the team page](https://developer.chrome.com/meet-the-team/) which doesn't work as it assumes innerHTML is set at the time the callbacks is made, but that's not guaranteed. Switched it to an attribute instead.

### DIFF
--- a/site/_includes/partials/event-card.njk
+++ b/site/_includes/partials/event-card.njk
@@ -112,7 +112,7 @@
 
               <div class="gap-bottom-300">
                 <p>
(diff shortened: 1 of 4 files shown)


matthiasrohmer (approved):
Thanks for the fix from the eng side - our lit components and their weirdness have a special place in my heart. Forwarding to @AaronForinton for the content side of things.

AaronForinton (approved):

PR #6116, opened 2023-04-26
Update Web Manifest id blog post

aleksandr-kiliushin:

- Update Web Manifest ID browser support section in [this](https://developer.chrome.com/blog/pwa-manifest-id) blogpost.
- Mention Firefox.
- Add Browser compatibility widget.
- Update phrases like `id` _will_ be available in Chrome 96.
- Remove Chrome Canary usage suggestions.
- Update "How do I test this today?" section. (I'm not quite sure, maybe it is no longer actual and should be deleted?)

### DIFF
--- a/site/en/blog/pwa-manifest-id/index.md
+++ b/site/en/blog/pwa-manifest-id/index.md
@@ -55,7 +55,7 @@ is to check the value calculated by Chrome.
 1. Add an `id` property to the web app manifest using the `id` value shown in
    the note.
 
-{% Img src="image/0g2WvpbGRGdVs0aAPc6ObG7gkud2/7vo2XUpQxz0RWfWBuLDa.png", class="screenshot", alt="Tooltip showing 'id' value.", width="800", height="175" %}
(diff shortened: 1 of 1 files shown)


aleksandr-kiliushin:
@rachelandrew @heyawhite please help with a screenshot updating. I can't figure out where it should be uploaded and how to get the `src` for `{% Img src="image/<path>.png" }`.

Screenshot should be updated at [this](https://github.com/aleksandr-kiliushin/developer.chrome.com/blob/9113ea8c1ed2480f6db1920cee7286c2d2938c6d/site/en/blog/pwa-manifest-id/index.md?plain=1#L58) line.

The new screenshot attached below.  
Dimensions for it: `width="750", height="120"`.

image

Thanks!

rachelandrew:
Hi - thanks for this, but this is a [blog post](https://developer.chrome.com/docs/handbook/where-to-publish/#blog-posts), so it doesn't need updating, it's essentially historical information which was correct at the time of publication.

PR #6125, opened 2023-04-26
Add LCP phase breakdown to the Lighthouse docs

adamraine:
Part of https://github.com/GoogleChrome/lighthouse/pull/14891


adamraine:
> Remove references to Chrome 77 (it's so long ago who cares!)

This works for me.

> Consider adding a "From Lighthosue 10.2, Lighthouse will display the LCP element along with the phase breakdown...". That will get old too so maybe not needed (especially when references to Chrome 77 are gone)? What do you usually do here?

I'm generally opposed to calling out when a specific feature was added unless it's *super* important or a breaking change.

malchata (approved):

PR #6187, opened 2023-05-03
[Deprecate Web SQL article] Update removal timeline

tomayac:

- Update removal timeline.

### DIFF
--- a/site/en/blog/deprecating-web-sql/index.md
+++ b/site/en/blog/deprecating-web-sql/index.md
@@ -10,7 +10,7 @@ description: >
 authors:
   - thomassteiner
 date: 2022-08-31
-updated: 2023-01-12
+updated: 2023-05-03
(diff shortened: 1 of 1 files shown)


rachelandrew:
@tomayac updating this like this makes no sense, no one is going to see it unless they specifically search for it. That's not helpful as a way of alerting people to an upcoming removal. 

It's so changed, I would suggest just making the it a new post, so it shows up in the listing and in RSS, and adding the outdated banner to the old one.

tomayac:
Fair enough, but we still need to keep this article fresh, since we said everywhere that we would (bindingly, in the blink-dev deprecation thread for example). Let's update this article, and I can also hammer out a quick new announcement post.

rachelandrew:
Please don't use the blog for this kind of thing in future. As I keep pointing out, it's very odd to do this kind of rewriting of history in a news stream.

rachelandrew (approved):

PR #6197, opened 2023-05-04
Redirecting popover docs to MDN

rachelandrew:
Also tidied up the internal renaming redirects.


matthiasrohmer (approved):
Tested via https://popover-redirects-dot-dcc-staging.uc.r.appspot.com/docs/web-platform/pop-up-api/popover-attribute/ et al, works!

PR #6200, opened 2023-05-04
Surface DNR limits in top-level documentation

oliverdunk:
Screenshot 2023-05-04 at 11 28 39

We already document all declarativeNetRequest limits at https://developer.chrome.com/docs/extensions/reference/declarativeNetRequest/, but mostly called them out alongside other documentation and with links to the constants in the auto-generated documentation.

This PR aims to make them easier to find by moving them in to their own section and specifically mentioning the limits. While this may add a small amount of additional work (we need to make sure we update the section if the limits change), I think this is an extremely beneficial change and that overhead will be worthwhile.

I also updated some of our code formatting to be consistent with other docs.

### DIFF
--- a/site/en/docs/extensions/reference/declarativeNetRequest/index.md
+++ b/site/en/docs/extensions/reference/declarativeNetRequest/index.md
@@ -7,7 +7,7 @@ has_warning: One or more of these permissions <a href="/docs/extensions/mv3/perm
 
 ---
 
-## Manifest
+## Manifest {: #manifest }
 
 Extensions must declare either the `declarativeNetRequest` or the
 `declarativeNetRequestWithHostAccess` (available since **Chrome 96**) permission in the extension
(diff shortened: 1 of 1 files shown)


AmySteam (changes requested):
@oliverdunk Great improvements 💯 

I only have one request: could you please incorporate [header-ids](https://developer.chrome.com/docs/handbook/extensions/#header-ids) throughout? It's especially important in this case because there are two headers with the same name (#rules).

oliverdunk:
> @oliverdunk Great improvements 💯
> 
> I only have one request: could you please incorporate [header-ids](https://developer.chrome.com/docs/handbook/extensions/#header-ids) throughout? It's especially important in this case because there are two headers with the same name (#rules).

Done!

AmySteam (approved):
LGTM 👍

PR #6201, opened 2023-05-04
feature: added description to enum types

### LINKED ISSUE #5619: Render descriptions for enum values in extension documentation

When viewing an enum in the extensions docs, we currently only show the enum values ([example](https://developer.chrome.com/docs/extensions/reference/offscreen/#type-Reason)):

image

We already have detailed descriptions available in the types this output is generated from. Example:

image

We should make sure to render these.

We may need to change the types workflow to make this easier on the 11ty side. I'm happy to help with that.

**Additional context**

https://github.com/GoogleChrome/developer.chrome.com/issues/4788

simgobb:
Fixes #5619 

- Added descriptions to enum types

### DIFF
--- a/site/_includes/macros/render-type.njk
+++ b/site/_includes/macros/render-type.njk
@@ -241,7 +241,7 @@
       <p>
         {% for t in types %}
           {% if loop.last and not loop.first %}or{% endif %}
-          <span style="white-space: nowrap">{{ renderSingleIconType(t, node) | trim }}{% if not loop.last %}<br>{% endif %}</span>
+          <span style="white-space: nowrap;">{{ renderSingleIconType(t, node) | trim }}</span>
         {% endfor %}
       </p>
     </div>
--- a/site/_includes/macros/render-type.njk
+++ b/site/_includes/macros/render-type.njk
@@ -40,7 +40,7 @@
 
   {% elif type.type === 'union' %}
     {# Type that contains many string (probably) options, basically an enum. #}
-    {{ renderIconTypesArray(type.types, 'Type') }}
+    {{ renderIconTypesArray(type.types, 'Type', node) }}
 
   {% elif type.type and not node._event %}
     {# Reference to another type (and not an Event). #}
@@ -206,8 +206,16 @@
     {% set iconType = type.type %}
   {% endif %}
 
+   
+  {% set description = "" %}
+  {% for enum in node._enums %}
+    {% if enum.value == type.value %}
+      {% set description = enum.description %}
+    {% endif %}
+  {% endfor %}
+
   {# Note that we avoid whitespace below, and instead set margins in CSS. #}
-  <p class="code-sections__icon code-sections__icon--{{ iconType }}">{{ renderSingleType(type, node) | trim | safe }}
+  <p class="code-sections__icon code-sections__icon--{{ iconType }}">{{ renderSingleType(type, description) | trim | safe }}
     {%- if node.flags.isOptional -%}
       &nbsp;<span class="code-sections__optional">optional</span>
     {%- endif -%}
@@ -223,8 +231,7 @@
 {#
   Renders an optional list of types, e.g., for an enum.
 #}
-{% macro renderIconTypesArray(types, label) %}
-
+{% macro renderIconTypesArray(types, label, node) %}
   <div class="code-sections">
     {% if label %}
       <h4 class="type--label code-sections__mode">{{ label }}</h4>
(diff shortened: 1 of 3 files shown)


site/_includes/macros/render-type.njk
```diff
@@ -206,8 +206,16 @@
...
+  {% set description = "" %}
+  {% for enum in node._enums %}
+    {% if enum.value == type.value %}
+      {% set description = enum.description %}
+    {% endif %}
+  {% endfor %}
+
   {# Note that we avoid whitespace below, and instead set margins in CSS. #}
-  <p class="code-sections__icon code-sections__icon--{{ iconType }}">{{ renderSingleType(type, node) | trim | safe }}
+  <p class="code-sections__icon code-sections__icon--{{ iconType }}">{{ renderSingleType(type, description) | trim | safe }}
```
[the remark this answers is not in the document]
matthiasrohmer: I think that works for now 🙂

matthiasrohmer, line 218: Looking at https://pr-6201-static-dot-dcc-staging.uc.r.appspot.com/docs/extensions/reference/offscreen/#type-Reason it seems as if the descriptions are markdown and could benefit from running through `| mdInline` 🙂

site/_includes/macros/render-type.njk
```diff
@@ -303,7 +310,7 @@
 
   {% elif type.type === 'literal' %}
     {# Literal value, e.g., 123 or '"foo"' #}
-    <span class="code-sections__value">{{ type.value | dump }}</span>
+    <span class="code-sections__value">{{ type.value }}</span> <br> <span class="type--xsmall" style='white-space: initial'>{{description}}<span>
```
matthiasrohmer, line 313: Classic `'` vs `"`.

matthiasrohmer:
Hey @oliverdunk, can you have a look at this as you much better know what you were looking for than I do 🙂  See https://pr-6201-static-dot-dcc-staging.uc.r.appspot.com/docs/extensions/reference/offscreen/#type-Reason for an example. I already noted above that the descriptions' markdown would need to be rendered, but otherwise this seems what you've been looking for. Thanks!

oliverdunk:
This is amazing, thank you so much 🙌 

One thing we noticed is that there's maybe some other code: https://github.com/GoogleChrome/developer.chrome.com/blob/34c282f01f31f448c93893aec1755398d645767d/site/_includes/macros/render-type.njk#L275

Do you know what styles that generates? Are they nicer styles we could reuse?

Alternatively, if we could do a quick fix to remove these commas that would at least increase the information density and then I think we're good to go 😄 

Screenshot 2023-05-04 at 18 37 30

simgobb:
@oliverdunk @matthiasrohmer 

I've tried the `code-sections__enum` style, but it didn't change much (the definition was black instead of blue) and it might require a slight refactor on how the macros are nested.

I removed the comma and added a new line to the original PR fix. Hope is more along the lines of what you had in mind.

oliverdunk:
@simgobb, thanks for checking! In that case, one more nit, could we reduce the gap a bit? e.g:

Screenshot 2023-05-05 at 11 01 48

simgobb:
@oliverdunk 

No problem. Added a 1rem margin instead of the original <br>.

Let me know if it looks better, otherwise, I can play with that space a bit until it's right!

oliverdunk:
@simgobb I think enums are looking great! A couple more things:

1:

Screenshot 2023-05-05 at 12 10 52

It looks like the class we changed is also used in some other places, e.g this type at https://pr-6201-static-dot-dcc-staging.uc.r.appspot.com/docs/extensions/reference/certificateProvider#type-CertificatesUpdateRequest. Is there a way we can only change enums and avoid adding extra margin to anything else?

2:

I think it probably makes sense to drop the "or" before the last item?

3:
Screenshot 2023-05-05 at 12 17 59

It looks like for enums with more than one underscore **and** a description we accidentally include some of the enum name in the description: https://pr-6201-static-dot-dcc-staging.uc.r.appspot.com/docs/extensions/reference/certificateProvider#type-Algorithm

I think that's probably a bug here where the assumption `raw.length + 1` will be the right amount to remove doesn't hold: https://github.com/GoogleChrome/developer.chrome.com/blob/main/external/build/lib/dts-parse.js#LL799C32-L799C32

I'm not sure how familiar with that code you are, let me know if you're comfortable taking a look or if not I can probably commit something.

jpmedley:
@oliverdunk We're not anywhere close to finishing the review of the text that we have no. For the descriptions we have extensions engineers have been writing for at least a decade knowing that the text would NOT appear on our public site. I don't want to just turn this content on without a review.

cc: @sebastianbenz

sebastianbenz:
Agree with @jpmedley that we should review all descriptions. However, I would not block on this - the benefit of having descriptions outweighs  occasional spelling mistakes / errors. Review should be a quick follow up though. Created https://github.com/GoogleChromeLabs/Extension-Docs/issues/59 for tracking.

jpmedley:
It's not spelling mistakes I'm worried about.
Joe Medley | Technical Writer, Chrome DevRel | ***@***.*** |
 816-678-7195
*If an API's not documented it doesn't exist.*

oliverdunk (approved):
Approving from an extension POV, thanks so much for the fast work here :)

matthiasrohmer (approved):

PR #6225, opened 2023-05-08
Improve Topics API subhead & description

adamsilverstein:
Matches other API pages by providing an actual description of the API instead of a call to action. eg. "The Topics API..." Learn about this..."


heyawhite (approved):

nancymic2 (commented):
This looks good. Perhaps remove the acronym in parens. Otherwise LGTM

PR #6269, opened 2023-05-11
Remove initialAspectRatio

beaufortfrancois:
Since we don't currently support locking the aspect ratio, having `initialAspectRatio` as an option doesn't really make sense, so we should remove it.

Spec PR: https://github.com/WICG/document-picture-in-picture/pull/36
Chromium CL: https://chromium-review.googlesource.com/c/chromium/src/+/4518212

FYI @steimelchrome @liberato-at-chromium

PR #6278, opened 2023-05-11
Revert "Remove initialAspectRatio"

AaronForinton:
Reverts GoogleChrome/developer.chrome.com#6269


rachelandrew:
@AaronForinton we don't need to do this now so I'll close this. Thanks!

PR #6280, opened 2023-05-11
Fix the basic syntax to include // and port between HOST and URI

dileepnitc:
The intent uri syntax misses "//", while the examples does include them. Also introducing the PORT between HOST and URI as an optional string.

-
-
-


hkl615:
Hi

PR #6295, opened 2023-05-13
Update example code: setupOffscreenDocument

hanguokai:
Replace hasOffscreenDocument() with setupOffscreenDocument() to resolve concurrency issues

### DIFF
--- a/site/en/docs/extensions/reference/offscreen/index.md
+++ b/site/en/docs/extensions/reference/offscreen/index.md
@@ -40,7 +40,7 @@ Use `chrome.offscreen.createDocument()` and `chrome.offscreen.closeDocument()` f
 
 ```js
 chrome.offscreen.createDocument({
-  url: chrome.runtime.getURL('off_screen.html'),
+  url: 'off_screen.html',
   reasons: ['CLIPBOARD'],
   justification: 'reason for needing the document',
 });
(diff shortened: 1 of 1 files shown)


hanguokai:
@sebastianbenz [suggested me](https://groups.google.com/a/chromium.org/g/chromium-extensions/c/D5Jg2ukyvUc/m/JFhAYDEpAAAJ) to create this PR.

site/en/docs/extensions/reference/offscreen/index.md
```diff
@@ -40,44 +40,54 @@ Use `chrome.offscreen.createDocument()` and `chrome.offscreen.closeDocument()` f
...
+async function setupOffscreenDocument(path) {
   // Check all windows controlled by the service worker to see if one 
   // of them is the offscreen document with the given path
+  let exist = false;
   const offscreenUrl = chrome.runtime.getURL(path);
   const matchedClients = await clients.matchAll();
   for (const client of matchedClients) {
     if (client.url === offscreenUrl) {
-      return true;
+      exist = true;
```
sebastianbenz, line 63: I'd prefer an early exit here. This also avoid the additional variable + if below. .

site/en/docs/extensions/reference/offscreen/index.md
```diff
@@ -40,44 +40,54 @@ Use `chrome.offscreen.createDocument()` and `chrome.offscreen.closeDocument()` f
...
+      break;
+    }
+  }
+  
+  // create offscreen document if it's not existing
+  if (!exist) {
+    if (creating) {
+      await creating;
```
sebastianbenz, line 71: Do we need to `await` here? We could also simply return `creating` and avoid the `else` block below.

sebastianbenz (commented):
Amazing! Thanks a lot. Left a suggestion to use early exit otherwise LGTM.

site/en/docs/extensions/reference/offscreen/index.md
hanguokai: Thanks, early exit here more simple

site/en/docs/extensions/reference/offscreen/index.md
hanguokai: I prefer the current way here, which looks more symmetrical

hanguokai:
Thanks for suggestion, I submitted a new version, that is a bit simpler

sebastianbenz (approved):
Thanks a lot - looks great!

site/en/docs/extensions/reference/offscreen/index.md
```diff
@@ -40,44 +40,50 @@ Use `chrome.offscreen.createDocument()` and `chrome.offscreen.closeDocument()` f
...
+  url: 'off_screen.html',
   reasons: ['CLIPBOARD'],
   justification: 'reason for needing the document',
 });
 
 chrome.offscreen.closeDocument()
 ```
 
-The following example shows how to check for existing offscreen documents. The `hasOffscreenDocument()` function uses [clients.matchAll()](https://developer.mozilla.org/docs/Web/API/Clients/matchAll) to find existing offscreen documents. Please note: the function assumes that your extension uses a single offscreen document. 
+The following example shows how to ensure that offscreen documents have already been created. The `setupOffscreenDocument()` function uses [clients.matchAll()](https://developer.mozilla.org/docs/Web/API/Clients/matchAll) to find existing offscreen documents. If no existing offscreen documents, create it. Please note: the function assumes that your extension uses a single offscreen document.
```
jpmedley, line 51:
```suggestion
The following example shows how to ensure that an offscreen document has already been created. The `setupOffscreenDocument()` function calls [`clients.matchAll()`](https://developer.mozilla.org/docs/Web/API/Clients/matchAll) to find the existing offscreen document. If no offscreen document exists, is created. Note that a service worker can only have one offscreen document. 
```

site/en/docs/extensions/reference/offscreen/index.md
```diff
@@ -40,44 +40,50 @@ Use `chrome.offscreen.createDocument()` and `chrome.offscreen.closeDocument()` f
...
+      justification: 'reason for needing the document',
+    });
+    await creating;
+    creating = null;
+  }
 }
 ```
 
-Before trying to create a new offscreen document, call `hasOffscreenDocument()` to make sure that there is no existing offscreen document, as demonstrated in the following example. 
+Before sending a message to an offscreen document, call `setupOffscreenDocument()` to make sure that there is existing offscreen document, as demonstrated in the following example.
```
jpmedley, line 81:
```suggestion
Before sending a message to an offscreen document, call `setupOffscreenDocument()` to make sure that there is an existing offscreen document, as demonstrated in the following example. 
```

jpmedley (changes requested):

site/en/docs/extensions/reference/offscreen/index.md
jpmedley: I fixed the first item.

I didn't remove the 'Please note...'. I just made it a bit pithier.

jpmedley: I appreciate your time and input, but this is problematic. I think you're trying to future proof this, which is a good instinct. But, baseline documentation shouldn't make promises about future product work. If we've done that elsewhere, I need to fix it.

What this line needs to do is say in a straightforward way that the function gets the existing offscreen document or creates it if it doesn't already exist and that it does so using...

In other words, stick to the facts.

hanguokai:
I submitted a new version followed by Joe's suggestion. Please review. Thanks!

jpmedley (approved):

PR #6306, opened 2023-05-15
[ext-doc] Convert XMLHttpRequest to fetch()

IanStanion-google:
Fixes (https://github.com/GoogleChromeLabs/Extension-Docs/issues/45)

- Replaces the use of XMLHttpRequest to fetch() in sample code blocks.

### DIFF
--- a/site/en/docs/extensions/mv3/content_scripts/index.md
+++ b/site/en/docs/extensions/mv3/content_scripts/index.md
@@ -711,14 +711,9 @@ All assets must be declared as [Web Accessible Resources][manifest-war] in the `
 
 ## Stay secure {: #security }
 
-{% Aside 'warning' %}
-In Manifest V3, `XMLHttpRequest` is not supported in Service Workers.
-Please consider using its modern replacement, `fetch()`.
-{% endAside %}
-
 While isolated worlds provide a layer of protection, using content scripts can create
 vulnerabilities in an extension and the web page. If the content script receives content from a
-separate website, such as making an [`XMLHttpRequest`][28] or fetch() request, be careful to filter content
+separate website, such as making a fetch() request, be careful to filter content
 [cross-site scripting][29] attacks before injecting it. Only communicate over HTTPS in order to
 avoid ["man-in-the-middle"][30] attacks.
 
--- a/site/en/docs/extensions/mv3/content_scripts/index.md
+++ b/site/en/docs/extensions/mv3/content_scripts/index.md
@@ -711,9 +711,14 @@ All assets must be declared as [Web Accessible Resources][manifest-war] in the `
 
 ## Stay secure {: #security }
 
+{% Aside 'warning' %}
+In Manifest V3, `XMLHttpRequest` is not supported in Service Workers.
+Please consider using its modern replacement, `fetch()`.
+{% endAside %}
+
 While isolated worlds provide a layer of protection, using content scripts can create
 vulnerabilities in an extension and the web page. If the content script receives content from a
-separate website, such as making an [`XMLHttpRequest`][28], be careful to filter content
+separate website, such as making an [`XMLHttpRequest`][28] or fetch() request, be careful to filter content
 [cross-site scripting][29] attacks before injecting it. Only communicate over HTTPS in order to
 avoid ["man-in-the-middle"][30] attacks.
 
--- a/site/en/docs/extensions/mv3/security/index.md
+++ b/site/en/docs/extensions/mv3/security/index.md
@@ -161,27 +161,6 @@ function constructDOM() {
 }
 ```
 
-### eval() {: #eval }
-
-Avoid using `eval()` whenever possible to prevent attacks, as `eval()` will execute any code passed
-into it, which may be malicious.
-
-```js
-const fetchValue = await fetch("https://api.example.com/data.json");
-const jsonData = await fetchValue.json();
-// WARNING! Might be evaluating an evil script!
-const result = eval(jsonData);
-```
-
-Instead, prefer safer, and faster, methods such as `JSON.parse()`
-
-```js
-const fetchValue = await fetch("https://api.example.com/data.json");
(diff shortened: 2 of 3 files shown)


site/en/docs/extensions/mv3/security/index.md
```diff
@@ -162,7 +162,7 @@ Avoid using `eval()` whenever possible to prevent attacks, as `eval()` will exec
 into it, which may be malicious.
 
 ```js
-var xhr = new XMLHttpRequest();
+var xhr = new fetch();
```
jpmedley, line 165: `XMLHttpRequest()` and `fetch()` have very different surfaces. You'll need to rewrite the the whole sample.

IanStanion-google:
@jpmedley Could you weight in on keeping reference to xhr in some of these pages? I left it in and added asides in some cases because the information is correct, but it depends on how much we want to remove reference to xhr or inform developers that they need to switch.

site/en/docs/extensions/reference/webRequest/index.md
```diff
@@ -180,7 +180,7 @@ permissions][9]. Moreover, only the following schemes are accessible: `http://`,
 `chrome-extension://`. In addition, even certain requests with URLs using one of the above schemes
 are hidden. These include `chrome-extension://other_extension_id` where `other_extension_id` is not
 the ID of the extension to handle the request, `https://www.google.com/chrome`, and other sensitive
-requests core to browser functionality. Also synchronous XMLHttpRequests from your extension are
+requests core to browser functionality. Also synchronous `fetch()` requests from your extension are
```
sebastianbenz, line 183: As Joe said, a simple find/replace is in most cases not possible, e.g. in this case, fetch requests are never synchronous and XHR should be kept here.

site/en/docs/extensions/mv3/content_scripts/index.md
```diff
@@ -711,9 +711,14 @@ All assets must be declared as [Web Accessible Resources][manifest-war] in the `
 
 ## Stay secure {: #security }
 
+{% Aside 'warning' %}
```
sebastianbenz, line 714: I don't think that we need to explicitly call this out. In particular in this file as it is not service worker related.

site/en/docs/extensions/mv3/content_scripts/index.md
```diff
@@ -711,9 +711,14 @@ All assets must be declared as [Web Accessible Resources][manifest-war] in the `
...
+In Manifest V3, `XMLHttpRequest` is not supported in Service Workers.
+Please consider using its modern replacement, `fetch()`.
+{% endAside %}
+
 While isolated worlds provide a layer of protection, using content scripts can create
 vulnerabilities in an extension and the web page. If the content script receives content from a
-separate website, such as making an [`XMLHttpRequest`][28], be careful to filter content
+separate website, such as making an [`XMLHttpRequest`][28] or fetch() request, be careful to filter content
```
sebastianbenz, line 721: We can drop the XHR reference here.

site/en/docs/extensions/mv3/security/index.md
sebastianbenz: +1

sebastianbenz (changes requested):
This needs more work. As part of this change /mv3/xhr/index.md should be completely rewritten for fetch as well.

jpmedley:
This needs to be replaced: https://developer.chrome.com/docs/extensions/mv3/xhr/

oliverdunk:
Took a quick look at this but I don't have any thoughts beyond what has been said already :)

site/en/docs/extensions/mv3/security/index.md
```diff
@@ -38,9 +38,14 @@ websites they depend on. Arbitrary code should be kept to a minimum.
 
 Limiting an extensions privileges limits what a potential attacker can exploit.
 
-### Cross-origin XMLHttpRequest {: #xhr }
+### Cross-origin fetch() {: #xhr }
 
-An extension can only use [XMLHttpRequest][6] to get resources from itself and from domains
+{% Aside 'warning' %}
+In Manifest V3, `XMLHttpRequest` is not supported in Service Workers.
+Please consider using its modern replacement, `fetch()`.
```
jpmedley, line 45:
```suggestion
`XMLHttpRequest()` is not supported in Service Workers.
Use its modern replacement, `fetch()`.
```

jpmedley: I made this suggestion because XHR not supported service workers at all. Fetch was created so that it didn't need to. "Please consider using" makes it sound as though there's a choice.

jpmedley (changes requested):
I'm suggesting a slightly different wording for warning.

jpmedley (changes requested):
I'm suggesting a slightly different wording for warning.

jpmedley (approved):

site/en/docs/extensions/mv3/security/index.md
```diff
@@ -162,30 +167,19 @@ Avoid using `eval()` whenever possible to prevent attacks, as `eval()` will exec
...
-    // WARNING! Might be evaluating an evil script!
-    var resp = eval("(" + xhr.responseText + ")");
-    ...
-  }
-}
-xhr.send();
+const fetchValue = await fetch("https://api.example.com/data.json");
+const jsonData = await fetchValue.json();
+// WARNING! Might be evaluating an evil script!
+const result = eval(jsonData);
```
sebastianbenz, line 173: Do we even need this section? eval is no longer supported in MV3.

cc @jpmedley

jpmedley: Even if it was still supported, why would evaluate JSON?

sebastianbenz: Please remove as it doesn't make sense in MV3.

sebastianbenz (approved):
Thanks for the quick fix

PR #6321, opened 2023-05-15
GTM and GA4 implementation

tunetheweb:

- Implements GTM and GA4 similar to web.dev
- Changes cookie banner implementation more inline with web.dev
- Allows storage values as well as href for announcement banners


site/_includes/layouts/base.njk
```diff
@@ -17,12 +17,12 @@
...
     <link rel="preload" as="font" crossorigin href="/fonts/google-sans-v2003/medium/latin.woff2">
 
-    
+
     {% block css %}{% endblock %}
 
     <script type="module" src="{{ helpers.hashForProd('/js/main.js') }}"></script>
-    {% if process.env.NODE_ENV === 'production' %}
-      <script async src="https://www.google-analytics.com/analytics.js"></script>
+    {% if process.env.NODE_ENV !== 'production' %}
```
tunetheweb, line 24: Reminder to myself - revert this change (which is only there for testing) before merging

tunetheweb:
```suggestion
    {% if process.env.NODE_ENV === 'production' %}
```

tunetheweb:
OK @matthiasrohmer  and @devnook I think this is ready for review.

Basically copies the implementation from web.dev.

devnook (approved):
Thanks @tunetheweb !

PR #6379, opened 2023-05-22
Add aside to Early Hints article warning about which resources to hint

tunetheweb:

- Adds aside about how Early Hints may want to preconnect/preload different resources than those in HTML

FYI: @kenjibaheux @rmarx


tunetheweb:
There's some whitespace clean up in this, so review with this setting on if you wanna see the actual changes:

image

https://github.com/GoogleChrome/developer.chrome.com/pull/6379/files?diff=unified&w=1

site/en/blog/early-hints/index.md
```diff
@@ -50,13 +51,21 @@ In some cases, the performance improvement to the [Largest Contentful Paint](htt
...
+{% Aside 'important' %}
+What you decide to `preconnect` or `preload` as Early Hints may differ from what you have traditionally used these directives for in your HTML.
+
+When using these in HTML you typically want to `preconnect` or `preload` resources that the [Preload Scanner](https://web.dev/preload-scanner/) will not discover in the HTML—for example, fonts, or background images that would otherwise be discovered late. For Early Hints you will not have the HTML and so want to `preconnect` to critical domains or `preload` critical resources that perhaps _would_ be discovered early in the HTML—for example, preloading `main.css` or `app.js`.
```
rmarx, line 63:
```suggestion
When using these in HTML you typically want to `preconnect` or `preload` resources that the [Preload Scanner](https://web.dev/preload-scanner/) will not discover in the HTML—for example, fonts, or background images that would otherwise be discovered late. For Early Hints you will not have the HTML and so you may want to instead `preconnect` to critical domains or `preload` critical resources that perhaps _would_ otherwise be discovered early in the HTML—for example, preloading `main.css` or `app.js`.
```

malchata (approved):

PR #6382, opened 2023-05-22
View Transition API => View Transitions API

hubgit:
The title of the article is "Smooth and simple transitions with the View Transitions API", so I think it makes sense to be consistent in the body text?

> View Transitions is a proposal for a new web API

https://github.com/WICG/view-transitions

> The `ViewTransition interface` of the [View Transitions API](https://developer.mozilla.org/en-US/docs/Web/API/View_Transitions_API)

https://developer.mozilla.org/en-US/docs/Web/API/ViewTransition

### DIFF
--- a/site/en/docs/web-platform/view-transitions/index.md
+++ b/site/en/docs/web-platform/view-transitions/index.md
@@ -4,7 +4,7 @@ title: Smooth and simple transitions with the View Transitions API
 authors:
   - jakearchibald
 description: >
-  The View Transition API allows page transitions within single-page apps, and will later include multi-page apps.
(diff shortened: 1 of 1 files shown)


rachelandrew (approved):

bramus:
While these type of transitions are called View Transitions indeed, the API is named “View Transition API”. Check the spec: https://drafts.csswg.org/css-view-transitions/

Admittedly, I’ve also been saying it wrong and so does everyone else _(even internally)_ …

hubgit:
Thanks - it sounds like the text is already correct, so I'll close this.

PR #6394, opened 2023-05-23
Updated xhr page to general Cross Origin Requests page

IanStanion-google:
Fixes [#SOME_ISSUE_NUMBER](https://github.com/GoogleChromeLabs/Extension-Docs/issues/82)

- Refactors the HTTPXMLRequest page into a more generalized Cross Origin Requests page to include and prioritize fetch() in the code snippets and information.

### DIFF
--- a/site/en/docs/extensions/mv3/xhr/index.md
+++ b/site/en/docs/extensions/mv3/xhr/index.md
@@ -1,18 +1,13 @@
 ---
 layout: "layouts/doc-post.njk"
-title: "Cross-origin XMLHttpRequest"
-seoTitle: "Chrome Extensions: Cross-origin XMLHttpRequest"
+title: "Cross-origin Network Requests"
+seoTitle: "Chrome Extensions: Cross-origin Network Requests"
 date: 2012-09-18
-updated: 2020-03-09
-description: How to implement cross-origin XHR in your Chrome Extension.
+updated: 2023-05-23
+description: How to implement cross-origin network requests in your Chrome Extension.
 ---
 
-{% Aside 'warning' %}
-In Manifest V3, `XMLHttpRequest` is not supported in background pages (provided by Service Workers).
-Please consider using its modern replacement, `fetch()`.
-{% endAside %}
-
-Regular web pages can use the [`XMLHttpRequest`][1] object to send and receive data from remote
+Regular web pages can use the [`XMLHttpRequest`][1] or [fetch()][13] objects to send and receive data from remote
 servers, but they're limited by the [same origin policy][2]. [Content scripts][3] initiate requests
 on behalf of the web origin that the content script has been injected into and therefore content
 scripts are also subject to the [same origin policy][4]. (Content scripts have been subject to [CORB
@@ -23,15 +18,13 @@ its origin, as long as the extension requests cross-origin permissions.
 ## Extension origin {: #extension-origin }
 
 Each running extension exists within its own separate security origin. Without requesting additional
-privileges, the extension can use `XMLHttpRequest` to get resources within its installation. For
+privileges, the extension can use `fetch()` to get resources within its installation. For
 example, if an extension contains a JSON configuration file called `config.json`, in a
 `config_resources` folder, the extension can retrieve the file's contents like this:
 
 ```js
-var xhr = new XMLHttpRequest();
-xhr.onreadystatechange = handleStateChange; // Implemented elsewhere.
-xhr.open("GET", chrome.extension.getURL('/config_resources/config.json'), true);
-xhr.send();
+const response = await fetch('/config_resources/config.json');
+const jsonData =  await response.json();
 ```
 
 If the extension attempts to use a security origin other than itself, say https://www.google.com,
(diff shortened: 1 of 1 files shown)


site/en/docs/extensions/mv3/xhr/index.md
```diff
@@ -1,18 +1,13 @@
...
+updated: 2023-05-23
+description: How to implement cross-origin network requests in your Chrome Extension.
 ---
 
-{% Aside 'warning' %}
-In Manifest V3, `XMLHttpRequest` is not supported in background pages (provided by Service Workers).
-Please consider using its modern replacement, `fetch()`.
-{% endAside %}
-
-Regular web pages can use the [`XMLHttpRequest`][1] object to send and receive data from remote
```
jpmedley, line 15: In lieu of this, I think you need a new heading farther down.

- Give it a name like "What about XMLHttpRequest()?".
- Explain that XHR() is supported for backwards compatibility.
- New work should favor fetch().
- Extensions service workers don't support XHR, but calling it in an extension triggers the SW's fetch handler.

jpmedley:
@oliverdunk, I'll do a copy pass when your accuracy pass is done.

site/en/docs/extensions/mv3/xhr/index.md
```diff
@@ -86,56 +79,31 @@ When using resources retrieved via `XMLHttpRequest`, your background page should
...
 ```js
-var xhr = new XMLHttpRequest();
-xhr.open("GET", "https://api.example.com/data.json", true);
-xhr.onreadystatechange = function() {
-  if (xhr.readyState == 4) {
+const response = await fetch("https://api.example.com/data.json");
+const jsonData = await response.json();
     // WARNING! Might be injecting a malicious script!
-    document.getElementById("resp").innerHTML = xhr.responseText;
+    document.getElementById("resp").innerHTML = jsonData;
```
sebastianbenz, line 87: @IanStanion-google out of curiosity - shouldn't this count as executing remote hosted code? Does this even work in MV3?

jpmedley: It seems like this falls under the guidance to [externalize logic with a remove service](https://developer.chrome.com/docs/extensions/migrating/improve-security/#remove-remote-code:~:text=Externalized%20logic%20with%20a%20remote%20service). If there's a subtlety I'm missing, then somewhere on our site, we need to provide clarification.

site/en/docs/extensions/mv3/xhr/index.md
```diff
@@ -1,18 +1,13 @@
...
+description: How to implement cross-origin network requests in your Chrome Extension.
 ---
 
-{% Aside 'warning' %}
-In Manifest V3, `XMLHttpRequest` is not supported in background pages (provided by Service Workers).
-Please consider using its modern replacement, `fetch()`.
-{% endAside %}
-
-Regular web pages can use the [`XMLHttpRequest`][1] object to send and receive data from remote
+Regular web pages can use the [`XMLHttpRequest`][1] or [fetch()][13] objects to send and receive data from remote
```
oliverdunk, line 10: Can we switch the order of fetch and XMLHttpRequest to indicate some preference? And maybe say APIs rather than "objects".

site/en/docs/extensions/mv3/xhr/index.md
```diff
@@ -23,15 +18,13 @@ its origin, as long as the extension requests cross-origin permissions.
...
 example, if an extension contains a JSON configuration file called `config.json`, in a
 `config_resources` folder, the extension can retrieve the file's contents like this:
 
 ```js
-var xhr = new XMLHttpRequest();
-xhr.onreadystatechange = handleStateChange; // Implemented elsewhere.
-xhr.open("GET", chrome.extension.getURL('/config_resources/config.json'), true);
-xhr.send();
+const response = await fetch('/config_resources/config.json');
+const jsonData =  await response.json();
```
oliverdunk, line 27:
```suggestion
const jsonData = await response.json();
```

site/en/docs/extensions/mv3/xhr/index.md
```diff
@@ -214,6 +182,13 @@ If you modify the default [Content Security Policy][12] for your extension by ad
 you'd like to connect are allowed. While the default policy doesn't restrict connections to hosts,
 be careful when explicitly adding either the `connect-src` or `default-src` directives.
 
+### What about XMLHttpRequest?
+
+The `XMLHttpRequest()` (XHR) API is supported for backwards compatability, but new work should favor the use of `fetch()` wherever possible. The `fetch()` API allows you to make network requests similar to XMLHttpRequest, but has the additional benefit of performing asynchronously via promises.
```
oliverdunk, line 187: Was there a specific reason for adding this? I think we probably don't need it - `XMLHttpRequest` is a web API, not an extension API, and so I think it's expected understanding and more confusing to call it out.

site/en/docs/extensions/mv3/xhr/index.md
```diff
@@ -86,56 +79,31 @@ When using resources retrieved via `XMLHttpRequest`, your background page should
 fall victim to [cross-site scripting][9]. Specifically, avoid using dangerous APIs such as the
```
oliverdunk, line 79: I can't comment but on the line above we should replace the mention of `XMLHttpRequest`.

site/en/docs/extensions/mv3/xhr/index.md
oliverdunk, line 87: This seems to have some strange indentation.

oliverdunk (changes requested):
Thanks for updating this! Did an initial pass and left some comments.

site/en/docs/extensions/mv3/xhr/index.md
jpmedley: Yes. `fetch()` was created specifically for service workers and follows a broader web trend away from synchronous operations. Manifest V3 doesn't allow remotely hosted code, but it does allow [calling services](https://developer.chrome.com/docs/extensions/migrating/improve-security/#remove-remote-code:~:text=Externalized%20logic%20with%20a%20remote%20service). For that, developers may want and probably should use the extension service worker as a network proxy. That was the reason service workers were created in the first place. In light of this, does it make sense code outside the service worker to use a different interface?

With apologies to Ian (this is largely based on my suggestion) I think I should have suggested something slightly different for this. It should cover these points without making it sound as though the web platform is about to deprecate XHR. It definitely is not.

* Prefer `fetch()` because it's asynchronous and was made explicitly for service workers.
* `fetch()` handlers in a service worker will intercept `XMLHttpRequest()` calls.

jpmedley, line 187: >The `XMLHttpRequest()` (XHR) API...

Please call it by it's full name throughout.

site/en/docs/extensions/mv3/xhr/index.md
```diff
@@ -214,6 +182,13 @@ If you modify the default [Content Security Policy][12] for your extension by ad
 you'd like to connect are allowed. While the default policy doesn't restrict connections to hosts,
 be careful when explicitly adding either the `connect-src` or `default-src` directives.
 
+### What about XMLHttpRequest?
+
+The `XMLHttpRequest()` (XHR) API is supported for backwards compatability, but new work should favor the use of `fetch()` wherever possible. The `fetch()` API allows you to make network requests similar to XMLHttpRequest, but has the additional benefit of performing asynchronously via promises. 
+In Manifest V3, calling XHR will trigger the service-worker's fetch handler.
```
jpmedley, line 188:
```suggestion
In Manifest V3, calling `XMLHttpRequest()` will trigger the service worker's fetch handler.
```

site/en/docs/extensions/mv3/xhr/index.md
```diff
@@ -82,60 +75,35 @@ non-secure HTTP access to a given host or set of hosts, it must declare the perm
 
 ### Avoiding cross-site scripting vulnerabilities {: #xss }
 
-When using resources retrieved via `XMLHttpRequest`, your background page should be careful not to
+When using resources retrieved via `fetch()`, your background page should be careful not to
```
jpmedley, line 78:
```suggestion
When using resources retrieved via `fetch()`, your extension service worker should be careful not to
```

jpmedley (changes requested):

jpmedley (approved):

jpmedley (approved):

site/en/docs/extensions/mv3/xhr/index.md
```diff
@@ -214,6 +182,11 @@ If you modify the default [Content Security Policy][12] for your extension by ad
 you'd like to connect are allowed. While the default policy doesn't restrict connections to hosts,
 be careful when explicitly adding either the `connect-src` or `default-src` directives.
 
+### Using  Fetch() vs. XMLHttpRequest()
+
+`fetch()` was created specifically for service workers and follows a broader web trend away from synchronous operations. The `XMLHttpRequest()` API is supported, but calling `XMLHttpRequest()` will trigger the service worker's fetch handler. New work should favor the use of `fetch()` wherever possible.
```
oliverdunk, line 187: Tweaks this a bit to highlight:

- `XMLHttpRequest` is not available in service workers
- I removed the brackets from `XMLHttpRequest`. I know we usually do that for functions, but I think this is an exception - it's a constructor so `XMLHttpRequest()` without `new XMLHttpRequest()` is invalid syntax and I think the brackets make things more confusing.

```suggestion
`fetch()` was created specifically for service workers and follows a broader web trend away from synchronous operations. The `XMLHttpRequest` API is supported in extensions outside of the service worker, and calling `XMLHttpRequest` will trigger the service worker's fetch handler. New work should favor the use of `fetch()` wherever possible.
```

oliverdunk (approved):
LGTM with nit 👍

PR #6395, opened 2023-05-23
Add DRE Oliver Dunk to authorsData.json

AmySteam:
Adding in preparation to the Side Panel launch blog post


rachelandrew:
You need to add to the other file as well, that's why it isn't building: https://github.com/GoogleChrome/developer.chrome.com/blob/main/site/_data/i18n/authors.yaml

AmySteam:
@rachelandrew whoops. thanks for pointing this out. Added the changes to the yaml file. PTAL

rachelandrew (approved):

PR #6398, opened 2023-05-23
Added new manifest key page for background

IanStanion-google:
Fixes  https://github.com/GoogleChromeLabs/Extension-Docs/issues/84

- Adds new manifest key page for background with code snippet for consistency with the rest of the manifest key pages.


site/en/docs/extensions/mv3/manifest/background/index.md
```diff
@@ -0,0 +1,19 @@
...
+description: Reference documentation for the background property of manifest.json.
+---
+
+An optional manifest key used to specify a javascript file as the extension service worker. A service worker is a persistent script that acts as the extension's main event handler. For more information, visit the [more comprehensive introduction to service workers](docs/extensions/mv3/service_workers/#manifest).
+
+```json
+{
+  // ...
+   "background": {
+      "service_worker": "service-worker.js"
```
jpmedley, line 15:
```suggestion
      "service_worker": "service-worker.js",
      "type": "module"
```

jpmedley: This is a rare case where duplication is acceptable. The content at the end of the link optimizes for readers who need to understand extension service workers in whole or in part. A page like this is optimized for quick lookups. Ideally, there's more content that should be on a reference page, but this will do for now.

jpmedley (changes requested):

site/en/docs/extensions/mv3/manifest/background/index.md
```diff
@@ -0,0 +1,20 @@
+---
+layout: "layouts/doc-post.njk"
+title: "Manifest - background"
+date: 2023-05-24
+updated: 
+description: Reference documentation for the background property of manifest.json.
+---
+
+An optional manifest key used to specify a javascript file as the extension service worker. A service worker is a persistent script that acts as the extension's main event handler. For more information, visit the [more comprehensive introduction to service workers](docs/extensions/mv3/service_workers/#manifest).
```
AmySteam, line 9: The usage of the term "persistent" has the potential to frustrate some developers. We should consider expressing this differently. @oliverdunk WDYT?

jpmedley (approved):

site/en/docs/extensions/mv3/manifest/background/index.md
```diff
@@ -0,0 +1,20 @@
...
+
+```json
+{
+  // ...
```
AmySteam, line 13:
```suggestion
  ...
```
I think we can use only an ellipsis here

site/en/docs/extensions/mv3/manifest/background/index.md
```diff
@@ -0,0 +1,20 @@
...
+   "background": {
+      "service_worker": "service-worker.js",
+      "type": "module"
+    },
+  // ...
```
AmySteam, line 18:
```suggestion
  ...
```
Same here

site/en/docs/extensions/mv3/manifest/background/index.md
jpmedley: I should have caught that. I suggest 'background script'.

site/en/docs/extensions/mv3/manifest/background/index.md
AmySteam: I believe we should state explicitly when a field is optional, such as "type":"module".

site/en/docs/extensions/mv3/manifest/background/index.md
jpmedley: Neither is technically possible in JSON. It seems like Amy's version is more common.

site/en/docs/extensions/mv3/manifest/background/index.md
jpmedley: I agree.

site/en/docs/extensions/mv3/manifest/background/index.md
AmySteam: Sounds good 👍

oliverdunk: Looks like this is already fixed but I agree :)

AmySteam (approved):
LGTM 👍 Thanks @IanStanion-google!

PR #6400, opened 2023-05-23
Clarify MV2 sunset timeline

dotproto:
It's very easy to misread the dates on the current version of the Manifest V2 support timeline page. The bold text for the date makes it difficult to see the strikethrough. Because of this, it's easy for developers to incorrectly read the page as saying that we are only days away from Manifest V2 not running in Chrome Stable as previously communicated. 

This PR addresses this issue by:
* Removing the bold text from the struck-through dates.
* Normalizing the text below struck through dates to "Postponed".
* Bolding all"Postponed" labels.
* Making all "Postponed" labels link to the relevant chromium-extensions post.


oliverdunk:
Thanks for opening this. We had some discussion internally and agree that the current page needs to change, but want to do a slightly larger overhaul. I'm going to close this one and follow-up with a new PR shortly.

PR #6425, opened 2023-05-26
Correct web-vitals import

tunetheweb:

- Corrects the web-vitals attribution import which was incorrectly looking at `dist` directory. Both worked but better to do this properly.
- Updates web-vitals to latest
- Adds additional INP breakdown, deviceMemory, and ECT.


tunetheweb:
@matthiasrohmer any chance of a review and merge for this one? @rviscomi and I doing some INP debugging and want this update. FYI will be opening a web.dev equivalent one too.

rviscomi (commented):
Could we also add `deviceMemory` for all metrics while we're at it?

```
if (navigator.deviceMemory) {
  params.debug_device_memory = navigator.deviceMemory;
}
```

Maybe also ECT?

tunetheweb:
Added. But by pushing to data layer on page load like the other ones that don't change for each metric like `prefers-reduced-media` (I know some of these _can_ change during page life time, but think checking once per page load is sufficient).

matthiasrohmer (approved):

PR #6432, opened 2023-05-26
docs imply that you can get the value in the storage.set promise fulfillment. you can't.

patrickkettner:
[docs](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/StorageArea/set#return_value). `Value`, as currently used, is `undefined`


AmySteam (approved):
Thanks @patrickkettner! LGTM 👍

PR #6451, opened 2023-05-30
Add Compute Pressure origin trial 2 article

tomayac:

- Add Compute Pressure origin trial 2 article
- [Google doc](https://docs.google.com/document/d/1798PiWC2K7iIGDawrzvXttoxWyyFwOUziyEl6-tiFH8/edit)

### DIFF
--- a/site/en/blog/compute-pressure-origin-trial-2/index.md
+++ b/site/en/blog/compute-pressure-origin-trial-2/index.md
@@ -1,0 +1,202 @@
+---
+layout: 'layouts/blog-post.njk'
+title: 'Announcing the second Compute Pressure origin trial'
+description: >
+  A new origin trial from Chrome that allows web developers to monitor compute
+  pressure.
+authors:
+  - kenchris
+  - arskama
+date: 2023-05-30
+hero: image/FNkVSAX8UDTTQWQkKftSgGe9clO2/H9cVTFMj4cYmoIfGYQqd.jpg
+alt: A pressure gauge and pipes.
+tags:
+  - chrome-115
+  - origin-trials
+  - capabilities
+---
+
+Over the last year, Intel has been collaborating with Google and other parties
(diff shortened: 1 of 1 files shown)


tomayac:
@kenchris and @arskama, please take a final look, too. The staging preview link will be added any minute, or is already here when you read this. I have embedded the demo, which makes for a nice self-contained experience. If you have access to its sources, you may want to tweak the margins a bit.

rachelandrew (approved):

kenchris:
LGTM but maybe we want to give you some creds for the review and feedback :-) Can we add an ack section?

kenchris:
OK fixed the margins and added the OT token in the demo

arskama:
@tomayac Thanks for the PR, I wasn't expecting that you would edit it yourself.
Looks good. LGTM
I m a bit worried about the bad feedbacks, without the token and the proper flag, the embedded demo is crashing as we already mentioned. It will be only fixed in M115. 

https://chromium-review.googlesource.com/c/chromium/src/+/4478073

rachelandrew:
Maybe don't embed the demo but link to it with a note?

arskama:
> Maybe don't embed the demo but link to it with a note?

Yes, it used to be like that in the previous version.

I also agree with @kenchris on the credits, @tomayac and @rachelandrew can also be co-authors on that one.

kenchris:
If it doesn't crash when you don't have "Experimental web platform features" enabled, I think we are good. Developers who enable that (probably not that many) do know that that can lead to crashes. Also, this goes live with the origin trial and the new version right, so probably not a major issue

arskama:
> Also, this goes live with the origin trial and the new version right, so probably not a major issue

True

tomayac:
- Picture for Kenneth added.
- Credited Rachel and myself.
- Agreed on the embedded iframe crash not being an issue for the general public. 

Let's merge once tests have run.

PR #6455, opened 2023-05-30
Update Lighthouse TTFB documentation to Server Response Times

tunetheweb:
Fixes [#14255](https://github.com/GoogleChrome/lighthouse/issues/14255) as [someone else noticed this recently](https://stackoverflow.com/questions/76308397/how-can-i-have-a-bad-time-to-first-byte-on-core-web-vitals-but-a-good-ttfb-on-a/76313847?noredirect=1#comment134657196_76313847)

- Rename time-to-first-byte article to server-response-time
- Update image to latest screenshot of Lighthouse
- Add explanation aside
- Update some links


malchata (approved):
I meant to do this when the optimize TTFB guide came out, but got buried by other stuff. Thanks for catching and doing this.

tunetheweb:
@brendankenny any chance you could give this a review as you're more familiar with it?

site/en/docs/lighthouse/performance/server-response-time/index.md
```diff
@@ -23,6 +23,12 @@ for the server to respond to the main document request.
 Users dislike when pages take a long time to load.
 Slow server response times are one possible cause for long page loads.
 
+{% Aside 'important' %}
+Lighthouse usually measures _Server Response Time_, rather than the full [Time to First Byte (TTFB)](https://web.dev/ttfb/). As well as including the Server Response Time, TTFB can often include DNS lookups and redirects (for example, if the final slash or the www subdomain or https protocol is omitted then the server might redirect to the correct URL, or for URL shorteners or Ads that redirect via several domains). Most Lighthouse tools will test the end URL only, missing the DNS and redirect parts.
```
brendankenny, line 27: possible change:
```suggestion
Server Response Time is only part of the full [Time to First Byte (TTFB)](https://web.dev/ttfb/). As well as including the Server Response Time, TTFB can often include DNS lookups and redirects (for example, if the final slash or the www subdomain or https protocol is omitted then the server might redirect to the correct URL, or for URL shorteners or Ads that redirect via several domains). Most Lighthouse tools will test the end URL only, missing the DNS and redirect parts.
```

Current wording seems a little general considering this is the docs for the server response time audit :)

brendankenny (approved):
Thanks for the update!

My only slight hesitation is that TTFB is now in Lighthouse in the LCP breakdown, so there's no reason this audit couldn't be replaced with a full TTFB audit very soon. There's been no concrete timelines for that possibility, however, so it's not worth waiting around for.

LGTM!

PR #6466, opened 2023-05-31
Use correct term for 'synchronous style', update code to use ES2017

mikemaccana:

- 'synchronous style' can potentially cause confusion, the proper name for direct style is 'direct style'
- Use ES2017 rather than .then() chains.


mikemaccana:
Ping?

PR #6472, opened 2023-05-31
Removing false information from page

IanStanion-google:
Fixes [#SOME_ISSUE_NUMBER](https://github.com/GoogleChromeLabs/Extension-Docs/issues/85)

-Verified that information is incorrect and removed false statement.

### DIFF
--- a/site/en/docs/extensions/mv3/settings_override/index.md
+++ b/site/en/docs/extensions/mv3/settings_override/index.md
@@ -19,8 +19,6 @@ Google Search Console) by the same developer account publishing the extension. N
 verify ownership for a domain (for example, https://example.com) you can use any subdomain or page 
 (for example, https://app.example.com or https://example.com/page.html) within your extension.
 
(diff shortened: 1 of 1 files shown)


IanStanion-google:
@AmySteam I wasn't getting the stated confirmation dialog when requesting multiple permissions and also using override settings. I can maintain the information that this is against the single purpose policy, should I leave off the mention of that dialog box, or are you able to see that when you test?

AmySteam:
Just to double check, were you [loading the extension locally](https://developer.chrome.com/docs/extensions/mv3/getstarted/development-basics/#load-unpacked) or by [packing the extension](https://developer.chrome.com/docs/extensions/mv3/permission_warnings/#view_warnings)?

IanStanion-google:
My mistake... I was loading the extension locally and wasn't aware that only packed extensions received warnings. Sorry for the confusion!

jpmedley:
Does @AmySteam's question have any bearing on whether this paragraph can be removed?

AmySteam:
@IanStanion-google I would try packing the extension and testing it. If it doesn't work, I would follow-up with @dasli to confirm the behavior.

AmySteam (changes requested):
Please check with David before removing.

IanStanion-google:
The information removed in this PR was verified to still be correct.

PR #6498, opened 2023-06-03
change "extension service workers" to "web service workers"

### LINKED ISSUE #6497: extension service workers intro makes a reference to "extension service workers" that should be "web service workers"
The [About extension service workers](https://developer.chrome.com/docs/extensions/mv3/service_workers/) page makes a reference to "extension service workers" that should be "web service workers" instead.

The last sentence in the 3rd paragraph should read `They're also registered and updated differently from web service workers.` instead of `They're also registered and updated differently from extension service workers.`

hansifer:
Fixes #6497

### DIFF
--- a/site/en/docs/extensions/mv3/service_workers/index.md
+++ b/site/en/docs/extensions/mv3/service_workers/index.md
@@ -2,7 +2,7 @@
 layout: 'layouts/doc-post.njk'
 title: About extension service workers
 description: Extension service workers are an extension's central event handler. That makes them different from web service workers. 
-date: 2023-05-03
+date: 2023-06-03
 ---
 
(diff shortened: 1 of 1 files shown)


site/en/docs/extensions/mv3/service_workers/index.md
```diff
@@ -2,15 +2,15 @@
 layout: 'layouts/doc-post.njk'
 title: About extension service workers
 description: Extension service workers are an extension's central event handler. That makes them different from web service workers. 
-date: 2023-05-03
+date: 2023-06-03
```
jpmedley, line 5:
```suggestion
date: 2023-05-03
```

jpmedley: There's two reasons I changed this back. First, this isn't how we indicate content updates. Second, we don't do this for typo fixes.

jpmedley (changes requested):

jpmedley (approved):
Thank you for your help. This was a good catch.

PR #6508, opened 2023-06-04
Remove purpose: "any maskable" recommendation in Lighthouse PWA Audit docs

ekafyi:
Chrome DevTools now shows a warning in the "Manifest" section when `purpose` value is `any maskable`.

The [web.dev article](https://web.dev/maskable-icon/) on Maskable Icon also advises against it (quote at the end of article).

> While you can specify multiple space-separated purposes such as "any maskable", in practice you shouldn't. Using "maskable" icons as "any" icons is suboptimal as the icon is going to be used as-is, resulting in excess padding, making the core icon content smaller. 

Wouldn't it be better to not recommend `any maskable` altogether going forward?


malchata (approved):

rachelandrew:
This didn't get merged prior to the migration, I've just reapplied it over on the new platform so it should go live soon.

Issue #6459, opened 2023-05-30
Clarify purpose of OPFS in article


morsssss:
**Is your feature request related to a problem? Please describe.**
When I read [this helpful article on OPFS](https://developer.chrome.com/articles/origin-private-file-system/), I don't get a sense of what problem this was really created to solve.

**Describe the solution you'd like**
I think it would be great if the introduction talked about:
* OPFS allows a web app to store and manipulate files in its very own origin-specific virtual filesystem
* It lets you do low-level file manipulation, byte by byte, implementing (I think) part of the standard POSIX file interface
* Those are both new things on the web. File streams!
* Oh, and this is available cross-browser.

(Since it's cross-browser, should this be on developers.google.com?)

beaufortfrancois:
@tomayac FYI

PR #6517, opened 2023-06-05
[OPFS article] Migrate article to web.dev

tomayac:
Fixes https://github.com/GoogleChrome/developer.chrome.com/issues/6459.

- Add celebration aside.


tomayac:
Once this is merged, I will create a separate PR to migrate the article to web.dev.

rachelandrew:
@tomayac we don't have that aside here, which is why the build is broken.

Add it when you move the article, there's no need to add it first here as we don't have that tag on this site.

tomayac:
Ooops. I made the edit just before cooking lunch and didn’t realize the aside isn’t there. Will fix this after lunch.

rachelandrew (approved):
To merge once the one on web.dev is merged.

PR #6529, opened 2023-06-06
Topics API  partner guide

nancymic2:
Fixes b/269348075

- Publish partner Topics API integration guide
- https://pr-6529-static-dot-dcc-staging.uc.r.appspot.com/docs/privacy-sandbox/topics/integration-guide/
- will remove the do not merge label after reviews


samdutton (approved):
LGTM with tweaks as per commit.

rachelandrew:
@nancymic2 there are some links to supposedly larger versions of images that open the exact same size image that's in the page and also open a new tab, can we lose these? They aren't adding anything right now.

rachelandrew (approved):

nancymic2:
> 

Sure I can delete those links but they open much larger and readable on my machine (both monitors), which I why I added the links. Strange.

rachelandrew:
I'm not keen on having links opening in new tabs with no warning that is going to happen. It's an accessibility issue, you should always warn the user. So if you must have these they need to include a warning, "opens in new window" in the link text. However I don't think this is something we routinely do.

nancymic2:
> 
ok sure thanks. I've seen it in other pages this way, but they're deleted now.

PR #6530, opened 2023-06-06
Add 'What's new' item for multiple offscreen reasons.

jpmedley:


AmySteam (changes requested):
LGTM, just moved the date and added a header id.

AmySteam (approved):

PR #6531, opened 2023-06-06
Add examples to chrome.alarms.

jpmedley:
@IanStanion-google, Here's the kind of thing we eventually need to do throughout. I know the work related to this probably seems tedious. Reference that is complete and consistent is unfortunately necessary part of the job. Thank you for your help.


IanStanion-google (approved):
Looks good to me!

PR #6550, opened 2023-06-07
[ext-doc] Removes aside in Side Panel API

AmySteam:
Now that the Side Panel API is available in Chrome 114 Stable, this aside warning is no longer needed.


IanStanion-google (approved):
LGTM

PR #6632, opened 2023-06-16
Fix code typo in index.md

navorite:
chrome.extension.sendNativeMessage tab cell has no ending `.


AmySteam (approved):
Thanks @navorite! Good catch 👍 🙂

PR #6692, opened 2023-06-27
Fix HTML rendering of title in post navigation

beaufortfrancois:
This PR fixes an issue with HTML rendering of post titles in bottom page navigation. See live example at https://developer.chrome.com/blog/topics-enhancements/#:~:text=Back

image


matthiasrohmer (approved):
Thanks, @beaufortfrancois!

PR #6744, opened 2023-07-04
`OpfsDb` lives under `oo1`.

ahmad-moussawi:
Fixes the code example of initializing Opfs SQLite db.
it seems in recent versions, the `OpfsDb` lives under `oo1` and not under `opfs` object.
check https://github.com/tomayac/sqlite-wasm#in-a-worker-with-opfs-if-available


tomayac:
Thanks, but this was fixed by restructuring the code, please see the [current version of the snippet](https://developer.chrome.com/blog/sqlite-wasm-in-the-browser-backed-by-the-origin-private-file-system/#:~:text=const%20oo%20%3D%20sqlite3.oo1%3B%20//%20High%2Dlevel%20OO%20API%0A%20%20log(%27SQLite3%20version%27%2C%20capi.sqlite3_libversion()%2C%20capi.sqlite3_sourceid())%3B%0A%20%20let%20db%3B%0A%20%20if%20(%27OpfsDb%27%20in%20oo)%20%7B).

PR #6748, opened 2023-07-04
Update offscreen documents docs to use runtime.getContexts

oliverdunk:
Updates our offscreen documents docs to use runtime.getContexts, which is available starting in Chrome 116. I'm unsure if we should merge this until that is closer to beta but happy either way.

Fixes https://github.com/GoogleChromeLabs/Extension-Docs/issues/75

### DIFF
--- a/site/en/docs/extensions/reference/offscreen/index.md
+++ b/site/en/docs/extensions/reference/offscreen/index.md
@@ -28,7 +28,7 @@ Pages loaded as offscreen documents are handled differently from other types of 
 * Offscreen documents cannot have their `opener` property set using the [`chrome.windows` API][api-windows] method `windows.setSelfAsOpener()`.
 * An extension can only have one offscreen document open at a time. If the extension is running in split mode with an active incognito profile, both the normal and incognito profiles can each have one offscreen document. 
 
-Use [`chrome.offscreen.createDocument()`](#method-createDocument) and [`chrome.offscreen.closeDocument()`](#method-closeDocument) for creating and closing an offscreen document. Only a single Document can be open at a time. `createDocument` requires besides the document's `url`, reason and justification:
+Use [`chrome.offscreen.createDocument()`](#method-createDocument) and [`chrome.offscreen.closeDocument()`](#method-closeDocument) for creating and closing an offscreen document. Only a single Document can be open at a time. `createDocument()` requires the document's `url`, a reason, and a justification:
 
 ```js
 chrome.offscreen.createDocument({
--- a/site/en/docs/extensions/reference/offscreen/index.md
+++ b/site/en/docs/extensions/reference/offscreen/index.md
@@ -48,7 +48,12 @@ chrome.offscreen.createDocument({
 chrome.offscreen.closeDocument()
 ```
 
-The following example shows how to ensure that the offscreen document has already been created. The `setupOffscreenDocument()` function calls [`clients.matchAll()`](https://developer.mozilla.org/docs/Web/API/Clients/matchAll) to find the existing offscreen document or creates it if it doesn't already exist. Note that an extension can only have one offscreen document.
+The following example shows how to ensure that the offscreen document has already been created. The `setupOffscreenDocument()` function calls [`runtime.getContexts()`][runtime-get-contexts] to find the existing offscreen document or creates it if it doesn't already exist. Note that an extension can only have one offscreen document.
+
+{% Aside 'gotchas' %}
+[`runtime.getContexts()`][runtime-get-contexts] was added in Chrome 116. In earlier versions of
+Chrome, you can check for the existence of the offscreen document using [`clients.matchAll()`](https://developer.mozilla.org/docs/Web/API/Clients/matchAll).
+{% endAside %}
 
 ```js
 let creating; // A global promise to avoid concurrency issues
@@ -56,11 +61,13 @@ async function setupOffscreenDocument(path) {
   // Check all windows controlled by the service worker to see if one 
   // of them is the offscreen document with the given path
   const offscreenUrl = chrome.runtime.getURL(path);
-  const matchedClients = await clients.matchAll();
-  for (const client of matchedClients) {
-    if (client.url === offscreenUrl) {
-      return;
-    }
+  const existingContexts = await chrome.runtime.getContexts({
+    contextTypes: ['OFFSCREEN_DOCUMENT'],
+    documentUrls: [offscreenUrl]
+  });
+
+  if (existingContexts.length > 0) {
+    return;
   }
 
   // create offscreen document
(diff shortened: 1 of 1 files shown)


site/en/docs/extensions/reference/offscreen/index.md
```diff
@@ -48,19 +48,26 @@ chrome.offscreen.createDocument({
 chrome.offscreen.closeDocument()
 ```
 
-The following example shows how to ensure that the offscreen document has already been created. The `setupOffscreenDocument()` function calls [`clients.matchAll()`](https://developer.mozilla.org/docs/Web/API/Clients/matchAll) to find the existing offscreen document or creates it if it doesn't already exist. Note that an extension can only have one offscreen document.
+The following example shows how to ensure that the offscreen document has already been created. The `setupOffscreenDocument()` function calls [`runtime.getContexts()`][runtime-get-contexts] to find the existing offscreen document or creates it if it doesn't already exist. Note that an extension can only have one offscreen document.
+
+{% Aside 'gotchas' %}
+[`runtime.getContexts()`][runtime-get-contexts] was added in Chrome 116. In earlier versions of
+Chrome, you can check for the existence of the offscreen document using [`clients.matchAll()`](https://developer.mozilla.org/docs/Web/API/Clients/matchAll).
```
sebastianbenz, line 55: I'd be helpful to include the snippet here as it's not straightforward how to use it for an offscreen doc.

sebastianbenz (approved):
As you're mentioning the pre 116 workaround I think it's fine to merge this once the beta is live.

site/en/docs/extensions/reference/offscreen/index.md
sebastianbenz: I'd simply add the following snippet to the aside:

```
const matchedClients = await clients.matchAll();
   for (const client of matchedClients) {
     if (client.url === offscreenUrl) {
       return;
     }
   }
```

oliverdunk: Done!

site/en/docs/extensions/reference/offscreen/index.md
```diff
@@ -48,19 +48,36 @@ chrome.offscreen.createDocument({
 chrome.offscreen.closeDocument()
 ```
 
-The following example shows how to ensure that the offscreen document has already been created. The `setupOffscreenDocument()` function calls [`clients.matchAll()`](https://developer.mozilla.org/docs/Web/API/Clients/matchAll) to find the existing offscreen document or creates it if it doesn't already exist. Note that an extension can only have one offscreen document.
+The following example shows how to ensure that the offscreen document has already been created. The `setupOffscreenDocument()` function calls [`runtime.getContexts()`][runtime-get-contexts] to find the existing offscreen document or creates it if it doesn't already exist. Note that an extension can only have one offscreen document.
+
+{% Aside 'gotchas' %}
```
jpmedley, line 53: This really goes beyond what notes are intended for. I would do this. Put all of the current (i.e. post 116) stuff together worded as just the way to do it. Right after that, put a next level heading that titled something like "Before Chrome 116".

site/en/docs/extensions/reference/offscreen/index.md
```diff
@@ -48,19 +48,36 @@ chrome.offscreen.createDocument({
 chrome.offscreen.closeDocument()
 ```
 
-The following example shows how to ensure that the offscreen document has already been created. The `setupOffscreenDocument()` function calls [`clients.matchAll()`](https://developer.mozilla.org/docs/Web/API/Clients/matchAll) to find the existing offscreen document or creates it if it doesn't already exist. Note that an extension can only have one offscreen document.
+The following example shows how to ensure that the offscreen document has already been created. The `setupOffscreenDocument()` function calls [`runtime.getContexts()`][runtime-get-contexts] to find the existing offscreen document or creates it if it doesn't already exist. Note that an extension can only have one offscreen document.
```
jpmedley, line 51: As soon as you have multiple examples, you should have subheadings explaining what they are. That makes it easy to scan. Otherwise I need to read the whole of the examples section when I'm only interested in `getContexts()`.

jpmedley (changes requested):

site/en/docs/extensions/reference/offscreen/index.md
sebastianbenz: Moved into a new section as you suggested (taking this over from Oliver).

site/en/docs/extensions/reference/offscreen/index.md
jpmedley: Actually, it is 2 examples. A basic open and close and checking if one already exists. Now that I'm writing that, shouldn't the basic example include the checking? I think some consolidation is in order here, including removing the example that doesn't include the checking.

sebastianbenz: You're right, now I got it. The problem is not that it's two samples, but that we're mixing usage and example. I've added a usage section (merged the reasons section into it. PTAL

site/en/docs/extensions/reference/offscreen/index.md
```diff
@@ -18,49 +18,53 @@ You must declare the `"offscreen"` permission in the [extension manifest][doc-ma
...
-
-| Reason            | Offscreen Document Lifetime                    |
-|-------------------|------------------------------------------------|
-| AUDIO_PLAYBACK    | Closed after 30 seconds without audio playing. |
-| All other reasons | Not currently limited                          |
-
-## Example
-
-Use `chrome.offscreen.createDocument()` and `chrome.offscreen.closeDocument()` for creating and closing an offscreen document. Only a single Document can be open at a time. 
+Use [`chrome.offscreen.createDocument()`](#method-createDocument) and [`chrome.offscreen.closeDocument()`](#method-closeDocument) for creating and closing an offscreen document. Only a single Document can be open at a time. `createDocument` requires besides the document's `url`, reason and justification:
```
jpmedley, line 31:
```suggestion
Use [`chrome.offscreen.createDocument()`](#method-createDocument) and [`chrome.offscreen.closeDocument()`](#method-closeDocument) for creating and closing an offscreen document. Only a single Document can be open at a time. `createDocument()` requires the document's `url`, a reason, and a justification:
```

site/en/docs/extensions/reference/offscreen/index.md
```diff
@@ -18,49 +18,53 @@ You must declare the `"offscreen"` permission in the [extension manifest][doc-ma
...
   justification: 'reason for needing the document',
 });
-
-chrome.offscreen.closeDocument()
 ```
 
-The following example shows how to ensure that the offscreen document has already been created. The `setupOffscreenDocument()` function calls [`clients.matchAll()`](https://developer.mozilla.org/docs/Web/API/Clients/matchAll) to find the existing offscreen document or creates it if it doesn't already exist. Note that an extension can only have one offscreen document.
+### Reasons
+
+Find all valid reasons listed [below][offscreen-reason]. Reasons are set upon document creation to determine the document's lifespan:
```
jpmedley, line 43:
```suggestion
Find all valid reasons listed [below][offscreen-reason]. Reasons are set on document creation to determine the document's lifespan: 
```

jpmedley (changes requested):

jpmedley (approved):

PR #6750, opened 2023-07-04
Clarify that trusted testers should be Google accounts

oliverdunk:
Just a small clarification that trusted testers of Chrome extensions should be email addresses of Google accounts.


AmySteam (approved):
LGTM 👍

PR #6759, opened 2023-07-05
Lazy load iframes on scroll driven animation page

tunetheweb:

- Adds `loading=lazy` to iframes to improve performance


bramus (approved):

bramus:
Nit: attributes are not quoted, diverging from currently used style.

But don't let that block a merge ;)

tunetheweb:
Every byte counts @bramus !!!

PR #6769, opened 2023-07-06
Fix links, tweak title of Geolocation in extensions guide

AmySteam:
Fixes links and changes title to the geolocations. Plus a few minot grammatically fixes.

### DIFF
--- a/site/en/docs/extensions/mv3/geolocation/index.md
+++ b/site/en/docs/extensions/mv3/geolocation/index.md
@@ -23,9 +23,9 @@ Permissions are not the only difference. As mentioned above, `navigator.geolocat
 
 ### Use geolocation in service workers
 
-There is no `navigator` object inside of service workers. It is only available inside contexts that have access to a page's `document` object. To get access inside of a service worker, use an [`Offscreen Document`]offscreen]. This extension API gives you access to an HTML file that you can bundle with your extension.
+There is no `navigator` object inside of service workers. It is only available inside contexts that have access to a page's `document` object. To get access inside of a service worker, use an [`Offscreen Document`][offscreen], which provides access to an HTML file that you can bundle with your extension.
 
-To get started, add `offscreen` to your manifest's `permissions` section.
+To get started, add `"offscreen"` to your manifest's `"permissions"` section.
 
 {% Label %}manifest.json:{% endLabel %}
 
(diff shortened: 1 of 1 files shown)


site/en/docs/extensions/mv3/geolocation/index.md
```diff
@@ -1,31 +1,31 @@
...
+Permissions are not the only difference. As mentioned above, `navigator.geolocation` is a _DOM_ API, that is, something that is a part of the APIs that make up websites. As a result, it is not accessible inside worker contexts, like the [extension service worker][sw] that is the backbone of manifest v3 extensions. You can absolutely still use `geolocation`, though. There are just nuances with how and where you use it.
 
 
 ### Use geolocation in service workers
 
-There is no `navigator` object inside of service workers. It is only available inside of contexts that have access to a page's `document` object. To get access inside of a service worker, use an [`Offscreen Document`](offscreen). This is a feature of extensions that gives you access to an html file you can bundle with our extension.
+There is no `navigator` object inside of service workers. It is only available inside contexts that have access to a page's `document` object. To get access inside of a service worker, use an [`Offscreen Document`]offscreen]. This extension API gives you access to an HTML file that you can bundle with your extension.
```
jpmedley, line 26:
```suggestion
There is no `navigator` object inside of service workers. It is only available inside contexts that have access to a page's `document` object. To get access inside of a service worker, use an [`Offscreen Document`][offscreen], which provides access to an HTML file that you can bundle with your extension.
```

site/en/docs/extensions/mv3/geolocation/index.md
```diff
@@ -1,31 +1,31 @@
...
 
-To get started add `offscreen` to the `permissions` section of your manifest.
+To get started, add `offscreen` to your manifest's `permissions` section.
```
jpmedley, line 28:
```suggestion
To get started, add `"offscreen"` to your manifest's `"permissions"` section.
```

site/en/docs/extensions/mv3/geolocation/index.md
```diff
@@ -51,9 +51,9 @@ After adding the `offscreen` permission, add an html file to your extension that
 <script src="offscreen.js"></script>
 ```
 
-Save this file in the root of our project, as `offscreen.html`.
+Save this file in the root of your project as `offscreen.html`.
 
-As mentioned you need a script called `offscreen.js`. You'll also need to bundle this with your extension. It will be where you actually get geolocation information, and send that information back to the service worker. You can send and receive messages between it and your service worker.
+As mentioned, you need a script called `offscreen.js`. You'll also need to bundle this with your extension. It will be where you actually get geolocation information and send that information back to the service worker. You can send and receive messages between it and your service worker.
```
jpmedley, line 56:
```suggestion
As mentioned, you need a script called `offscreen.js`. You'll also need to bundle this with your extension. It will be the service worker's source of geolocation information. You can pass messages between it and your service worker.
```

site/en/docs/extensions/mv3/geolocation/index.md
```diff
@@ -190,13 +189,13 @@ So now whenever you want to get the geolocation from your service worker, you ju
...
+If you want to use geolocation within a [popup][popup] or [side panel][sidepanel], it is very straightforward. Popups and side panels are just web documents and therefore have access to the normal DOM APIs. You can access `navigator.geolocation` directly. The only difference from standard websites would be that you need to use the `manifest.json` `"permission"` field to request the `geolocation` permission. If you do not include the permission, you _will_ still have access to `navigator.geolocation`. However, any attempt to use it will cause an immediate error, the same as if the user rejected the request. You can see this in the [popup sample][popup-sample].
```
jpmedley, line 194:
```suggestion
Using geolocation within a [popup][popup] or [side panel][sidepanel] is very straightforward. Popups and side panels are just web documents and therefore have access to the normal DOM APIs. You can access `navigator.geolocation` directly. The only difference from standard websites is that you need to use the `manifest.json` `"permission"` field to request the `"geolocation"` permission. If you do not include the permission, you _will_ still have access to `navigator.geolocation`. However, any attempt to use it will cause an immediate error, the same as if the user rejected the request. You can see this in the [popup sample][popup-sample].
```

jpmedley: >If you do not include the permission, you _will_ still have access to `navigator.geolocation`. However, any attempt to use it will cause an immediate error

I'm guessing that unlike an extension API, calls to geolocation without permission don't throw and 'undefined' error. Who does this tidbit help and how does it help them? I think we could remove this text completely.

site/en/docs/extensions/mv3/geolocation/index.md
```diff
@@ -190,13 +189,13 @@ So now whenever you want to get the geolocation from your service worker, you ju
...
 
-### Using geolocation in a content script
+### Using geolocation in a content script {: #geo-content }
 
-Just like a popup, a [content script](content) has full access to the DOM API; however you will go through the normal user permission flow. That means that adding `geolocation` to your `permissions` _will not_ automatically give you access to the users' geolocation information. You can see this in the [content script sample](contentscript-sample)
+Just like a popup, a [content script][content] has full access to the DOM API; however, you will go through the normal user permission flow. That means that adding `"geolocation"` to your `permissions` _will not_ automatically give you access to the users' geolocation information. You can see this in the [content script sample][contentscript-sample].
```
jpmedley, line 198:
```suggestion
Just like a popup, a [content script][content] has full access to the DOM API; however, users will go through the normal user permission flow. That means that adding `"geolocation"` to your `"permissions"` _will not_ automatically give you access to the users' geolocation information. You can see this in the [content script sample][contentscript-sample].
```

jpmedley (dismissed):
Amy,

All of your work looks good. I spotted some other things we can fix while we're here, and since this is so short.

Joe

AmySteam:
@jpmedley PTAL 🙏✨

jpmedley (approved):

PR #6798, opened 2023-07-11
Update index.md

induratized:
fetch with event.request.url instead of event.request

Fixes #SOME_ISSUE_NUMBER

-
-
-


malchata:
[Per MDN](https://developer.mozilla.org/en-US/docs/Web/API/FetchEvent/request), it's valid syntax to pass `event.request`. I believe `event.request.url` works as well, but I'm not sure if this is a necessary change.

PR #6812, opened 2023-07-12
Remove minimum_chrome_version from DevTools article

jpmedley:
When the article was written, there was a possibility that someone would try to use DevTools APIs in an unsupported version (earlier than 10). In MV3, this isn't even possible.

Fixes https://github.com/GoogleChromeLabs/Extension-Docs/issues/84


AmySteam (approved):
LGTM 👍

PR #6825, opened 2023-07-14
PostMessage for trusted web activity article

tropicadri:
Implementation guide for postMessage for trusted web activity.
Plus adding Sayed as author and fixing errors with my author info.

### DIFF
--- a/site/en/articles/post-message-twa/index.md
+++ b/site/en/articles/post-message-twa/index.md
@@ -1,0 +1,196 @@
+---
+layout: 'layouts/blog-post.njk'
+title: PostMessage for TWA
+description: >
+  From Chrome 115 Trusted Web Activities (TWA) can send messages using postMessages. This article walks through the setup needed to communicate between your app and the web.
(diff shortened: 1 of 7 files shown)


site/en/articles/post-message-twa/index.md
```diff
@@ -0,0 +1,196 @@
...
+// Handle the received message.
+
+ }
+};
+```
+
+
+## Communicating from the web
+
+Now we can send and receive messages from our host app, how do we do the same from the web? Communication has to start from the host app, then the web page needs to get the port from the first message. This port is used to communicate back. Your js file will look something like the following example:
```
rachelandrew, line 166:
```suggestion
Now we can send and receive messages from our host app, how do we do the same from the web? Communication has to start from the host app, then the web page needs to get the port from the first message. This port is used to communicate back. Your JavaScript file will look something like the following example:
```

rachelandrew (approved):
Update the date before merging!

PR #6833, opened 2023-07-15
Update index.md

mdee5test:
Fixes #SOME_ISSUE_NUMBER

- Change the variable name "session" - which is nonexistent - to "mSession" which is the correct global variable being referenced according to the code flow. Using "session" led to errors and made the code execution to fail. After correction, code executed successfully.


sebastianbenz (approved):
Thanks for fixing!

AmySteam (approved):

PR #6872, opened 2023-07-20
Update aside in MV2 articles

AmySteam:
Fixes #https://github.com/GoogleChromeLabs/Extension-Docs/issues/136

- Adds aside to MV2 articles that links directly to the MV3 equivalent.

PS: It is not a partial, because I wanted developers to be able to go directly to the mv3 equivalent page.


AmySteam:
@sebastianbenz I did only the a11y article for now. Once you review it, I'll update the rest.

Screenshot 2023-07-20 at 07 20 53

sebastianbenz (approved):
Can we turn this into a partial?

jpmedley (approved):

PR #6880, opened 2023-07-20
Correct script name in "Communication with the embedding page"

alvinjiooo:
It seems to use the wrong name for the script which should run in example.html. This PR correct the name for it.

Fixes #SOME_ISSUE_NUMBER

    - correct the name of the script which runs in example.html for "Communication with the embedding page".


AmySteam (approved):
Thanks @alvinjiooo, good catch! 😃💯

PR #6904, opened 2023-07-22
Add TestingBot cloud service for physical device testing

jochen-testingbot:
Adds another option to use browsers and physical devices with TestingBot, for testing on other browsers & devices.


rachelandrew:
Please don't add links to commercial services to our documentation without request from a member of the Chrome team.

Issue #97, opened 2020-12-17
Make the diagram on docs/webstore/authentication accessible


robdodson:
@kaycebasques commented on [Mon Nov 16 2020](https://github.com/GoogleChrome/dcc-private/issues/268)

The image in `localhost:8080/docs/webstore/authentication#sample-workflow` should be converted to text. There is a lot of important information in there that is not accessible.

robdodson:
This would be nice but I can definitely understand if there are higher priorities

jpmedley:
There is a closed PR, so closing this ticket.

Issue #80, opened 2020-12-17
Make responsive tables more accessible


robdodson:
@robdodson commented on [Wed Dec 02 2020](https://github.com/GoogleChrome/dcc-private/issues/378)

We're already doing some of this, but Adrian has a good post on some additional things we can do to make our overflow tables more accessible.

https://adrianroselli.com/2020/11/under-engineered-responsive-tables.html

samthor:
Rob, what do you think of this vs #38? Should we instead try to prevent overflow tables where possible?

If we stick with overflow tables, it seems like we should have them intentionally extend _more_, as right now it's possible for some columns to be very tiny and make text hard to read.

robdodson:
Commented over in https://github.com/GoogleChrome/developer.chrome.com/issues/38#issuecomment-764158386

PR #6921, opened 2023-07-25
Overhaul documentation for declarative net request

jpmedley:
Fixes [#80](https://github.com/GoogleChromeLabs/Extension-Docs/issues/80) and [#97](https://github.com/GoogleChromeLabs/Extension-Docs/issues/80)

Out of scope: Improvements that go beyond the information provided in the existing article.


jpmedley:
@oliverdunk, you are now free to tackle this.

site/en/docs/extensions/reference/declarativeNetRequest/index.md
```diff
@@ -9,19 +9,22 @@ has_warning: One or more of these permissions <a href="/docs/extensions/mv3/perm
...
-[manifest][1] to use this API. The former allows extensions to block and upgrade requests without
-any [host permissions][2]. Host permissions are still required if the extension wants to redirect a
-request or modify headers on it. The `declarativeNetRequestWithHostAccess` permission always
-requires host permissions to the request URL and initiator to act on a request.
+### Permissions {: #permission-definitions }
 
-The `declarativeNetRequestFeedback` permission is required to access functions and events which
-return information on declarative rules matched.
+`declarativeNetRequest`
+: Allows extensions to block and upgrade requests.
```
oliverdunk, line 15: With how this is currently written, it implies to me that `declarativeNetRequest` is less capable than `declarativeNetRequestWithHostAccess`. Could we clarify that with sufficient host permissions both are equivalent (and can redirect requests and modify headers)?

site/en/docs/extensions/reference/declarativeNetRequest/index.md
```diff
@@ -9,19 +9,22 @@ has_warning: One or more of these permissions <a href="/docs/extensions/mv3/perm
...
-a list containing dictionaries of type [Ruleset][3], as shown below.
+`declarativeNetRequestWithHostAccess`
+: Allows extensions to redirect requests and modify request and response headers. This permission always requires [host permissions](/docs/extensions/mv3/declare_permissions/) to the requestED URL and initiator to act on a request.
+
+`declarativeNetRequestFeedback`
+: Allows access to functions and events that return information on matched rules during debugging. This permission only works in [unpacked extensions](/docs/extensions/mv3/getstarted/development-basics/#load-unpacked).
+
+Certain rules such as redirect and header modification rules also require host permissions that match the network requests that these rules apply to.
+
+### Static rulesets {: #manifest-rule-resources }
```
oliverdunk, line 25: It feels like this should come after the concepts section where we explain what a ruleset is?

site/en/docs/extensions/reference/declarativeNetRequest/index.md
```diff
@@ -51,57 +54,63 @@ a list containing dictionaries of type [Ruleset][3], as shown below.
 }
 ```
 
-### Rule Resources {: #manifest-rule-resources }
+## Concepts and Usage {: #concepts-and-usage }
 
-Rules are specified in JSON files referenced under the `"rule_resources"` manifest key.
+To use declarative net requests specify one or more rulesets. A ruleset contains an array of rules. A single rule does one of the following:
```
oliverdunk, line 59: I think "declarative net requests" should be either the API name or namespace, since it's not a thing that makes sense to pluralise.

oliverdunk, line 59: A rule can actually do five things (e.g upgrade requests). Should we say something that makes it clear this is not an exhaustive list?

site/en/docs/extensions/reference/declarativeNetRequest/index.md
```diff
@@ -51,57 +54,63 @@ a list containing dictionaries of type [Ruleset][3], as shown below.
...
+-  Redirect a network request.
+-  Modify request or response headers.
 
-**Note:** Errors and warnings about invalid static rules are only displayed for unpacked extensions.
-Invalid static rules in packed extensions are ignored. It's therefore important to verify that your
-static rulesets are valid by testing with an unpacked version of your extension.
+There are three types of rulesets.
 
-## Limits {: #limits }
+-  Static rules
```
oliverdunk, line 68:
```suggestion
-  Static rulesets
```

etc.?

site/en/docs/extensions/reference/declarativeNetRequest/index.md
```diff
@@ -51,57 +54,63 @@ a list containing dictionaries of type [Ruleset][3], as shown below.
...
-There is a performance overhead to loading and evaluating rules in the browser,
-and so a number of limits apply when using the API. Limits depend on the type of
-rule you're using.
+### Dynamic and session-scoped rulesets {: #dynamic-and-session-rules }
+
+Dynamic and session-scoped rulesets are managed using JavaScript while an extension is in use.
 
-### Static rules {: #static-rules }
+-  Dynamic rules persist across browser sessions and extension upgrades.
+-  Session rules are destroyed when the browser shuts down and when a new version of the extension is installed.
```
oliverdunk, line 77:
```suggestion
-  Session rules are cleared when the browser shuts down and when a new version of the extension is installed.
```

site/en/docs/extensions/reference/declarativeNetRequest/index.md
```diff
@@ -158,102 +164,139 @@ examples of URL filters:
...
 
-An extension can add or remove rules dynamically using the [`updateDynamicRules()`][9] and the [`updateSessionRules()`][13] API methods.
-- Rules added using the [`updateDynamicRules()`][9] API method are persisted across both sessions and extension updates.
-- Rules added using the [`updateSessionRules()`][13] API method are not persisted across Chrome sessions. These rules are backed in memory by Chrome.
-- The number of dynamic and session-scoped rules that an an extension can add is bounded by the [`MAX_NUMBER_OF_DYNAMIC_AND_SESSION_RULES`][10] constant.
+Thinking of matching this way: whatever rule a particular extension prioritizes will then be prioritized against rules from other extensions.
 
-## Updating enabled rulesets {: #update-enabled-rulesets }
+{% Aside %}
+Avoid building rulesets that depend on rules with the same priority running in a particular order. Rules with the same priority in the same set (e.g static rules with priority 1) are executed in an arbitrary order which is not defined and can change between runs or browser versions. This guidline applies between browsers as well.
```
oliverdunk, line 177:
```suggestion
Avoid building rulesets that depend on rules with the same priority running in a particular order. Rules with the same priority in the same set (e.g static rules with priority 1) are executed in an arbitrary order which is not defined and can change between runs or browser versions. This is the case across browsers.
```

site/en/docs/extensions/reference/declarativeNetRequest/index.md
```diff
@@ -158,102 +164,139 @@ examples of URL filters:
...
-  of rules an extension can still enable before the global limit is reached.
-- The set of enabled static rulesets is persisted across sessions but not across extension updates.
-  The `"rule_resources"` manifest key will determine the set of enabled static rulesets on initial
-  extension install and on each subsequent extension update.
+Within a single extension, prioritization is worked out using the following process:
 
-## Implementation details {: #implementation }
+1. The rule with the highest developer-defined priority (in other words, the `"priority"` field) is returned.
+1. If there is more than one rule with the highest developer-defined priority, rules are prioritized using the `"action"` field, in the following order:
+    -  `allow`
```
oliverdunk, line 186: Maybe an numbered list would make sense here?

site/en/docs/extensions/reference/declarativeNetRequest/index.md
```diff
@@ -158,102 +164,139 @@ examples of URL filters:
...
+    -  `allowAllRequests`
+    -  `block`
+    -  `upgradeScheme`
+    -  `redirect`
 
-### web_accessible_resources {: #implementation-web-accessible-resources }
+1. If the action type is not `block` or `redirect`, any matching `modifyHeaders` rules are evaluated. Be aware that if there are any rules with a developer-defined priority lower than the priority specified for `allow` and `allowAllRequests`, such rules are ignored.
```
oliverdunk, line 192: TIL. Are we comfortable this is true?

site/en/docs/extensions/reference/declarativeNetRequest/index.md
```diff
@@ -158,102 +164,139 @@ examples of URL filters:
...
-- If a rule has set a header, then lower priority rules cannot further modify the header, except for
-  `append` rules from the same extension.
-- If a rule has removed a header, then lower priority rules cannot further modify the header.
+#### Dynamic and session rules {: #dynamic-session-rules }
+
+The limits applied to dynamic and session rules are simpler than static rules. The total number of both cannot exceed **5000**. This is called the [`MAX_NUMBER_OF_DYNAMIC_AND_SESSION_RULES`][#property-MAX_NUMBER_OF_DYNAMIC_AND_SESSION_RULES].
+
+#### Rules that use regex {: #regex-rules }
+
+All types of rules can use regular expressions; however the total number of regex rules, of each type cannot exceed **1000**. This is called the [MAX_NUMBER_OF_REGEX_RULES](#property-MAX_NUMBER_OF_REGEX_RULES).
```
oliverdunk, line 229: There's also a memory limit on RegEx rules (I believe 2KB) that we should check and document.

site/en/docs/extensions/reference/declarativeNetRequest/index.md
```diff
@@ -158,102 +164,139 @@ examples of URL filters:
...
+If a rule matches a resource stored in cache by a host page's web service worker, that rule may be ignored until the resource is cleared from the cache. This is because cached storage is intended to be persistent, and many features such as offline use do not expect the cache to be cleared without also clearing a service worker's registration. For your rules to work properly, you need to call [`browsingData.removeCache()`](/docs/extensions/reference/browsingData/#method-removeCache) or [`browsingData.removeCacheStorage()`](/docs/extensions/reference/browsingData/#method-removeCacheStorage) after installation of your extension. You may need to call one of these functions after an extension upgrade, depending on whether you've changed your rules.
```
oliverdunk, line 233: Note: I think we still need to discuss this one.

site/en/docs/extensions/reference/declarativeNetRequest/index.md
```diff
@@ -158,102 +164,139 @@ examples of URL filters:
...
 
-## Example {: #example }
+### Web accessible resources {: #implementation-web-accessible-resources }
 
-{% Label %}manifest.json:{% endLabel %}
+A declarative net request cannot redirect from a public resource request to a resource that is not web accessible. Doing so triggers an error. This is true even if the specified web accessible resource is owned by the redirecting extension. To declare resources for declarative net requests, use the manifest's [`"web_accessible_resources"`](/docs/extensions/mv3/manifest/web_accessible_resources/) array.
```
oliverdunk, line 237:
```suggestion
A declarative net request rule cannot redirect from a public resource request to a resource that is not web accessible. Doing so triggers an error. This is true even if the specified web accessible resource is owned by the redirecting extension. To declare resources for declarative net requests, use the manifest's [`"web_accessible_resources"`](/docs/extensions/mv3/manifest/web_accessible_resources/) array.
```

oliverdunk (changes requested):
Left some initial comments. Overall this is a huge improvement though, super excited about it :D

site/en/docs/extensions/reference/declarativeNetRequest/index.md
jpmedley: We can, but I need some clarification because this definition regurgitates the current definition. Your explanation makes it sound as though there's no difference between the permissions. 

What's the difference between the two permissions?

oliverdunk: Hopefully our discussion yesterday was helpful :)

site/en/docs/extensions/reference/declarativeNetRequest/index.md
```diff
@@ -158,102 +150,139 @@ examples of URL filters:
...
-    "*://*.headers.com/*",
-  ],
-  "manifest_version": 3
+#### Rules that use regex {: #regex-rules }
+
+All types of rules can use regular expressions; however the total number of regex rules, of each type cannot exceed **1000**. This is called the [MAX_NUMBER_OF_REGEX_RULES](#property-MAX_NUMBER_OF_REGEX_RULES).
+
+### Interactions with service workers {: #interact-w-service-workers }
+
+Requests and responses intercepted by the `onfetch` handler in a host page's service worker are not affected by declarativeNetRequet. Consequently such service workers can make it appear as though a rule doesn't work. To properly test your rules, you may need to disable or remove one or more of the service workers in your installation of Chrome.
```
sebastianbenz, line 219: Do I understand this right that publishers could simply handle every request in the `onfetch` handler to avoid ad blockers functioning on their pages?

redirects.yaml
```diff
@@ -465,6 +465,18 @@ redirects:
 - from: /docs/extensions/declarativenetrequest
   to: /docs/extensions/reference/declarativeNetRequest
 
+- from: /docs/extensions/reference/declarativeNetRequest#implementation
```
sebastianbenz, line 468: Do redirects work for hash tags?

site/en/docs/extensions/reference/declarativeNetRequest/index.md
oliverdunk: So I was trying to avoid opening this can of worms and I think the above text is still vague enough to sidestep this - but if you call fetch in a service worker, those requests are affected by DNR. So even if a web service worker intercepts a request, the only way it can meaningfully respond is (a) if it already has a response cached or (b) if it does a fetch to make the request anyway, and DNR will apply. Hopefully that makes sense.

sebastianbenz: We need to add this detail otherwise it is very misleading.

jpmedley: That opens another can of worms. If I understand what you're saying correctly, whether and how dNR works depends on which [caching algorithm](https://web.dev/offline-cookbook/#serving-suggestions) a host page uses.

- *Cache only*- dNR will always fail
- *Network only*- dNR will always succeed
- *Cache, falling back to network*- dNR will fail when the item is in the cache, succeed otherwise
- *Cache and network race*- Who knows what dNR will do?
- *Network falling back to cache*- Whenever there's an item in the cache, dNR will appear to fail, but succeed otherwise.
- *Generic fallback*- dNR will appear to fail when the network can't be reached.
- *Service worker templating*- dNR will appear to fail.

I can't see how an extension developer would design around this.

jpmedley: There must be some very good reason why dNR doesn't affect items retrieved from `CacheStorage`.

oliverdunk: @jpmedley - all of the above sounds right 👍 

I don't feel quite as pessimistic about being able to document this. How about the following (changes in bold)?

> Requests and responses intercepted by the `onfetch` handler in a host page's service worker are not affected by declarativeNetRequest. Consequently such service workers can make it appear as though a rule doesn't work. To properly test your rules, you may need to disable or remove one or more of the service workers in your installation of Chrome. **Note that DNR does apply if the service worker goes on to make subsequent requests to the network, even if these are to the same path that was previously intercepted.**

On your other point:

> I can't see how an extension developer would design around this.

I think this depends on the developer's goal. If you're making something to block trackers, a request to /track_user that never reaches the network is useless, so the developers of those websites aren't going to enable any caching. In that case you can probably just ignore all of this.

If you are trying to make something to block a page from loading, you have bigger problems than this. DNR doesn't have any affect on single page apps that use `history.pushState` to change what is showing in the address bar for example.

So TLDR, I think making developers aware of this gotcha is important, but I don't expect it actually causes many issues in practice.

sebastianbenz: I don't thinks this phrasing is correct on it's own:

> Requests and responses intercepted by the `onfetch` handler in a host page's service worker are not affected by declarativeNetRequest. 


Because they are affected, if the sw makes a subsequent fetch request. I still like Oliver's original phrasing, we could add an example to make it more clear:

```
DNR only applies to requests that reach the network stack. This does include responses from the HTTP cache, but does not include responses generated by a website's service worker. For example, if a service worker intercepts a request in an `onFetch` handler and returns a cached or generated response, the response will not be affected by your DNR rules. However, if the service worker makes a subsequent `fetch` request, your DNR rules will apply to this new request.
```

site/en/docs/extensions/reference/declarativeNetRequest/index.md
```diff
@@ -158,102 +150,139 @@ examples of URL filters:
...
-    "*://*.headers.com/*",
-  ],
-  "manifest_version": 3
+#### Rules that use regex {: #regex-rules }
+
+All types of rules can use regular expressions; however the total number of regex rules, of each type cannot exceed **1000**. This is called the [MAX_NUMBER_OF_REGEX_RULES](#property-MAX_NUMBER_OF_REGEX_RULES).
+
+### Interactions with service workers {: #interact-w-service-workers }
+
+declarativeNetRequest only applies to requests that reach the network stack. This includes responses from the HTTP cache, but may not include responses that go through a service worker's `onfetch` handler. declarativeNetRequest won't affect responses generated by the service worker or retrieved from `CacheStorage`, but it will affect calls to `fetch()` made in a service worker.
```
jpmedley, line 219: New language.

@oliverdunk @sebastianbenz

sebastianbenz: Looks good!

oliverdunk: Sounds great! 🎉

site/en/docs/extensions/reference/declarativeNetRequest/index.md
oliverdunk: Confirmed that the memory is limited to 2KB here: https://source.chromium.org/chromium/chromium/src/+/refs/heads/main:extensions/browser/api/declarative_net_request/utils.cc;l=317;drc=0340dd86c6c1f01ccd1a0b983633cfef5fcf6697

You get an error that looks like this (which is somewhat helpful, although it doesn't tell you what "complex" means behind the scenes 😔):

Screenshot 2023-08-01 at 11 03 21

jpmedley: I'm open to wording suggestions.

redirects.yaml
AmySteam: You can use an anchor redirect
https://developer.chrome.com/docs/handbook/how-to/create-anchor-redirects/

jpmedley: I can't work out from those instructions what I'm supposed to do.

sebastianbenz: Sweet, I wasn't aware that we support anchor redirects. The yaml file defines server-side redirects which are not supported for anchor links. Hence the client-side based workaround which needs to be defined in the page frontmatter itself:

```
---
...
anchorRedirects:
  implementation: /docs/extensions/reference/declarativeNetRequest#implementation-web-accessible-resources
---
```

jpmedley: A redirect needs two targets. This only has one.

sebastianbenz: image

jpmedley: I would expect that to only work when the anchor name is the same. I reused anchor names wherever I could. But there are several cases where it just made no sense.

jpmedley: Plus how would the server know to look in /recorder/reference for a redirect of something in recorder/?

sebastianbenz: This is handled client side afaik.

site/en/docs/extensions/reference/declarativeNetRequest/index.md
oliverdunk: Hmm, now it says "There are three types of rulesets." and then lists "static rules". That feels wrong too?

site/en/docs/extensions/reference/declarativeNetRequest/index.md
oliverdunk: Actually, I dug in to this more and it looks like you're right. TIL!

site/en/docs/extensions/reference/declarativeNetRequest/index.md
oliverdunk: > Additionally, each rule must be less than 2KB once compiled. This correlates with the general complexity of the rule specified. If you try to load a rule which exceeds this limit, you will see a warning and the rule will be ignored.

First draft to work from :)

site/en/docs/extensions/reference/declarativeNetRequest/index.md
jpmedley: I thought I had removed the word 'rules'.

site/en/docs/extensions/reference/declarativeNetRequest/index.md
jpmedley: Added the paragraph, plus an example of the warning (in case someone google's the text of the warning).

site/en/docs/extensions/reference/declarativeNetRequest/index.md
```diff
@@ -158,102 +150,145 @@ examples of URL filters:
...
-  ],
-  "manifest_version": 3
+#### Rules that use regex {: #regex-rules }
+
+All types of rules can use regular expressions; however the total number of regex rules, of each type cannot exceed **1000**. This is called the [MAX_NUMBER_OF_REGEX_RULES](#property-MAX_NUMBER_OF_REGEX_RULES).
+
+Additionally, each rule must be less than 2KB once compiled. This roughly correlates with the complexity of the rule. If you try to load a rule that exceeds this limit, you will see a warning like the one below and the rule will be ignored.
+
+```bash
+rules_1.json: Rule with id 1 specified a more complext regex than allowed as part of the "regexFilter" key.
```
oliverdunk, line 220:
```suggestion
rules_1.json: Rule with id 1 specified a more complex regex than allowed as part of the "regexFilter" key.
```

site/en/docs/extensions/reference/declarativeNetRequest/index.md
oliverdunk: Love it!

jpmedley:
@oliverdunk and @sebastianbenz Are you done with this? I'd like to get a copy edit from Amy now.

site/en/docs/extensions/reference/declarativeNetRequest/index.md
```diff
@@ -51,57 +39,64 @@ a list containing dictionaries of type [Ruleset][3], as shown below.
...
 
-### Rule Resources {: #manifest-rule-resources }
+## Concepts and Usage {: #concepts-and-usage }
 
-Rules are specified in JSON files referenced under the `"rule_resources"` manifest key.
+To use this API specify one or more rulesets. A ruleset contains an array of rules. A single rule does one of the following:
 
-Each file should contain an array of [rules](#rules) as in the [example](#example).
+-  Block a network request.
+-  Upgrade a request.
```
sebastianbenz, line 47: Can you maybe add a short explanation what `upgrade` means (at least I don't know what upgrade means in this context).

site/en/docs/extensions/reference/declarativeNetRequest/index.md
```diff
@@ -158,102 +150,145 @@ examples of URL filters:
...
 
-When rules are applied to browsers with pages in the service worker's cached storage, the browser may ignore the set rule for those specific pages until the cached storage is cleared. This is because cached storage is intended to be persistent, and many features like offline use do not expect the cache to be cleared without also clearing a service worker's registration as well. For cases when extensions utilizing declarativeNetRequest must be enabled and disabled repeatedly, the [`chrome.browsingData`](/docs/extensions/reference/browsingData/) API may be used to clear the cache to guarantee proper functionality.
+#update-static-rulesets
```
sebastianbenz, line 207: Is this supposed to be a heading?

site/en/docs/extensions/reference/declarativeNetRequest/index.md
```diff
@@ -158,102 +150,145 @@ examples of URL filters:
...
 
-## Example {: #example }
+#### Dynamic and session rules {: #dynamic-session-rules }
 
-{% Label %}manifest.json:{% endLabel %}
+The limits applied to dynamic and session rules are simpler than static rules. The total number of both cannot exceed **5000**. This is called the [`MAX_NUMBER_OF_DYNAMIC_AND_SESSION_RULES`][#property-MAX_NUMBER_OF_DYNAMIC_AND_SESSION_RULES].
```
sebastianbenz, line 211:
```suggestion
The limits applied to dynamic and session rules are simpler than static rules. The total number of both cannot exceed **5000**. This is called the [`MAX_NUMBER_OF_DYNAMIC_AND_SESSION_RULES`](#property-MAX_NUMBER_OF_DYNAMIC_AND_SESSION_RULES).
```

site/en/docs/extensions/reference/declarativeNetRequest/index.md
sebastianbenz, line 220: Can we make the `pre` wrap the content here? Otherwise maybe add linebreaks.

site/en/docs/extensions/reference/declarativeNetRequest/index.md
```diff
@@ -158,102 +150,145 @@ examples of URL filters:
...
+
+#### The "priority" key {: #priority-key }
+
+These examples require the following host permission:
+
+`*://*.example.com/*`
+
+To work out the priority of a particular URL, look at the (developer-defined) `"priority"` key, the `"action"` key and the `"urlFilter"` key. These examples refer to the example rule file shown below them.
+
+**Navigation to https://google.com**
```
sebastianbenz, line 284: The formatting looks a bit weird when rendered to html as the linebreak disappears.

sebastianbenz (approved):
Looks great! Left a few comments that should be addressed.

site/en/docs/extensions/reference/declarativeNetRequest/index.md
jpmedley: Done.

site/en/docs/extensions/reference/declarativeNetRequest/index.md
jpmedley: Removed. It's an old anchor.

site/en/docs/extensions/reference/declarativeNetRequest/index.md
jpmedley: Fixed.

site/en/docs/extensions/reference/declarativeNetRequest/index.md
jpmedley: Done.

site/en/docs/extensions/reference/declarativeNetRequest/index.md
jpmedley: Fixed.

jpmedley:
@sebastianbenz PTAL

site/en/docs/extensions/reference/declarativeNetRequest/index.md
```diff
@@ -287,13 +288,13 @@ These examples require the following host permission:
 To work out the priority of a particular URL, look at the (developer-defined) `"priority"` key, the `"action"` key and the `"urlFilter"` key. These examples refer to the example rule file shown below them.
 
 **Navigation to https://google.com**
-Two rules cover this URL: the rules with IDs 1 and 4. The rule with ID 1 applies because `"block"` actions have a higher priority than `"redirect"` actions. The remaining rules do not apply because they are for longer URLs.
+: Two rules cover this URL: the rules with IDs 1 and 4. The rule with ID 1 applies because `"block"` actions have a higher priority than `"redirect"` actions. The remaining rules do not apply because they are for longer URLs.
```
sebastianbenz, line 291: There are now two subsequent colons in this paragraph

jpmedley: This is how that renders:
image

jpmedley:
@AmySteam I think @oliverdunk is done even though he's clicked the approval button. You should be able to copy edit soon.

site/en/docs/extensions/reference/declarativeNetRequest/index.md
sebastianbenz: an nice, all good then

site/en/docs/extensions/reference/declarativeNetRequest/index.md
```diff
@@ -158,102 +157,144 @@ examples of URL filters:
...
 
-{% Label %}rules.json:{% endLabel %}
+### Rule examples {: #rule-examples }
+The examples below illustrate how Chrome prioritizes rules in an extension. You may want to open the [prioritization](#implementation-matching-algorithm) rules in a separate window when reviewing them.
+
+#### The "priority" key {: #priority-key }
+
+These examples require the following host permission:
+
+`*://*.example.com/*`
```
AmySteam, line 286: > `*://*.example.com/*`

Is this the only permission required for the example below? It might be nice to have a full manifest code sample for ease of copy/paste

site/en/docs/extensions/reference/declarativeNetRequest/index.md
```diff
@@ -158,102 +157,144 @@ examples of URL filters:
...
+
+To work out the priority of a particular URL, look at the (developer-defined) `"priority"` key, the `"action"` key and the `"urlFilter"` key. These examples refer to the example rule file shown below them.
+
+Navigation to https://google.com
+: Two rules cover this URL: the rules with IDs 1 and 4. The rule with ID 1 applies because `"block"` actions have a higher priority than `"redirect"` actions. The remaining rules do not apply because they are for longer URLs.
+
+Navigation to https://google.com/1234
+: Because of the longer URL, the rule with ID 2 now matches in addition to  the rules with IDs 1 and 4. The rule with ID 2 applies because `"allow"` has a higher priority than `"block"` and `"redirect"`.
```
AmySteam, line 294:
```suggestion
: Because of the longer URL, the rule with ID 2 now matches in addition to the rules with IDs 1 and 4. The rule with ID 2 applies because `"allow"` has a higher priority than `"block"` and `"redirect"`.
```

site/en/docs/extensions/reference/declarativeNetRequest/index.md
```diff
@@ -281,140 +322,90 @@ When rules are applied to browsers with pages in the service worker's cached sto
...
-      "urlFilter" : "script.js",
-      "resourceTypes" : ["script"]
+  "condition": {
+    "urlFilter": "||example.com",
+    "resourceTypes": ["main_frame"]
+  }
+}
+```
+
+The following example uses regular expressions to redirect from `https://www.abc.xyz.com/path` to `https://abc.xyz.com/path`. In the `"regexFilter"` key notice how periods are escaped and that the capturing group selects either "abc" or "def". The `"regexSubstitution"` key specifies the first returned match of the regular expression using "\\1". In this case "abc" is captured from the redirected URL and placed in the substitution.
```
AmySteam, line 367:
```suggestion
The following example uses regular expressions to redirect from `https://www.abc.xyz.com/path` to `https://abc.xyz.com/path`. In the `"regexFilter"` key, notice how periods are escaped and that the capturing group selects either "abc" or "def". The `"regexSubstitution"` key specifies the first returned match of the regular expression using "\\1". In this case, "abc" is captured from the redirected URL and placed in the substitution.
```

site/en/docs/extensions/reference/declarativeNetRequest/index.md
```diff
@@ -281,140 +322,90 @@ When rules are applied to browsers with pages in the service worker's cached sto
...
+    "resourceTypes": [
+      "main_frame"
+    ]
+  }
+}
+```
+
+#### Headers {: #headers }
+
+The following example removes all cookies from both a main frame and any sub frames.
```
AmySteam, line 390:
```suggestion
The following example removes all cookies from a main frame and sub frames.
```

site/en/docs/extensions/reference/declarativeNetRequest/index.md
```diff
@@ -281,140 +322,90 @@ When rules are applied to browsers with pages in the service worker's cached sto
...
+]
+```
+
+#### Redirects {: #redirects }
+
+To use the example file shown below, your manifest file would need the following host permission:
+
+`*://*.example.com/*`
+
+The following example shows how to redirect a request from example.com to a page within the extension itself. The extension path `/a.jpg` resolves to `chrome-extension://EXTENSION_ID/a.jpg`, where `EXTENSION_ID` is the ID of your extension.
```
AmySteam, line 334: It might be a good idea to mention that the a.jpg also needs to be declared as a WAR. Just in case a developers skipped the WAR section.

site/en/docs/extensions/reference/declarativeNetRequest/index.md
```diff
@@ -158,102 +157,144 @@ examples of URL filters:
...
+
+```js
+function updateStaticRules(enableRulesetIds, disableCandidateIds) {
+  // Create the options structure for the call to updateEnabledRulesets()
+  let options = { enableRulesetIds: enableRulesetIds }
+  // Get the number of enabled static rules
+  const enabledStaticCount = await chrome.declarativeNetRequest.getEnabledRulesets();
+  // Compare rule counts to determine if anything needs to be disabled so that
+  // new rules can be enabled
+  const propsedCount = enableRulesetIds.length;
```
AmySteam, line 270:
```suggestion
  const proposedCount = enableRulesetIds.length;
```

AmySteam (changes requested):
Good job! So much easier to read :)

Left a few suggestions.

AmySteam:
⭐️ Just a quick note for a future iteration. I think developers would appreciate an explicit section that explains explicitly use cases for each permission and which ones may trigger a warning.

site/en/docs/extensions/reference/declarativeNetRequest/index.md
oliverdunk: I took a look at the updated summary and I think it would still be nice to work on. It says "both require host permissions" which is not true - "declarativeNetRequestWithHostAccess" doesn't require anything, but for it to be useful you have to separately request host permissions for the sites you're interested in.

site/en/docs/extensions/reference/declarativeNetRequest/index.md
```diff
@@ -158,102 +157,144 @@ examples of URL filters:
...
+as part of the "regexFilter" key.
+```
+
+### Interactions with service workers {: #interact-w-service-workers }
+
+declarativeNetRequest only applies to requests that reach the network stack. This includes responses from the HTTP cache, but may not include responses that go through a service worker's `onfetch` handler. declarativeNetRequest won't affect responses generated by the service worker or retrieved from `CacheStorage`, but it will affect calls to `fetch()` made in a service worker.
+
+### Web accessible resources {: #implementation-web-accessible-resources }
+
+A declarativeNetRequest rule cannot redirect from a public resource request to a resource that is not web accessible. Doing so triggers an error. This is true even if the specified web accessible resource is owned by the redirecting extension. To declare resources for declarativeNetRequests, use the manifest's [`"web_accessible_resources"`](/docs/extensions/mv3/manifest/web_accessible_resources/) array.
```
oliverdunk, line 235:
```suggestion
A declarativeNetRequest rule cannot redirect from a public resource request to a resource that is not web accessible. Doing so triggers an error. This is true even if the specified web accessible resource is owned by the redirecting extension. To declare resources for declarativeNetRequest, use the manifest's [`"web_accessible_resources"`](/docs/extensions/mv3/manifest/web_accessible_resources/) array.
```

oliverdunk:
> @oliverdunk and @sebastianbenz Are you done with this? I'd like to get a copy edit from Amy now.

Left a couple of comments but otherwise I'm done.

site/en/docs/extensions/reference/declarativeNetRequest/index.md
jpmedley: What's your reasoning? Mine sounds more specific to my ears.

site/en/docs/extensions/reference/declarativeNetRequest/index.md
```diff
@@ -158,102 +157,144 @@ examples of URL filters:
...
-- The number of dynamic and session-scoped rules that an an extension can add is bounded by the [`MAX_NUMBER_OF_DYNAMIC_AND_SESSION_RULES`][10] constant.
+Rules are triggered by requests sent from web pages. If multiple rules match a particular request, then the rules must be prioritized. This section explains how they're prioritized. Prioritization happens in two stages.
 
-## Updating enabled rulesets {: #update-enabled-rulesets }
+1. Priority is determined for rules within an extension.
+1. If more than one extension can apply a rule to a request, priority is determined for all extensions that match a particular request.
 
-An extension can update the set of enabled static rulesets using the [`updateEnabledRulesets()`][11]
-method.
+Thinking of matching this way: whatever rule a particular extension prioritizes will then be prioritized against rules from other extensions.
```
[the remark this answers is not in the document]
jpmedley: The result of the first part is a single rule. The word 'the' is not needed to make this grammatically correct.

site/en/docs/extensions/reference/declarativeNetRequest/index.md
jpmedley: Done.

jpmedley:
@AmySteam PTAL

site/en/docs/extensions/reference/declarativeNetRequest/index.md
```diff
@@ -158,102 +157,142 @@ examples of URL filters:
...
 }
 ```
 
-{% Label %}rules.json:{% endLabel %}
+### Rule examples {: #rule-examples }
+The examples below illustrate how Chrome prioritizes rules in an extension. When reviewing them, you may want to open the [prioritization](#implementation-matching-algorithm) rules in a separate window.
+
+#### The "priority" key {: #priority-key }
+
+These examples require [host permission](/docs/extensions/mv3/declare_permissions/) to `*://*.example.com/*`.
```
AmySteam, line 284: > These examples require [host permission](/docs/extensions/mv3/declare_permissions/) to `*://*.example.com/*`.

@jpmedley Are we sure the rules below don't need to declare "https://*.google.com/*" as a host permission.

AmySteam (approved):
@jpmedley I'm approving now, but can you double check [the host permissions for the priority rules sample](https://github.com/GoogleChrome/developer.chrome.com/pull/6921/files#r1284695314)?

site/en/docs/extensions/reference/declarativeNetRequest/index.md
jpmedley: They do.

jpmedley, line 284:
```suggestion
These examples require [host permission](/docs/extensions/mv3/declare_permissions/) to `https://.google.com/`.
```

PR #6933, opened 2023-07-26
Update sw intro to better reflect recent changes.

jpmedley:


site/en/docs/extensions/mv3/service_workers/index.md
```diff
@@ -3,14 +3,15 @@ layout: 'layouts/doc-post.njk'
...
+Extension service workers have some importatn differences with web service worker5s. Fist, they're are more than network proxies (as web service workers are often described). In addition to the [standard service worker events](https://developer.mozilla.org/docs/Web/API/ServiceWorkerGlobalScope#events), they also respond to extension events such as navigating to a new page, clicking a notification, or closing a tab. In some cases, these APIs can extend the service worker's lifetime or even start a service worker in response to specific events. `chrome.tabs.onCreated` does this, for example. In fewer cases, an API can keep a service worker alive indefinitely. An example of this is communicating with a nativew host. Finally, extension service wokers are registered and updated differently from web service workers.
```
sebastianbenz, line 14:
```suggestion
Extension service workers have some important differences with web service workers. First, they're are more than network proxies (as web service workers are often described). In addition to the [standard service worker events](https://developer.mozilla.org/docs/Web/API/ServiceWorkerGlobalScope#events), they also respond to extension events such as navigating to a new page, clicking a notification, or closing a tab. In some cases, these APIs can extend the service worker's lifetime or even start a service worker in response to specific events. `chrome.tabs.onCreated` does this, for example. In fewer cases, an API can keep a service worker alive indefinitely. An example of this is communicating with a native host. Finally, extension service workers are registered and updated differently from web service workers.
```

"or even start a service worker in response to specific events" - but that's the case for every service worker isn't it?

"an API can keep a service worker alive indefinitely" - not sure if indefinitely is the right word here. Maybe: "Some APIs will keep the service worker alive as long as they're being used".

oliverdunk, line 14: > An example of this is communicating with a nativew host

I think we should probably elaborate here and link to the Native Messaging docs?

jpmedley: I added a link, but I don't want to elaborate because this is an introduction.

jpmedley:
PTAL

site/en/docs/extensions/mv3/service_workers/index.md
```diff
@@ -3,14 +3,15 @@ layout: 'layouts/doc-post.njk'
...
+Extension service workers have some important differences with web service workers. First, they're are more than network proxies (as web service workers are often described). In addition to the [standard service worker events](https://developer.mozilla.org/docs/Web/API/ServiceWorkerGlobalScope#events), they also respond to extension events such as navigating to a new page, clicking a notification, or closing a tab. In some cases, these APIs can extend the service worker's lifetime or even start a service worker in response to specific events. `chrome.tabs.onCreated` does this, for example. A few APIs can keep a service worker alive indefinitely. An example of this is communicating with a [native host](/docs/extensions/mv3/nativeMessaging/). Finally, extension service workers are registered and updated differently from web service workers.
```
AmySteam, line 14:
```suggestion
Extension service workers have some important differences with web service workers. First, they are more than network proxies (as web service workers are often described). In addition to the [standard service worker events](https://developer.mozilla.org/docs/Web/API/ServiceWorkerGlobalScope#events), they also respond to extension events such as navigating to a new page, clicking a notification, or closing a tab. In some cases, these APIs can extend the service worker's lifetime or even start a service worker in response to specific events. `chrome.tabs.onCreated` does this, for example. A few APIs can keep a service worker alive indefinitely. An example of this is communicating with a [native host](/docs/extensions/mv3/nativeMessaging/). Finally, extension service workers are registered and updated differently from web service workers.
```

oliverdunk, line 14: Hmm, I don't think "In some cases, these APIs can [...] even start a service worker in response to specific events." makes sense. Don't all extension events wake up the service worker?

jpmedley: I should have said "some APIs can extend..."

oliverdunk: Would you be ok just removing that line? I'm actually not sure it's helpful here given we haven't introduced the idea of a lifetime yet. I feel like trying to include it means we then have to try to get really precise about the language in what is supposed to be an intro 😅

sebastianbenz: +1 to Oliver

PR #6953, opened 2023-07-28
Update Page Lifecycle API diagram to add some missing states

### LINKED ISSUE #5062: Page Lifecycle API guide does not mention 'frozen' as possible previous state for 'active', 'passive' or 'hidden'
**Describe the bug**

Both the [Page Lifecycle states](https://wd.imgix.net/image/eqprBhZUGfb8WYnumQ9ljAxRrA72/1mN2saGWo57O6vIOdcGy.svg) and the table in chapter [States](https://developer.chrome.com/blog/page-lifecycle-api/#states) indicate that the 'frozen' state has the following possible next states:

* active (via the resume event, then the pageshow event)
* passive (via the resume event, then the pageshow event)
* hidden (via the resume event)
* discarded (no events fired)

While 'frozen' is listed as possible previous state for 'discarded', it's not mentioned for 'active', 'passive' or 'hidden'.

**Expected behavior**

If state A is listed under possible next state under B, I expect B to be listed as possible previous state under A.

### LINKED ISSUE #5123: The Page Lifecycle diagram is missing page load events
URL: https://developer.chrome.com/blog/page-lifecycle-api/#overview-of-page-lifecycle-states-and-events

The diagram does not contain the `DOMContentLoaded`, `load`, and `pageshow` events that fire during page load. This makes the diagram confusing because it’s incomplete. It’s especially confusing since the `pageshow` event that fires on unfreezing *is* included, so it makes it look like that’s the only time that event fires.

tunetheweb:
Fixes #5123
Fixes #5062

- Update diagram
- Add missing "previous" states

Section of previous diagram:

image

Same section on new diagram:

image

FYI @fergald

### DIFF
--- a/site/en/blog/memory-and-energy-saver-mode/index.md
+++ b/site/en/blog/memory-and-energy-saver-mode/index.md
@@ -51,9 +51,9 @@ These are the worst times to store state because these events are completely unr
 You can refer to the [Page Lifecycle event diagram](/blog/page-lifecycle-api/#overview-of-page-lifecycle-states-and-events) to see what events are expected to fire as a page is being discarded. As you can see from that diagram, a tab can get from the "hidden" state to the "discarded" state without any events firing.
 
 <figure>
-  <a href="https://wd.imgix.net/image/eqprBhZUGfb8WYnumQ9ljAxRrA72/KCIeOsJ0lCWMthBSSBrn.svg">
+  <a href="https://wd.imgix.net/image/W3z1f5ZkBJSgL1V1IfloTIctbIF3/gtCHLOWSAwYAswQY9Gqv.svg">
     {% Img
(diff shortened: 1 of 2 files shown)


philipwalton:
Nit: my preference would be to keep `DOMContentLoaded`, `load`, and `pageshow` all on the same horizontal line.

I know the diagram doesn't fully do this now, but the original plan with the diagram was to have the states along the vertical axis and phases (loading => unloading) along the horizontal axis. Our inability to change the order of `pagehide` and `visibilitychange` during unload breaks that paradigm, but I still do think it's a bit easier to read if it's fully linear in both directions.

tunetheweb:
Required a bit of squishing, but done!

image

philipwalton (approved):

malchata (approved):

PR #7008, opened 2023-08-07
Clarify when execution of DNR rules is undefined behaviour

oliverdunk:
I missed this when reviewing our recent update - the current wording is too strong because it implies there's no ordering whatsoever, when there is ordering between, say, allow and redirect rules (we already have this documented a few lines below!). The key thing to cover is:

- There's no guarantee if two rules of the same action and priority will execute first
  -  This is the case even if one of them came from a static ruleset and another from a session/dynamic one

### DIFF
--- a/site/en/docs/extensions/reference/declarativeNetRequest/index.md
+++ b/site/en/docs/extensions/reference/declarativeNetRequest/index.md
@@ -167,7 +167,7 @@ Rules are triggered by requests sent from web pages. If multiple rules match a p
 Thinking of matching this way: whatever rule a particular extension prioritizes will then be prioritized against rules from other extensions.
 
 {% Aside 'caution' %}
-Avoid building rulesets that depend on rules with the same priority running in a particular order. Rules with the same priority in the same set (e.g., static rules with priority 1) are executed in an arbitrary order which is not defined and can change between runs or browser versions. This is the case across browsers.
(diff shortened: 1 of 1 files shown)


site/en/docs/extensions/reference/declarativeNetRequest/index.md
```diff
@@ -167,7 +167,7 @@ Rules are triggered by requests sent from web pages. If multiple rules match a p
 Thinking of matching this way: whatever rule a particular extension prioritizes will then be prioritized against rules from other extensions.
 
 {% Aside 'caution' %}
-Avoid building rulesets that depend on rules with the same priority running in a particular order. Rules with the same priority in the same set (e.g., static rules with priority 1) are executed in an arbitrary order which is not defined and can change between runs or browser versions. This is the case across browsers.
+Avoid writing rules with the same action and priority that must run in a particular order. Rules with the same action and priority (e.g., redirect rules with priority 1) are executed in an arbitrary order which is not defined and can change between runs or browser versions, even when spread between multiple types of ruleset (e.g, a static rule and a session rule). This is the case across browsers.
```
jpmedley, line 170: >(e.g, a static rule and a session rule)

The rest of this looks good, but it's not clear to me what the last part is adding. Tell *me* what it means in as many words as it takes. We'll loop back to the wording of this afterwards.

jpmedley (changes requested):

site/en/docs/extensions/reference/declarativeNetRequest/index.md
oliverdunk: For sure! If you have two redirect rules with the same priority, with one added in the manifest and the other as a dynamic rule, then there are no guarantees about where you'll be redirected to. We might prefer session rules over dynamic rules in this Chrome version but in the next prefer dynamic rules first and then session rules second.

jpmedley (approved):

PR #7017, opened 2023-08-08
Add dns API reference page

AmySteam:
Fixes #https://github.com/GoogleChromeLabs/Extension-Docs/issues/81

- Add missing Dev Channel API
-
-

### DIFF
--- a/site/en/docs/extensions/reference/dns/index.md
+++ b/site/en/docs/extensions/reference/dns/index.md
@@ -1,0 +1,20 @@
+---
+api: dns
+---
+
+## Manifest
+
+You must declare the `"dns"` permission in the [manifest][manifest] to use this API.
+
+```json
+{
+  "name": "My extension",
+  ...
+  "permissions": [
+    "dns"
+  ],
+  ...
+}
+```
+
+[manifest]: /docs/extensions/mv3/manifest/


patrickkettner:
It would be great to have some information on what the API actually is, or its use. the [MDN](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/dns) page has some info about it.

AmySteam:
@patrickkettner Good point 👍 . I added a code sample. LMKWYT

site/en/docs/extensions/reference/dns/index.md
```diff
@@ -0,0 +1,43 @@
...
+  ...
+  "permissions": [
+    "dns"
+  ],
+  ...
+}
+```
+
+{% Aside %}
+This API is only available in [Chrome Dev](https://www.google.com/chrome/dev/).
```
patrickkettner, line 21: I would explicitly mention that there are not plans for it to move out of dev channel with the normal chrome release train.

patrickkettner (commented):
looks great! only comment would be to elaborate on the first aside

AmySteam:
@jpmedley This is a short one, can you do a quick copy edit? Thank you! 🙏

patrickkettner (approved):
LGTM

site/en/docs/extensions/reference/dns/index.md
```diff
@@ -0,0 +1,43 @@
...
+}
+```
+
+{% Aside %}
+This API is only available in [Chrome Dev](https://www.google.com/chrome/dev/). There are no foreseeable plans to move this API from the dev channel into Chrome stable.
+{% endAside %}
+
+### Usage {: #usage }
+
+The following code uses the [`resolve()`](#method-resolve) method to retrieve the IP address of `example.com`.
```
jpmedley, line 26:
```suggestion
The following code calls [`resolve()`](#method-resolve) to retrieve the IP address of `example.com`.
```

site/en/docs/extensions/reference/dns/index.md
```diff
@@ -0,0 +1,43 @@
...
+const resolveDNS = async () => {
+    let record = await chrome.dns.resolve('example.com');
+    console.log(record.address); // "192.0.2.172"
+};
+
+resolveDNS();
+```
+
+{% Aside 'important' %}
+Do not include the scheme or trailing slash in the hostname. For example, https://example.com/ is invalid.
```
jpmedley, line 40:
```suggestion
Do not include the scheme or trailing slash in the hostname. For example, `https://example.com/` is invalid.
```

jpmedley (changes requested):

jpmedley (approved):

PR #7027, opened 2023-08-09
remote-debugging: add adb workflow

paulirish:
This adb workflow was the 'old' way to do it and we removed it from the docs. However over the past couple years a bunch of folks have needed guidance on a direct CDP connection to Chrome on Android. 

Thus.. here's the workflow. (This information is almost entirely not googleable right now) :/


sofiayem:
@paulirish Thank you for drafting. I applied some edits to better comply with our style guide. Please fact check and make sure I didn't alter the meaning. Otherwise, LGTM.

sofiayem (approved):

paulirish:
> I applied some edits to better comply with our style guide. Please fact check and make sure I didn't alter the meaning. Otherwise, LGTM.

omg these edits are SO good. thank you! :)

PR #7042, opened 2023-08-10
Copy edit "Give users options".

jpmedley:
I was reading this to see if it covered everything in a FAQ question. This has some strange bits in it. If you know anything about it off the top of your head, please suggest it.


AmySteam (approved):
Left a few NITs. But overall LGTM 👍 

Eventually this page should be revisited and the screenshots updated.

There are other ways to allow users to set options. For example, setting options in another HTML page included in the extension.

PR #7049, opened 2023-08-11
Further updates to the Unload deprecation blog post

tunetheweb:

- Some tweaks to unload post
- Move Page Lifecycle API from /blog/ to /article/ and add redirect


site/en/blog/deprecating-unload/index.md
```diff
@@ -31,18 +31,28 @@ Scenarios where this event was most commonly used include:
...
+However the `unload` event [is extremely unreliable](/articles/page-lifecycle-api/#the-unload-event). In most browsers the code often won't run and it has a negative impact on a site's performance, by preventing the usage of [bfcache (back/forward cache)](https://web.dev/bfcache/#never-use-the-unload-event).
 
-This is a historical legacy and the `unload` handler [should not prevent use of the bfcache according to specification](https://github.com/fergald/docs/blob/master/explainers/permissions-policy-deprecate-unload.md#unload-as-specced). Chrome aims to move to conform more to specification (as Safari already does), and with the aim of being able to fully deprecate the legacy `unload` handler at some point in the future.
+This is a historical legacy and the `unload` handler [should not prevent use of the bfcache according to specification](https://github.com/fergald/docs/blob/master/explainers/permissions-policy-deprecate-unload.md#unload-as-specced). Chrome aims to move to conform more to specification (as Safari already does).
```
fergald, line 36: We don't aim to conform to the spec. This is actually the main point I wanted to correct. The spec says that we should ignore unload if BFCache is possible. We are not going to do that. If someone enabled it via PP and installed an unload handler then we will block BFCache just as we do today.

tunetheweb: Ah I thought your point was more we were aiming to beyond the spec (and fully deprecate unload) - which is covered in the next paragraph.

WDYT about this wording then, for this paragraph?:

```suggestion
This is a historical legacy and the `unload` handler [should not prevent use of the bfcache according to specification](https://github.com/fergald/docs/blob/master/explainers/permissions-policy-deprecate-unload.md#unload-as-specced). Chrome aims to move to conform more to specification by default (as Safari already does). Chrome will also provide options for sites to opt-out of this new default to allow them time to migrate away from using `unload` handlers.
```

fergald: That's not really it. We really don't aim to conform to the spec. Our aim is to get rid of `unload` so that that bit of spec is no longer relevant. For context, what happened was that WebKit favoured BFCache over running unload. We decided that on mobile this made sense since unload was already so unreliable. We then updated the spec to match our new mobile behaviour. This meant that WebKit was finally in line with spec. When it came to desktop, we realized that we couldn't follow the spec without breaking sites. We decided that if we are going to break sites, it's better to break them getting rid of unload entirely than to break them with an end state of "maybe unload runs maybe it doesn't, you probably can't predict it".

Actualy, I think the paragraph before could be changed too. So something like change in the first para

> On mobile browsers and on all types of Safari, the code often won't run ...

And then 2nd para is something like

> On desktop Chrome and Firefox, unload is quite reliable and starting to follow the specced behaviour would be disruptive by making it unreliable [maybe link to https://github.com/fergald/docs/blob/master/explainers/permissions-policy-deprecate-unload.md#unload-as-specced-is-a-footgun]. Chrome's aim is to eventually remove the `unload` event. Until then it will remain reliable on desktop for those who have explicitly opted-out of the deprecation.

tunetheweb: OK let's avoid any mention of "specced" behaviour then.

Have had another go in https://github.com/GoogleChrome/developer.chrome.com/pull/7049/commits/79a05d389ba6b51aaf733d197abed4fb1a216e4a so let me know if that adequately addresses your concerns.

site/en/blog/deprecating-unload/index.md
```diff
@@ -54,6 +54,8 @@ Instead of `unload` it is recommended to use:
...
 - [`pagehide`](/articles/page-lifecycle-api/#event-pagehide): To determine when the user has navigated away from the page. This event happens when the user navigates away from the page, reloads the page, or closes the browser window. The `pagehide` event is not fired when the page is simply minimized or switched to another tab. Note that, as `pagehide` does not make a page ineligible for the back/forward cache, it is possible a page can be restored after this event fires. If you're cleaning up any resources in this event, then they may have to be restored on page restore.
 
+The `beforeunload` event has a slightly different use case to `unload` in that it is a cancellable event. It is often use to warn users of unsaved data when navigating away. This event is also unrealiable as it will not fire if a background tab is killed. It is recommended to limit use of `beforeunload` and [only add this conditionally](/blog/page-lifecycle-api/#the-beforeunload-event), and instead use the above events for most `unload` replacements.
```
tomayac, line 57:
```suggestion
The [`beforeunload`](https://developer.mozilla.org/docs/Web/API/Window/beforeunload_event) event has a slightly different use case to `unload` in that it is a cancellable event. It is often used to warn users of unsaved changes when navigating away. This event is also unrealiable as it will not fire if a background tab is killed. It is recommended to limit use of `beforeunload` and [only add it conditionally](/blog/page-lifecycle-api/#the-beforeunload-event). Instead, use the above events for most `unload` replacements.
```

tomayac (approved):
LGTM with some suggestions. Thanks for adding this!

fergald (commented):
LGTM, thanks.

PR #7054, opened 2023-08-11
Move item from an FAQ

jpmedley:
Fixes #SOME_ISSUE_NUMBER

-
-
-

### DIFF
--- a/site/en/docs/extensions/reference/management/index.md
+++ b/site/en/docs/extensions/reference/management/index.md
@@ -22,6 +22,11 @@ API. For example:
 [`management.getPermissionWarningsByManifest`][2], [`management.uninstallSelf`][3], and
 [`management.getSelf`][4] do not require the management permission.
 
+## Notes
+
+The management API is intended to create new tab page replacement extensions. It is not
+intended to fire install and uninstall events for an extension.
+
 [1]: /docs/extensions/mv3/manifest
 [2]: #method-getPermissionWarningsByManifest
 [3]: #method-uninstallSelf


site/en/docs/extensions/reference/management/index.md
```diff
@@ -22,6 +22,11 @@ API. For example:
 [`management.getPermissionWarningsByManifest`][2], [`management.uninstallSelf`][3], and
 [`management.getSelf`][4] do not require the management permission.
 
+## Notes
+
+The management API is intended to create new tab page replacement extensions. It is not
```
AmySteam, line 27: So to create a new tab extension, you set that in the manifest https://developer.chrome.com/docs/extensions/mv3/override/

I'm not sure exactly why [this faq](https://developer.chrome.com/docs/extensions/mv3/faq/#faq-management) says that. @oliverdunk do you have any experience with the management API and new tab extensions?

jpmedley: This is almost word for word what the FAQ says.

jpmedley: For what it's worth.

AmySteam: Joe, you did the correct thing. I missed this when we were critically examining the facts... I mean faqs

jpmedley: LGTM, then?

AmySteam: I'd like a DRE recommendation because this might be old information and the note might cause more confusion.

oliverdunk: Thanks both for looking in to this - I'm not sure what led to that FAQ, and I couldn't find anything. If I had to guess I'd say that it may be useful for new tab page extensions that want to list all of the other extensions that are installed. That said, these days it's useful for far more than that - so I think it would be clearer without the note.

PR #7080, opened 2023-08-15
feat: DCC homepage refresh

mamieorine:

- Developed CSS for the homepage refresh
- Added new assets to the project
- Enhanced gulp script to be able to copy /static/dcc/* to the /dist folder.


matthiasrohmer (commented):
That doesn't look too bad. I'll follow up via e-mail too to share some more details.

image

- Cards stick to the edge of the screen

- Please also add the Open Sans font to both the assets as well as the CSS. While we wait for approval for Android Euclid use Poppins (https://fonts.google.com/specimen/Poppins, Bold 600/700 maybe) as a stand-in.

- hero.svg doesn't gracefully scale on mobile screens

- You can also use this PR to add a second header which is then only used for the homepage. Implement it in a Nunjucks partial that we can then include - we'll don't need the 'Subscribe' button, avatar and menu button from the Figma but everything else.

site/_static/static/dcc/hero.svg
matthiasrohmer, line 1: This shouldn't be an SVG as base64 encoding the image makes it considerably bigger than what the original image would be.

site/_static/static/dcc/open-icon.svg
matthiasrohmer, line 1: Same as with hero.svg - this is a raster image actually and should be exported as one, so that rather be a PNG or something.

site/_static/static/dcc/powerful-icon.svg
matthiasrohmer, line 1: Same as with hero.svg - this is a raster image actually and should be exported as one, so that rather be a PNG or something.

site/_static/static/dcc/solutions-icon.svg
matthiasrohmer, line 1: Same as with hero.svg - this is a raster image actually and should be exported as one, so that rather be a PNG or something.

matthiasrohmer (commented):
- Please remove the sidebar, it's redundant to the updated header
image

- Add articles to the updated header as it would be inaccessible when we remove the sidebar

- Please add `:hover` states to the header nav items

- The search button dangles in the void for some narrow viewports
image

- When search is opened on narrow viewports it weirdly breaks
image

- Click on the burger menu doesn't do anything for mobile

- Please check the horizontal rhythm in the header. I know it's not perfect in the Figma either, but the typo upper boundaries should match, respectively the Developers `s` should always be as big as the `G` of Get inspired.

Screenshot 2023-08-25 at 18 10 46

- There is a serif font flash before the web fonts are loaded

site/_includes/layouts/devsite.njk
```diff
@@ -16,7 +16,7 @@
 
     <link rel="preload" as="font" crossorigin href="/fonts/google-sans-v2003/regular/latin.woff2">
     <link rel="preload" as="font" crossorigin href="/fonts/google-sans-v2003/medium/latin.woff2">
-
+    <link href="https://fonts.googleapis.com/icon?family=Material+Icons|Material+Icons+Outlined" rel="stylesheet">
```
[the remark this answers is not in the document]
devnook: afaik they will be loaded from googleapis anyways on Devsite, so I'd keep it simple here. We should add preconnect though link rel="preconnect"

site/_includes/partials/devsite-header.njk
```diff
@@ -0,0 +1,41 @@
...
+      <a class="display-inline-flex" href="/" aria-label="developer.chrome.com">
+        {{ icon('logo', {id: 'chromeLogo', hidden: true}) }}
+        <div class="logo__text">
+          {{ icon('logo-lockup', {id: 'chromeDevelopersText', hidden: true}) }}
+        </div>
+      </a>
+    </div>
+
+    <nav class='dcc-top-menu'>
+      <a href='/' class='no-visited'>Get Inspired</a>
```
devnook, line 25: Please no untraslatable strings in template

site/_scss/layouts/devsite-theme.scss
```diff
@@ -1,9 +1,890 @@
...
+  }
+
+  .button {
+    cursor: pointer;
+    font-weight: #{$TEXT_WEIGHT};
+    height: auto;
+    letter-spacing: 0.5px;
+    line-height: 1.5rem;
+    margin-left: 0;
+    padding: 0.8125rem 2.5rem;
```
devnook, line 111: use --devsite-buttons-padding

site/_scss/layouts/devsite-theme.scss
```diff
@@ -1,9 +1,890 @@
...
+    }
+  }
+
+  .button {
+    cursor: pointer;
+    font-weight: #{$TEXT_WEIGHT};
+    height: auto;
+    letter-spacing: 0.5px;
+    line-height: 1.5rem;
+    margin-left: 0;
```
devnook, line 110: --devsite-buttons-margin

devnook (changes requested):
It seems you are targeting a lot of devsite- classes there, which is not allowed. See the instructions doc.

site/_includes/partials/devsite-header.njk
mamieorine: Hi Ewa, sorry we don't quite understand what this means. Where are the no untranslatable strings in the lines that you have requested change?

site/_includes/layouts/devsite.njk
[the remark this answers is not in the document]
devnook: Please address this comment.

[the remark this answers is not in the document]
mamieorine: Hi Ewa, @simgobb  has added `rel="preconnect"` already. 
```
<link rel="preconnect" href="https://fonts.googleapis.com">
<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
```

Please re-review, thank you 😊

site/_includes/partials/devsite-header.njk
```diff
@@ -0,0 +1,39 @@
...
+      locale="{{locale}}"
+      buttonLabel="{{ 'i18n.common.open_search' | i18n(locale) }}"
+      docsLabel="{{ 'i18n.common.documentation' | i18n(locale) }}"
+      articlesLabel="{{ 'i18n.common.articles' | i18n(locale) }}"
+      blogLabel="{{ 'i18n.common.blog' | i18n(locale) }}"
+      placeholder="{{'i18n.common.search_placeholder' | i18n(locale)}}"
+    ></search-box>
+
+    <nav class='dcc-top-menu'>
+      <a href='/' class='no-visited'>Get Inspired</a>
```
devnook, line 33: These strings will not get translated if you hardcode them. Please use i18n filter as in the example above.

site/_scss/layouts/devsite-theme.scss
```diff
@@ -1,9 +1,877 @@
...
+    padding: 0.46875rem;
+    position: relative;
+    top: 0.5rem;
+  }
+
+  .dcc-blog-card {
+    --devsite-buttons-margin: 0.5rem 0 0 -0.75rem;
+    --devsite-item-gap: 0;
+
+    padding: 5rem 2.125rem 3rem;
```
devnook, line 123: These paddings seem too big (e.g. bottom 68px vs 39 in Figma)

site/_scss/layouts/devsite-theme.scss
```diff
@@ -1,9 +1,877 @@
...
+
+    span {
+      position: absolute;
```
devnook, line 126: I assume you target a card label here. Please use the specific selector, rather than every span on the card (yes, this is this place where devsite- selector makes more sense than "span")

site/_scss/layouts/devsite-theme.scss
```diff
@@ -1,9 +1,877 @@
...
+      }
+    }
+
+    @media (max-width: 840px) {
+      padding-left: 1rem;
+
+      figure {
+        margin-bottom: 1.5rem;
+      }
+
```
devnook, line 159: Remove whitespace

site/_scss/layouts/devsite-theme.scss
```diff
@@ -1,9 +1,877 @@
...
+      }
+    }
+
+    h3 + div {
+      align-items: start;
+      flex-direction: column;
+      max-width: 12.5rem;
+    }
+
+    .button::before {
```
devnook, line 218: Please make it a universal class, that can be also used outside of services card (like .dcc-icon-button). Adapt dcc-web-app-features-link calls below as well.

site/_scss/layouts/devsite-theme.scss
```diff
@@ -1,9 +1,877 @@
...
+        margin-bottom: 0;
+        padding: 0;
+      }
+
+      img{
+        height: 12.5rem;
+        object-fit: cover;
+      }
+    }
+
```
devnook, line 540: Whitespace.

site/_scss/layouts/devsite-theme.scss
```diff
@@ -1,9 +1,877 @@
...
+      --devsite-h1-font: 700 1.5625rem #{$TEXT_PRIMARY_FONT_FAMILY};
+
+      .dcc-blog-card {
+        max-width: 100%;
+        margin-bottom: 2rem;
+      }
+    }
+  }
+
+
```
devnook, line 559: Whitespace.

site/_scss/layouts/devsite-theme.scss
```diff
@@ -1,9 +1,877 @@
...
+      }
+    }
+
+  footer {
+    margin-top: 0;
+  }
+}
+
+/* Layout, header, and footer */
+[theme=web-theme] {
```
devnook, line 668: Since these styles do not influence the look and feel of the content from devsite, please remove to another file.

devnook (changes requested):
Reviewed

site/_includes/partials/devsite-navigation-rail.njk
```diff
@@ -0,0 +1,34 @@
+{% from 'macros/icon.njk' import icon with context %}
+
```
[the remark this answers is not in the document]
mamieorine: There are 2 places different from the navigation-rail.njk file, the logo and the rail items `{% for name, item in nav.devsite_rail %}` (please see the [latest commit ](https://github.com/GoogleChrome/developer.chrome.com/pull/7080/commits/a04df0e7394fd2c3381eecb7b1456b98aa83a485)).

_**Please remove duplicated code and just replace what differs**_ - I am not sure what approach we should take, do you want us to use the old file and add the condition checking the new template?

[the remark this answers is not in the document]
devnook: Any solution that get you the desired outcome and does not involve repeating 134 lines of code.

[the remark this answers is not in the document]
mamieorine: Thank you for clarifying, I have fixed this, please see the latest commit -https://github.com/GoogleChrome/developer.chrome.com/pull/7080/commits/55105f1fa1e560a8c23798db4cadd309bdcf4256.

site/_includes/partials/devsite-header.njk
```diff
@@ -0,0 +1,39 @@
+{% from 'macros/icon.njk' import icon with context %}
```
matthiasrohmer, line 1: I think if we use the new header across all of the site we should also get rid of the sidebar everywhere and use the exact same header on every page. This would also fix the content jump when you navigate away from the homepage, as the other header is slightly smaller.

matthiasrohmer (commented):
Quite a few images are still SVGs, why they are actually raster images.

- site/_static/static/dcc/bringing-safety-thumb.svg
- site/_static/static/dcc/case-study-thumb.svg
- site/_static/static/dcc/chrome-111-thumb.svg
- site/_static/static/dcc/chrome-112-thumb.svg
- site/_static/static/dcc/chrome-dev-insider-thumb.svg

site/_includes/layouts/devsite.njk
```diff
@@ -105,29 +107,23 @@
...
       <section  class="dcc-devsite-wrapper" theme="web-theme">
         {% block devsite_main %}{% endblock %}
       </section>
 
       {% block footer %}
-        {% include 'partials/footer.njk' %}
+        {% include 'partials/devsite-footer.njk' %}
       {% endblock %}
     </div>
     {% include 'partials/cookie-banner.njk' %}
```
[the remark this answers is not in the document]
devnook: This is still happening:
Screenshot 2023-09-07 at 15 47 47

site/_includes/layouts/devsite.njk
```diff
@@ -132,6 +126,18 @@
...
 
-    {% block scripts %}{% endblock %}
+    {% block scripts %}
+    
+    <script>
+        const button = document.querySelector('.dcc-hero .button');
```
devnook, line 132: This script prevents from the anchor to get attached in the url bar.
Please use css rule scroll-behavior: smooth; instead and reove unnecessary js.

site/_scss/layouts/devsite-theme.scss
devnook: Was this addressed?

site/_scss/_global.scss
```diff
@@ -30,6 +30,8 @@
 
 :root {
   --flow-space: #{get-size(500)};
+  -moz-osx-font-smoothing: grayscale;
+  -webkit-font-smoothing: antialiased;
```
matthiasrohmer, line 34: Can we please rather set those on `body`? Using `:root` for actual properties and not only var declarations seems weird.

site/_scss/globals/_fonts.scss
```diff
@@ -108,3 +108,51 @@
...
+  font-size: 24px;
+  -webkit-font-smoothing: antialiased;
+  font-style: normal;
+  font-weight: normal;
+  letter-spacing: normal;
+  line-height: 1;
+  text-transform: none;
+  white-space: nowrap;
+  word-wrap: normal;
+}
```
matthiasrohmer, line 158: Good to have this added, but please move to devsite-theme instead so that they do not get loaded on pages that are not the homepage, where we do not use those icons yet. Please correct me if I am wrong and they are required for other pages as well.

site/_scss/layouts/devsite-theme.scss
mamieorine: Sorry, I did not explain why we did it that way. 

All `dcc-blog-card` cards have the label on top so I set the position of the label to **_absolute_** because the label is wrapped in the description element and the description in some cards is displayed below the image. I cannot think of other ways to move this label up by using CSS except to set the position to absolute. To avoid overlapping between the label and heading, I need to add extra padding-top into it. The layout of the blog card looks good to me, please see the screenshot below:

Screenshot 2023-09-07 at 15 16 53

mamieorine: In my opinion, the better way is to adjust the HTML element by moving the label out of the description element and place on top of the card. As you see in the screenshot below, the label is in `devsite-landing-row-item-description` element. 
Screenshot 2023-08-24 at 15 35 33.

Please let me know if there is another better way. I will fix this ASAP. Thank you!

site/_scss/globals/_fonts.scss
mamieorine: I moved this to the `devsite-theme.scss` already.

site/_scss/_global.scss
mamieorine: I have fixed this in [this commit](https://github.com/GoogleChrome/developer.chrome.com/pull/7080/commits/f943fb03a2d1af33ef021e0f8fb58d14606fce77).

site/_scss/layouts/devsite-theme.scss
matthiasrohmer: As altering the HTML is not an option I agree that this is the least invasive solution without heavily altering the surrounding layout logic, thanks for explaining.

site/_scss/layouts/devsite-theme.scss
matthiasrohmer: @mamieorine, please target `.devsite-landing-row-item-labels` (the parent element of the span you are currently targeting) instead to avoid accidentally affecting any spans that might appear in that element in the future.

site/_scss/layouts/devsite-theme.scss
matthiasrohmer: @simgobb, @mamieorine: that's a double new-line which can be condensed to one.

site/_includes/layouts/devsite.njk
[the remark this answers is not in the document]
mamieorine: I have fixed this. 
Screenshot 2023-09-07 at 15 43 28

site/_includes/partials/devsite-side-nav.njk
```diff
@@ -0,0 +1,13 @@
+<side-nav type="{{ 'project' if project_key else 'site' }}">
+  <div class="side-nav__container lg:display-flex">
+    {% include 'partials/devsite-navigation-rail.njk' %}
+    {#
+      If there's a toc in the global data scope then it implies that we're
+      on an individual doc page and need to display the project's ToC on
+      the left-hand side.
+    #}
+    {% if project_key %}
```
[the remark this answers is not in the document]
mamieorine: This file has been removed. I replaced the current header with the new (devsite) header so now the new header and the new side nav are applied globally.

site/_includes/partials/devsite-footer.njk
```diff
@@ -0,0 +1,87 @@
+{% from 'macros/icon.njk' import icon with context %}
+
+<footer class="type--footer gap-top-1000 lg:pad-left-600 lg:pad-right-600" data-side-nav-inert data-search-inert>
```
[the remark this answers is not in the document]
mamieorine: This issue has been fixed. I added this condition `{% if not isDevsite %}` to check if the page is the homepage, if so just remove the social row.

site/_scss/layouts/devsite-theme.scss
mamieorine: @matthiasrohmer  I think you might reviewed the old file, we updated this selector a few days back. Please see the latest file [here](https://github.com/GoogleChrome/developer.chrome.com/blob/4956d82e2973630b36fd268f5e720413ce6baa7c/site/_scss/layouts/devsite-theme.scss. ). Thank you 😊

site/_scss/layouts/devsite-theme.scss
mamieorine: @matthiasrohmer  I think you might reviewed the old file, we added `.dcc-icon-button` for each link and added style to this selector a few days back. Please see the [latest file here](https://github.com/GoogleChrome/developer.chrome.com/blob/4956d82e2973630b36fd268f5e720413ce6baa7c/site/_scss/layouts/devsite-theme.scss. ). Thank you 😊

site/_scss/layouts/devsite-theme.scss
mamieorine: We removed this double new line. Please see the [latest file here](https://github.com/GoogleChrome/developer.chrome.com/blob/4956d82e2973630b36fd268f5e720413ce6baa7c/site/_scss/layouts/devsite-theme.scss.). Thank you 😊

site/_scss/layouts/devsite-theme.scss
mamieorine: Removed

site/_scss/layouts/devsite-theme.scss
mamieorine: We moved all code in `/* Layout, header, and footer */ ` out, and replaced the current layout with the devsite layout following Matthias' feedback so the whole site now has the new header and the new side nav for mobile.

site/_scss/layouts/devsite-theme.scss
mamieorine: removed 😊

matthiasrohmer:
Merged in https://github.com/GoogleChrome/developer.chrome.com/pull/7302.

PR #7089, opened 2023-08-16
Speculation rules: match any hash fragment.

jeremyroman:
This expands the scope slightly, but also we're contemplating a change which would make this unnecessary -- but while developer.chrome.com is using this syntax, it's more difficult to tell at a glance how common other uses are.

### DIFF
--- a/site/_includes/layouts/base.njk
+++ b/site/_includes/layouts/base.njk
@@ -52,11 +52,11 @@
           "where": {
             "and": [
               {
-                "href_matches": "/*\\?*", "relative_to": "document"
+                "href_matches": "/*\\?*#*", "relative_to": "document"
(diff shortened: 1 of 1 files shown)


jeremyroman:
@tunetheweb What do you think? See also GoogleChrome/web.dev#10340. Currently, /*\\?* still doesn't match URLs which have a non-empty fragment part, but we're contemplating making it do so. However, as part of evaluating that I noticed that these two sites are major existing cases where the behavior varies (albeit probably in a benign or even positive way). Moving these to the behavior we're considering both allows us to see whether that would be undesirable, and better measure any _other_ existing uses which might be of concern.

tunetheweb (approved):

PR #7138, opened 2023-08-21
Update an image

jpmedley:
@AmySteam I hesitate to change the 'updated' date since the content itself has not been reviewed.


AmySteam (approved):
LGTM 👍 @jpmedley I agree, let's not update the date until we review the content.

PR #7142, opened 2023-08-22
Added clarification to what "activeTab" allows.

materialknight:
Clarified that the "scripting" permission is also needed (in addition to the "activeTab" permission) in order to call "scripting.insertCSS" or "scripting.executeScript".


AmySteam (approved):
Thanks @materialknight. This change makes it very explicit 💯 .

PR #7146, opened 2023-08-22
Clarify use of "author" key.

jpmedley:


AmySteam:
@jpmedley Interesting... do you know when this became a requirement for the Chrome web store? (I ask because this hasn't been a requirement in the past).

If this is the case, I think we should add a comment on the bottom keys that they are optional, because right now it implies that all keys under author are required for the CWS and optional for locally loaded extensions
Screenshot 2023-08-23 at 08 10 21

Also, we should update the information in the key page as well
https://pr-7146-static-dot-dcc-staging.uc.r.appspot.com/docs/extensions/mv3/manifest/author/

site/en/docs/extensions/mv3/manifest/index.md
```diff
@@ -27,7 +27,7 @@ The following code shows the supported manifest keys. For additional usage infor
   <span class="token property">"<a href="/docs/extensions/mv3/manifest/description">description</a>"</span><span class="token operator">:</span> <span class="token string">"A plain text description"</span><span class="token punctuation">,</span>
   <span class="token property">"<a href="/docs/extensions/mv3/manifest/icons">icons</a>"</span><span class="token operator">:</span> <span class="token punctuation">{</span>...<span class="token punctuation">}</span><span class="token punctuation">,</span>
 
-  <span class="token comment">// Optional</span>
+  <span class="token comment">// Required for Chrome Web Store. Optional for unpacked extensions.</span>
   <span class="token property">"<a href="/docs/extensions/mv3/manifest/author">author</a>"</span><span class="token operator">:</span> <span class="token string">"developer@example.com"</span><span class="token punctuation">,</span>
```
AmySteam, line 32: We should explicitly say that these keys are optional

jpmedley: Maybe I did this in the wrong place. Isn't there only an `"email"` key below this? It's `"email"` that makes it required.

I'm also trying to find out if it has to match the contact email on the dashboard.

AmySteam: Should this information go under the ["author" key](https://pr-7146-static-dot-dcc-staging.uc.r.appspot.com/docs/extensions/mv3/manifest/author/) page?

jpmedley:
There's something more complicated going on here. I'll open a new PR when I figure it out.

PR #7169, opened 2023-08-25
Make a heading more accurate.

jpmedley:
The whole section this title applies to has one link to information about themes and _nothing else_. It's misleading to include that in its title.


AmySteam (approved):
LGTM 👍

PR #7176, opened 2023-08-25
Add a more reasonable link for colors.

jpmedley:
(We should define these in the documentation instead of sending people to source code. I can't prioritize the time right now.)


AmySteam (approved):
👍

PR #7183, opened 2023-08-27
typo in match pattern 

letsbuildanything:
corrected 'https' to 'http'  in valid patterns example of 'http://*/*'


AmySteam (approved):
Thanks @letsbuildanything! 😄

PR #7215, opened 2023-08-31
Updated Match Patterns

wizardofdocs:
Condensed descriptions and examples. Converted example table to definition list. Integrated the invalid list with the descriptions and examples, to not confuse the reader with a list of things to not do.

Please make sure the special cases and examples are up to date.

Note on the invalid list: It's standard technical writing practice to state instructions positively whenever possible. Readers skimming a document are likely to confuse a list of don'ts with a list of dos and make more mistakes than if the don'ts weren't listed.

### DIFF
--- a/site/en/docs/extensions/mv3/match_patterns/index.md
+++ b/site/en/docs/extensions/mv3/match_patterns/index.md
@@ -19,14 +19,6 @@ A match pattern is a URL with the following structure, used to specify a group o
 - `https`
 - A wildcard `*`, which matches only `http` or `https`
 - `file`
-- `ftp`
-- `chromeui`
-- `extension`
-- `filesystem`
-- `ws`
-- `wss`
-- `data`
-- `uuid_in_package`
 
 For information on injecting content scripts into unsupported schemes, such as `about:` and `data:`, see [Injecting in related frames][cs-frames].
 


site/en/docs/extensions/mv3/match_patterns/index.md
```diff
@@ -3,154 +3,66 @@ layout: 'layouts/doc-post.njk'
...
 
 ```text
 &lt;scheme&gt;://&lt;host&gt;/&lt;path&gt;
 ```
 
-Each part can use wildcards `*`. Below is a detailed description:
-
-- **scheme**: Must include a valid scheme: `'http'`, `'https'`, `'file'`, `'ftp'`, or `'urn'`. The wildcard `*` only matches `http` or `https`.
-
+- **scheme**: Must be one of `'http'`, `'https'`, `'file'`, `'ftp'`, or `'urn'`. The wildcard `*` only matches `http` or `https`.
```
jpmedley, line 16:
```suggestion
**scheme**: Must be one of `'http'`, `'https'`, `'file'`, `'ftp'`, or `'urn'`. The wildcard `*` only matches `http` or `https`.
```

jpmedley: The run-in heading does an adequate job of making this scannable, so that nested bullets aren't needed.

site/en/docs/extensions/mv3/match_patterns/index.md
```diff
@@ -3,154 +3,66 @@ layout: 'layouts/doc-post.njk'
...
+  - For information on injecting content scripts into unsupported schemes, such as `about:` and `data:`, see [Injecting in related frames][cs-frames].
+  - The scheme must be separated from the rest of the pattern using a double slash (`//`).
 - **host**: A hostname (`www.example.com`), a `*` before the hostname to match subdomains (`*.example.com`), or just a wildcard `*`.
```
jpmedley, line 19:
```suggestion
**host**: A hostname (`www.example.com`). A `*` before the hostname to match subdomains (`*.example.com`), or just a wildcard `*`.
```

site/en/docs/extensions/mv3/match_patterns/index.md
```diff
@@ -3,154 +3,66 @@ layout: 'layouts/doc-post.njk'
...
+  - If you use a wildcard in the host pattern, it must be the first or only character, and it must be followed by a period (`.`) or forward slash (`/`).
+- **path**: Must contain at least a forward slash. The slash by itself is treated as a wildcard `/*`. Other examples include the following:
```
jpmedley, line 21: @AmySteam do you know what the `/*` is trying to tell me?

jpmedley, line 21:
```suggestion
**path**: Must contain at least a forward slash. The slash by itself is treated as a wildcard `/*`. For example:
```

site/en/docs/extensions/mv3/match_patterns/index.md
```diff
@@ -3,154 +3,66 @@ layout: 'layouts/doc-post.njk'
...
+  * `/foo*`
+  * `/foo/bar`.
 
-- **path**: Must start with `/` and be present. If alone, it is treated as `/*`. For example: `/*`, `/foo*`, or `/foo/bar`. Each '`*`' matches 0 or more characters.
+Extensions use match patterns for a variety of purposes, including declaring [host permissions][host-permissions] and other permissions for Chrome APIs or [content script][content-scripts] injection, as well as [`"web_accessible_resources.matches"`][war] and [`"externally_connectable.matches"`][ext-connect].
```
jpmedley, line 25: This seems like it should be a bulleted list. 

@oliverdunk @AmySteam Are we missing any places where match patterns can be used?

site/en/docs/extensions/mv3/match_patterns/index.md
```diff
@@ -3,154 +3,66 @@ layout: 'layouts/doc-post.njk'
...
 
 ## Special cases {: #special }
 
 `<all_urls>`
```
jpmedley, line 29:
```suggestion
`"<all_urls>"`
```

jpmedley: Quotes are part of the syntax of this items.

site/en/docs/extensions/mv3/match_patterns/index.md
```diff
@@ -3,154 +3,66 @@ layout: 'layouts/doc-post.njk'
...
-: It matches any URL that starts with a permitted scheme, including any pattern listed under [valid patterns](#valid-examples). This is a broad host permission, meaning the Chrome web store [review may take longer](/docs/webstore/review-process/#review-time-factors).
+: Matches any URL that starts with a permitted scheme, including any pattern listed under [valid patterns](#valid-examples). Because it affects all hosts, Chrome web store reviews for extensions that use it [may take longer](/docs/webstore/review-process/#review-time-factors).
 
 `file:///`
```
jpmedley, line 32:
```suggestion
`"file:///"`
```

jpmedley: @oliverdunk @AmySteam This looks to me like it should only have two slashes.

site/en/docs/extensions/mv3/match_patterns/index.md
```diff
@@ -3,154 +3,66 @@ layout: 'layouts/doc-post.njk'
...
-: Allows your extension to run on local files. It has no host and requires the user to manually [grant access][permissions-allow-access].
+: Allows your extension to run on local files. This pattern has no host and requires the user to manually [grant access][permissions-allow-access].
```
jpmedley, line 33:
```suggestion
: Allows your extension to run on local files. This pattern requires the user to manually [grant access][permissions-allow-access].
```

jpmedley: File URLs never have a host.

jpmedley (changes requested):

site/en/docs/extensions/mv3/match_patterns/index.md
AmySteam: I've always used three slashes on local files in extensions. This works in my samples.
https://en.wikipedia.org/wiki/File_URI_scheme

AmySteam:
The formatting needs to be tweaked
Screenshot 2023-08-31 at 16 18 32

wizardofdocs:
> The formatting needs to be tweaked Screenshot 2023-08-31 at 16 18 32

Done. Thanks for catching that.

site/en/docs/extensions/mv3/match_patterns/index.md
oliverdunk: I agree it's slightly weird, but I think three is right. The first two are the protocol (similar to http://) and the last one is for the slash at the start of a file path.

site/en/docs/extensions/mv3/match_patterns/index.md
```diff
@@ -3,154 +3,69 @@ layout: 'layouts/doc-post.njk'
...
 ## Special cases {: #special }
 
-`<all_urls>`
-: It matches any URL that starts with a permitted scheme, including any pattern listed under [valid patterns](#valid-examples). This is a broad host permission, meaning the Chrome web store [review may take longer](/docs/webstore/review-process/#review-time-factors).
+`"<all_urls>"`
+: Matches any URL that starts with a permitted scheme, including any pattern listed under [valid patterns](#valid-examples). Because it affects all hosts, Chrome web store reviews for extensions that use it [may take longer](/docs/webstore/review-process/#review-time-factors).
 
 `file:///`
-: Allows your extension to run on local files. It has no host and requires the user to manually [grant access][permissions-allow-access].
+: Allows your extension to run on local files. This pattern requires the user to manually [grant access][permissions-allow-access].
```
jpmedley, line 36:
```suggestion
: Allows your extension to run on local files. This pattern requires the user to manually [grant access][permissions-allow-access]. Note that this case requires three slashes, not two.
```

jpmedley: This link is broken. Please add this to the list at the bottom:

[permissions-allow-access]: /docs/extensions/mv3/declare_permissions/#allow_access

site/en/docs/extensions/mv3/match_patterns/index.md
jpmedley: @wizardofdocs I added some text to the definition to explain this.

jpmedley (changes requested):

site/en/docs/extensions/mv3/match_patterns/index.md
```diff
@@ -3,159 +3,74 @@ layout: 'layouts/doc-post.njk'
...
 ```
 
-Each part can use wildcards `*`. Below is a detailed description:
+**scheme**: Must be one of `'http'`, `'https'`, `'file'`, `'ftp'`, or `'urn'`. The wildcard `*` only matches `http` or `https`.
+  - For information on injecting content scripts into unsupported schemes, such as `about:` and `data:`, see [Injecting in related frames][cs-frames].
+  - The scheme must be separated from the rest of the pattern using a double slash (`//`).
 
-- **scheme**: Must include a valid scheme: `'http'`, `'https'`, `'file'`, `'ftp'`, or `'urn'`. The wildcard `*` only matches `http` or `https`.
+**host**: A hostname (`www.example.com`). A `*` before the hostname to match subdomains (`*.example.com`), or just a wildcard `*`.
+  - If you use a wildcard in the host pattern, it must be the first or only character, and it must be followed by a period (`.`) or forward slash (`/`).
```
oliverdunk, line 21: Is this right? Later in the same article we talk about things like `/foo*`.

site/en/docs/extensions/mv3/match_patterns/index.md
oliverdunk: I can't think of any others - thanks for checking!

oliverdunk (changes requested):

oliverdunk:
Thanks for the work @wizardofdocs! Really glad we are finding the time for some spring cleaning :)

site/en/docs/extensions/mv3/match_patterns/index.md
jpmedley: I don't know. @AmySteam?

site/en/docs/extensions/mv3/match_patterns/index.md
wizardofdocs: To clean up the bullet list structure: what do WAR and Externally Connectable do with the match patterns? It sounds to me like those aren't part of the declaring permissions part of the sentence.

jpmedley: The match patterns are for URLs. They effectively grant access permission to URLs.

wizardofdocs: Updated. How's this?

jpmedley: That looks good. Can you make it a bulleted list?

AmySteam: ## Original

So originally I separated the permissions part with the "Other ways that extensions use match patterns".

Screenshot 2023-09-08 at 16 28 28
 
## New approach

The new approach groups it all together as "permissions", which is not accurate.

Screenshot 2023-09-08 at 16 27 37

## Explanation

The match patterns under **web accessible resources** declare which websites can access specific resources that are bundled with the Chrome extension.

The match patterns under "externally_connectable" declare which websites can send and receive messages from the Chrome extension. 

These last two use cases IMO don't have anything to do with permissions.

## What problem are we solving?

I think the original version was clear enough without the need of a rewrite. I'm still confused about what problem we are solving with rewriting this section

wizardofdocs: That was my question.
I was asked to make that long sentence into a bulleted list to make it easier to read. To do that, I needed to understand  how all the use cases related to each other, and specifically whether those last two use cases had anything to do with permissions. I misinterpreted the previous answer to that question as "yes, these are all about permissions," which is how I ended up with that list.

I'd still like to understand what WAR and Externally Connectable are and how they use match patterns, but I don't need that information to fix the list.

AmySteam: **Match pattern use explanation**

- The match patterns under web accessible resources declare which URLs are allowed to access specific resources that are bundled with the Chrome extension.

- The match patterns under "externally_connectable" declare which URLS are allowed to send and receive messages from the Chrome extension.

site/en/docs/extensions/mv3/match_patterns/index.md
wizardofdocs: I think I've fixed it. Anything else I need to add?

site/en/docs/extensions/mv3/match_patterns/index.md
wizardofdocs: So is it useful to say, for example, "Granting access to [web-accessible resources]" and "Allowing message sending and receiving using [`"externally_connectable.matches"`]" in the list?

AmySteam: > So is it useful to say, for example, "Granting access to [web-accessible resources]" and "Allowing message sending and receiving using [`"externally_connectable.matches"`]" in the list?

This sounds like a good idea 👍.

wizardofdocs: > > So is it useful to say, for example, "Granting access to [web-accessible resources]" and "Allowing message sending and receiving using [`"externally_connectable.matches"`]" in the list?
> 
> This sounds like a good idea 👍.

Done in the new commit.

site/en/docs/extensions/mv3/match_patterns/index.md
jpmedley: @AmySteam

site/en/docs/extensions/mv3/match_patterns/index.md
jpmedley: @AmySteam

AmySteam: The  `/foo*` example below is for the **_path_**, not the **_host_**. So this is correct.

site/en/docs/extensions/mv3/match_patterns/index.md
```diff
@@ -3,159 +3,77 @@ layout: 'layouts/doc-post.njk'
...
-- **host**: A hostname (`www.example.com`), a `*` before the hostname to match subdomains (`*.example.com`), or just a wildcard `*`.
+**path**: Must contain at least a forward slash. The slash by itself is treated as a wildcard `/*`. For example:
+* `/` or `/*`
+* `/foo*`
+* `/foo/bar`
 
-- **path**: Must start with `/` and be present. If alone, it is treated as `/*`. For example: `/*`, `/foo*`, or `/foo/bar`. Each '`*`' matches 0 or more characters.
+Extensions use match patterns in a variety of use cases, including the following:
+
+* [Declaring permissions](/docs/extensions/mv3/declare_permissions) for Chrome APIs and [content script][content-scripts] injection
```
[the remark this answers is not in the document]
jpmedley: +1

site/en/docs/extensions/mv3/match_patterns/index.md
```diff
@@ -3,159 +3,78 @@ layout: 'layouts/doc-post.njk'
...
+`urn:*`
+: Matches any URL that starts with `urn:`. Examples of matches include `rn:uuid:54723bea-c94e-480e-80c8-a69846c3f582`.
 
 [content-scripts]: /docs/extensions/mv3/content_scripts
 [cs-frames]: /docs/extensions/mv3/content_scripts/#injecting-in-related-frames
 [ext-connect]: /docs/extensions/mv3/manifest/externally_connectable/#manifest
-[host-permissions]: /docs/extensions/mv3/declare_permissions/#host-permissions
 [mdn-tld]: https://developer.mozilla.org/docs/Glossary/TLD
-[permissions]: /docs/extensions/mv3/declare_permissions/
+[permissions]: /docs/extensions/mv3/declare_permissions/#allow_access
```
AmySteam, line 79:
```suggestion
[permissions]: /docs/extensions/mv3/declare_permissions/#allow_access
[host-permissions]: /docs/extensions/mv3/declare_permissions/#host-permissions
```
Fix broken link

site/en/docs/extensions/mv3/match_patterns/index.md
AmySteam: > '/' means the same thing as '/*', but do we really need to tell people that? It's what linguist Bryan Garner would call a [needless variant](https://www.cjr.org/analysis/there_ought_to_be_alternative.php#:~:text=might%20call%20a%20%E2%80%9C-,needless%20variant,-%E2%80%9D%3A%20%E2%80%9Ctwo%20or%20more)

It's just more explicit for developers to know if you add `*.example.com/`, it will also match `*.example.com/foo` and `*.example.com/bar`, not just `www.example.com`.

I think when it comes to match patterns, it's okay to be explicit.  But at this point I'm ready to agree with anything ya'll decide.

jpmedley:
@oliverdunk Are you done?

site/en/docs/extensions/mv3/match_patterns/index.md
wizardofdocs: Is this phrasing better?

site/en/docs/extensions/mv3/match_patterns/index.md
```diff
@@ -20,10 +20,7 @@ A match pattern is a URL with the following structure, used to specify a group o
 **host**: A hostname (`www.example.com`). A `*` before the hostname to match subdomains (`*.example.com`), or just a wildcard `*`.
   - If you use a wildcard in the host pattern, it must be the first or only character, and it must be followed by a period (`.`) or forward slash (`/`).
 
-**path**: Must contain at least a forward slash. The slash by itself is treated as a wildcard `/*`. For example:
-* `/` or `/*`
-* `/foo*`
-* `/foo/bar`
+**path**: Must contain at least a forward slash. The slash by itself matches any path, as if it were followed by a wildcard (`/*`).
```
AmySteam, line 23: Love it!!! 💯 🤩

site/en/docs/extensions/mv3/match_patterns/index.md
AmySteam: image

jpmedley (approved):

jpmedley:
@oliverdunk, are you done?

site/en/docs/extensions/mv3/match_patterns/index.md
```diff
@@ -3,159 +3,76 @@ layout: 'layouts/doc-post.njk'
...
-
-A match pattern is a URL that begins with a permitted scheme, separated by a `://`, followed by a host and path. It can contain wildcard (`*`) characters. Most match patterns are composed of three parts:
+A match pattern is a URL with the following structure, used to specify a group of URLs:
 
 ```text
 &lt;scheme&gt;://&lt;host&gt;/&lt;path&gt;
 ```
 
-Each part can use wildcards `*`. Below is a detailed description:
+**scheme**: Must be one of `'http'`, `'https'`, `'file'`, `'ftp'`, or `'urn'`. The wildcard `*` only matches `http` or `https`.
```
oliverdunk, line 16: Are we sure about `urn`?

site/en/docs/extensions/mv3/match_patterns/index.md
```diff
@@ -3,159 +3,76 @@ layout: 'layouts/doc-post.njk'
...
 
-- **host**: A hostname (`www.example.com`), a `*` before the hostname to match subdomains (`*.example.com`), or just a wildcard `*`.
+**path**: Must contain at least a forward slash. The slash by itself matches any path, as if it were followed by a wildcard (`/*`).
 
-- **path**: Must start with `/` and be present. If alone, it is treated as `/*`. For example: `/*`, `/foo*`, or `/foo/bar`. Each '`*`' matches 0 or more characters.
+Extensions use match patterns in a variety of use cases, including the following:
+
+* Injecting [content script][content-scripts].
+* [Declaring host permissions][host-permissions] that some Chrome APIs require in addition to their own permissions.
+* Granting access to [web-accessible resources][war]
```
oliverdunk, line 29:
```suggestion
* Granting access to [web-accessible resources][war].
```

site/en/docs/extensions/mv3/match_patterns/index.md
```diff
@@ -3,159 +3,76 @@ layout: 'layouts/doc-post.njk'
...
-         <td><code>about:*</code></td>
-         <td>Unsupported <em>scheme</em>.</td>
-      </tr>
-   </tbody>
-</table>
-
-{% Aside 'success' %}
-Content scripts can be injected into related frames like `about:`, `data:`, etc by setting `"match_origin_as_fallback"` to `"true"` in the manifest. See [Injecting in related frames][cs-frames] for details.
-{% endAside %}
+: Chrome doesn't support match patterns for [top Level domains (TLD)][mdn-tld]. Specify your match patterns within individual TLDs, as in `http://google.es/*` and `http://google.fr/*`.
```
oliverdunk, line 44: What does this mean? That you can't have `*.co.uk`?

site/en/docs/extensions/mv3/match_patterns/index.md
```diff
@@ -3,159 +3,76 @@ layout: 'layouts/doc-post.njk'
...
+: Matches any URL that uses the `http` scheme and is on the host 127.0.0.1. Examples of matches include `http://127.0.0.1/` and `http://127.0.0.1/foo/bar.html`.
+
+`http://localhost/*`
+: Matches any localhost port.
+
+`*://mail.google.com/` or `*://mail.google.com/*`
+: Matches any URL that starts with `http://mail.google.com` or `https://mail.google.com`.
+
+`urn:*`
+: Matches any URL that starts with `urn:`. Examples of matches include `rn:uuid:54723bea-c94e-480e-80c8-a69846c3f582`.
```
oliverdunk, line 70: We should drop this if it doesn't work (it doesn't as far as I can tell from testing).

site/en/docs/extensions/mv3/match_patterns/index.md
```diff
@@ -3,159 +3,76 @@ layout: 'layouts/doc-post.njk'
...
 
-`file:///`
-: Allows your extension to run on local files. It has no host and requires the user to manually [grant access][permissions-allow-access].
+`"file:///"`
+: Allows your extension to run on local files. This pattern requires the user to manually [grant access][permissions]. Note that this case requires three slashes, not two.
 
 Localhost URLs and IP addresses
-: To match any localhost port during development, use `http://localhost:*/*`. For example, it will match any paths under `http://localhost:8080` or `http://localhost:3000`.  For IP addresses, specify the address plus a wildcard in the path. For example: `http://127.0.0.1/*`. Another approach is to use `http://*:*/*` which will match localhost, IP addresses and any port.
+: To match any localhost port during development, use `http://localhost:*/*`. Specify a port using the port number, as in `http://localhost:8080`.  For IP addresses, specify the address plus a wildcard in the path, as in `http://127.0.0.1/*`. You can also use `http://*:*/*` to match localhost, IP addresses, and any port.
```
oliverdunk, line 41:
```suggestion
: To match any localhost port during development, use `http://localhost:*/*`. Specify a port using the port number, as in `http://localhost:8080/*`.  For IP addresses, specify the address plus a wildcard in the path, as in `http://127.0.0.1/*`. You can also use `http://*:*/*` to match localhost, IP addresses, and any port.
```

oliverdunk (changes requested):
Left a few final comments. Thanks for the continued work on this one.

site/en/docs/extensions/mv3/match_patterns/index.md
jpmedley: I'm trying to find out.

jpmedley: If I understood Devlin correctly, this is the correct list of schemes. https://source.chromium.org/chromium/chromium/src/+/main:extensions/common/url_pattern.cc;l=31-52;drc=8c08dd32bf007ca09f57b63b6b812478b44ac2c4

0569395312:
SA‎8880000203608010980048

0569395312

في الأربعاء، ٢٠ سبتمبر ٢٠٢٣, ٧:٢٣ م Joe Medley ***@***.***>
كتب:

0569395312:
https://drive.google.com/file/d/1pV-qenYzqnmr3RV_t1BeMzDHzmBT4CKy/view?usp=drivesdk

0569395312

في الأربعاء، ٢٠ سبتمبر ٢٠٢٣, ٧:٢٤ م عثمان شافعي ***@***.***>
كتب:

site/en/docs/extensions/mv3/match_patterns/index.md
wizardofdocs: The original says you can't do `google.*`. `*.example.com` is allowed.

site/en/docs/extensions/mv3/match_patterns/index.md
wizardofdocs: Updated. Does it still need the single quotes?

site/en/docs/extensions/mv3/match_patterns/index.md
wizardofdocs: Removed.

site/en/docs/extensions/mv3/match_patterns/index.md
jpmedley: I'd say no quotes. I insist on quotes for manifest keys because they're part of what actually needs to be in the manifest. Quotes aren't part of a URL scheme.

wizardofdocs: Done.

site/en/docs/extensions/mv3/match_patterns/index.md
jpmedley: It would need to be `*://*.co.uk`, which is [in one of our unit tests](https://source.chromium.org/chromium/chromium/src/+/main:extensions/common/url_pattern_unittest.cc;l=1056;drc=707b3abc38ded8ab7eb0fa21664b1556324929b4). I'm told that gets the same result as `<all_urls>`. So, should we care?

site/en/docs/extensions/mv3/match_patterns/index.md
```diff
@@ -13,9 +13,22 @@ A match pattern is a URL with the following structure, used to specify a group o
...
+- `'https'`
+- A wildcard `*`, which matches only `http` or `https`
+- `'file'`
+- `'ftp'`
+- `'chromeui'`
+- `'extension'`
+- `'filesystem'`
+- `'ws'`
+- `'wss'`
+- `'data'`
```
AmySteam: I'm a bit uncomfortable adding all these scheme's because I know that extensions have restrictions. For example, what is the 'extension' scheme? If it's `chrome-extension://` URLs, these are not valid for content script injection.

jpmedley: This list came from source code: https://source.chromium.org/chromium/chromium/src/+/main:extensions/common/url_pattern.cc;l=31-52;drc=8c08dd32bf007ca09f57b63b6b812478b44ac2c4

You're probably right about extension UI. (And a few other probably need to be spelled differently.) But confirm that in code yet.

site/en/docs/extensions/mv3/match_patterns/index.md
jpmedley: @oliverdunk

site/en/docs/extensions/mv3/match_patterns/index.md
oliverdunk: Notes:

- Chrome no longer supports FTP so I don't think `ftp` makes sense.
- I'm not familiar with (and couldn't find) any use case for `chromeui`. It seems internal so I would remove.
- The same for `extension` and `filesystem`.
- Looking at the code, we don't seem to actually support `data` in any APIs, so I would remove that.
- `uuid_in_package` was removed here: https://groups.google.com/a/chromium.org/g/blink-dev/c/iyhes6tMz2Y

So I think the full list should be:

```
http
https
A wildcard *, which matches only http or https
file
```

site/en/docs/extensions/mv3/match_patterns/index.md
AmySteam: The TLDs section was ported directly from the extension's FAQs article. This may have been so in an older version of Chrome, since @oliverdunk has been able to test that content scripts can be injected into TLDs match patterns.

Mohammed71o:
T ou criss, là, vous vous dite développer,non pirate

Le mar. 10 oct. 2023, 12 h 37, amysteamdev ***@***.***> a
écrit :

Mohammed71o:
Ta rive tu je reste rais pas dehors toute la journée

Mohammed71o:
Désolé madame mais il y a quelqu'un de githud qui dit venir ici

site/en/docs/extensions/mv3/match_patterns/index.md
jpmedley: @AmySteam, I didn't follow that.

site/en/docs/extensions/mv3/match_patterns/index.md
jpmedley: +1

site/en/docs/extensions/mv3/match_patterns/index.md
jpmedley: @oliverdunk told me over chat that the current text is fine.l

jpmedley:
@oliverdunk I can't merge this without your LGTM.

site/en/docs/extensions/mv3/match_patterns/index.md
```diff
@@ -3,159 +3,86 @@ layout: 'layouts/doc-post.njk'
...
+- A wildcard `*`, which matches only `http` or `https`
+- `file`
+- `ftp`
+- `chromeui`
+- `extension`
+- `filesystem`
+- `ws`
+- `wss`
+- `data`
+- `uuid_in_package`
```
AmySteam, line 29:
```suggestion
- `http`
- `https`
- A wildcard `*`, which matches only `http` or `https`
- `file`
```
@wizardofdocs @jpmedley 
Could we apply this change before merging? [Here's the conversation where we agreed to simplify this list](https://github.com/GoogleChrome/developer.chrome.com/pull/7215#discussion_r1352008813)

AmySteam (changes requested):
One request change before we merge please. Based on our team discussion:

https://github.com/GoogleChrome/developer.chrome.com/pull/7215#discussion_r1352008813

AmySteam (approved):
Thanks! 👍

oliverdunk (approved):
LGTM, thanks for the patience on some of the trickier bits here 👍

PR #7238, opened 2023-09-04
Fix prerender analytics

tunetheweb:

- Move OT token up to top of page (so it is registered ASAP)
- Do not initialise analytics until page is activated (to allow it to get the correct navigation type)
- Remove redundant speculation rules setting

The second one is the main thing I want as noticed the prerender page type is not being set correctly.


matthiasrohmer (approved):
LGTM, once checks pass.

PR #7247, opened 2023-09-06
Refactor 'Improve extension security'.

jpmedley:
Gang,

I've been looking for an example of why we need to integrate new material carefully. Though this PR doesn't change this section's architecture, I think it illustrates the point. This page will be six months old on Friday, and we've already mangled the structure. (This one is on me. I created the page, so I should have been paying closer attention.)

Yesterday, I tried to add a caution about third-party bundles and remotely-hosted code. It was when I tried to resolve a merge conflict that I realized this needed more than an aside. The result is this PR. Please have a look.

@sebastianbenz @patrickkettner @AmySteam

### DIFF
--- a/site/en/docs/extensions/migrating/improve-security/index.md
+++ b/site/en/docs/extensions/migrating/improve-security/index.md
@@ -4,7 +4,7 @@ title: Improve extension security
 subhead: 'Improving security in Manifest V3'
 description: 'The last of three sections describing changes needed for code that is not part of the extension service worker.'
 date: 2023-03-08
-updated: 2023-09-06
+updated: 2023-09-26
 ---
 
 {% Partial 'extensions/mv3-support.md' %}
@@ -34,8 +34,7 @@ In Manifest V3, all of your extension's logic must be part of the extension pack
 
 - JavaScript files pulled from the developer's server.
 - Any library hosted on a [CDN][mdn-cdn].
-- Some third-party libraries.
-- How third-party libraries are used by your code.
+- Bundled third-party libraries that dynamically fetch remote hosted code.
 
 Alternative approaches are available, depending on your use case and the reason for remote hosting. This section describes approaches to consider.
 
--- a/site/en/docs/extensions/migrating/improve-security/index.md
+++ b/site/en/docs/extensions/migrating/improve-security/index.md
@@ -118,7 +118,7 @@ In the background service worker.
 {% endCompare %}
 </div>
 
-The [Chrome Extension Samples repo](https://github.com/GoogleChrome/chrome-extensions-samples.git) contains a [function injection example](https://github.com/GoogleChrome/chrome-extensions-samples/blob/main/reference/mv3/intro/mv3-migration/content-scripts/popup.js) you can step through. An example of `getCurrentTab()` is in the [reference](/docs/extensions/reference/tabs/#get-the-current-tab) for that function.
+The [Chrome Extension Samples repo](https://github.com/GoogleChrome/chrome-extensions-samples.git) contains a [function injection example](https://github.com/GoogleChrome/chrome-extensions-samples/blob/main/functional-samples/reference.mv3-content-scripts/popup.js) you can step through. An example of `getCurrentTab()` is in the [reference](/docs/extensions/reference/tabs/#get-the-current-tab) for that function.
 
 ### Look for other workarounds {: #look-for-workarounds }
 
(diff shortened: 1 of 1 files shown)


jpmedley:
@oliverdunk You might wait a bit before doing a formal review.

site/en/docs/extensions/migrating/improve-security/index.md
```diff
@@ -29,36 +30,45 @@ There are a few special cases in which executing arbitrary strings is still poss
 
 ## Remove remotely hosted code {: #remove-remote-code }
 
-In Manifest V3, all of your extension's logic must be part of the extension package. You can no longer load and execute remotely hosted files. Examples include:
+In Manifest V3, all of your extension's logic must be part of the extension package. You can no longer load and execute remotely hosted files according to [Chrome Web Store policy](/docs/webstore/program-policies/mv3-requirements/). Examples include:
 
 - JavaScript files pulled from the developer's server.
 - Any library hosted on a [CDN][mdn-cdn].
+- Some third-party libraries.
```
sebastianbenz, line 37: I find this a bit unspecified, I'd suggest combining the last two items into:

```
- Bundled third-party libraries that dynamically fetch remote hosted code.
```

site/en/docs/extensions/migrating/improve-security/index.md
```diff
@@ -29,36 +30,45 @@ There are a few special cases in which executing arbitrary strings is still poss
...
-    ```
-    In some cases, libraries, such as Firebase, will fetch additional code on a by need basis at runtime. In this case, you have to download all possible dynamic imports at build time. Here is an [example script that bundles Firebase using Rollup.js][firebase].
+Your extension calls a remote web service. This lets you keep code private and change it as needed while avoiding the extra overhead of resubmitting to the Chrome Web Store.
 
-Look for other workarounds
-: If the previous approaches don’t help with your use case you might have to either find an alternative solution (i.e. migrate to a different library) or find other ways to use the library's functionality. For example, in the case of Google Analytics, you can switch to the Google measurement protocol instead of using the official remote hosted JavaScript version as described in our [Google Analytics 4 guide][google-analytics].
+### Embed remotely hosted code in a sandboxed iframe {: #embed-in-sandbox }
+
+Remotely hosted code [is supported in sandboxed iframes][sandbox-eval]. Unfortunately, this approach does not work if the code requires access to the embedding page's DOM.
```
sebastianbenz, line 52:
```suggestion
Remotely hosted code [is supported in sandboxed iframes][sandbox-eval]. Please note that this approach does not work if the code requires access to the embedding page's DOM.
```

`Unfortunately` is the wrong word here as it is "by design" and intended.

sebastianbenz (approved):
Two suggestions, but otherwise LGTM. Thanks for fixing this!

site/en/docs/extensions/migrating/improve-security/index.md
jpmedley: Done.

site/en/docs/extensions/migrating/improve-security/index.md
```diff
@@ -111,6 +120,10 @@ In the background service worker.
 
 The [Chrome Extension Samples repo](https://github.com/GoogleChrome/chrome-extensions-samples.git) contains a [function injection example](https://github.com/GoogleChrome/chrome-extensions-samples/blob/main/reference/mv3/intro/mv3-migration/content-scripts/popup.js) you can step through. An example of `getCurrentTab()` is in the [reference](/docs/extensions/reference/tabs/#get-the-current-tab) for that function.
```
AmySteam, line 121:
```suggestion
The [Chrome Extension Samples repo](https://github.com/GoogleChrome/chrome-extensions-samples.git) contains a [function injection example](https://github.com/GoogleChrome/chrome-extensions-samples/blob/main/functional-samples/reference.mv3-content-scripts/popup.js) you can step through. An example of `getCurrentTab()` is in the [reference](/docs/extensions/reference/tabs/#get-the-current-tab) for that function.
```

site/en/docs/extensions/migrating/improve-security/index.md
```diff
@@ -111,6 +120,10 @@ In the background service worker.
...
 
+### Look for other workarounds {: #look-for-workarounds }
+
+If the previous approaches don’t help with your use case you might have to either find an alternative solution (i.e. migrate to a different library) or find other ways to use the library's functionality. For example, in the case of Google Analytics, you can switch to the Google measurement protocol instead of using the official remote hosted JavaScript version as described in our [Google Analytics 4 guide][google-analytics].
```
AmySteam, line 125:
```suggestion
If the previous approaches don’t help with your use case you might have to either find an alternative solution (i.e. migrate to a different library) or find other ways to use the library's functionality. For example, in the case of Google Analytics, you can switch to the Google measurement protocol instead of using the official remotely-hosted JavaScript version as described in our [Google Analytics 4 guide][google-analytics].
```

AmySteam (approved):
I'm approving this to not be a blocker, but I'm suggesting a link fix and a spelling tweak.

PR #7262, opened 2023-09-08
Add note about GA and prerender support

tunetheweb:

- Adds note about GA now supporting Prerender


malchata (approved):
Teeny nit, otherwise LGTM.

PR #7306, opened 2023-09-14
Adjust the next version number since it was written as '117'

ahmet-ustun:


rachelandrew:
The original is correct, it refers to the next release.

PR #7310, opened 2023-09-15
Add monitorTypeSurfaces option

beaufortfrancois:
This PR adds the monitorTypeSurfaces option shipping in Chrome 119.

@eladalon1983 Please have a look.

### DIFF
--- a/site/en/docs/web-platform/screen-sharing-controls/index.md
+++ b/site/en/docs/web-platform/screen-sharing-controls/index.md
@@ -7,17 +7,19 @@ authors:
   - beaufortfrancois
   - eladalon
 date: 2022-09-29
-updated: 2023-03-14
+updated: 2023-09-15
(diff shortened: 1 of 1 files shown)


eladalon1983:
LGTM.
Please consider if we should mention what happens if this is set in conjunction with a few other settings, such excluding monitors while *preferring* them.

beaufortfrancois:
> LGTM. Please consider if we should mention what happens if this is set in conjunction with a few other settings, such excluding monitors while _preferring_ them.

I had added the note below for that purpose:

```
Note that an explicit `monitorTypeSurfaces: "exclude"` is mutually exclusive with `displaySurface: "monitor"`.
```

eladalon1983:
It looks like my multitasking game is not where I thought it was. I missed that line. Cool.

beaufortfrancois:
@rachelandrew Please review.

rachelandrew (approved):

PR #7339, opened 2023-09-19
Add correction to Decision: Batching Strategy section in index.md

danghle:
A more frequent batching strategy implies fewer conversions which implies less signal to noise, a fact that is reiterated in the documents linked at the end of this section.


samdutton:
Hi @maudnals — I know you're not working on ARA now, but would you be able to review this one word change? Thanks.

nancymic2 (approved):
This keeps getting mixed up and I feel like "signal to noise " should maybe not be used?

maudnals (commented):
LGTM! This was indeed a mistake.

PR #7340, opened 2023-09-19
Make it more explicit that `effectiveConnectionType` is not supported

iamakulov:
This PR documents that the `effectiveConnectionType` field is not supported in CrUX History API requests.

Right now, this is mentioned in one or two places in the article but not in the Schema section, which might be misleading (it misled me a bit 🙂).

### DIFF
--- a/site/en/docs/crux/history-api/index.md
+++ b/site/en/docs/crux/history-api/index.md
@@ -314,7 +314,7 @@ The URL uses [gRPC Transcoding](https://google.aip.dev/127) syntax.
 
 ### Request body
 
-As the CrUX History API uses the same request bodies, you can reference [the daily CrUX API request body documentation](../api/#request-body) for more details.
+The CrUX History API uses the same request bodies as [the daily CrUX API](../api/#request-body), with the exception of not supporting the `effectiveConnectionType` request field.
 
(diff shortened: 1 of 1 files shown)


tunetheweb (approved):

site/en/docs/crux/history-api/index.md
```diff
@@ -314,7 +314,7 @@ The URL uses [gRPC Transcoding](https://google.aip.dev/127) syntax.
 
 ### Request body
 
-As the CrUX History API uses the same request bodies, you can reference [the daily CrUX API request body documentation](../api/#request-body) for more details.
+The CrUX History API uses the same request bodies as [the daily CrUX API](../api/#request-body), with the exception of not supporting the `effectiveConnectionType` request field.
```
malchata, line 317: Would strongly prefer an absolute URL here, as opposed to relative.

malchata (approved):
LGTM, with a minor nit about a relative URL reference.

site/en/docs/crux/history-api/index.md
tunetheweb:
```suggestion
The CrUX History API uses the same request bodies as [the daily CrUX API](/docs/crux/api/#request-body), with the exception of not supporting the `effectiveConnectionType` request field.
```

PR #7345, opened 2023-09-20
Soft Navs better explanation of LCP time

tunetheweb:

- Show how LCP time is calced in a soft nav
- Reordered the doc a little to make it make more sense for that


malchata (approved):
LGTM

PR #7398, opened 2023-09-27
Consolidate action material

jpmedley:
If the action material is on the same page is should be in the same place.

This will make sense if you compare the staged rendering to the current version of the page.

### DIFF
--- a/site/en/docs/extensions/mv3/user_interface/index.md
+++ b/site/en/docs/extensions/mv3/user_interface/index.md
@@ -154,20 +154,6 @@ chrome.action.setBadgeText({text: 'ON'});
 chrome.action.setBadgeBackgroundColor({color: '#4688F1'});
 ```
 
-## Provide the extension icons
-
-An extension requires at least one icon to represent it. Provide icons in PNG format for the best
-visual results, although any raster format supported by Chrome is accepted. This includes BMP,
-GIF, ICO, and JPEG.
-
-{% Aside 'caution' %}
-
-SVG files are not supported for any icons declared in the manifest.
-
-{% endAside %}
-
-Ensure your icon follows the [extension icon best practices][docs-icon-guidelines].
-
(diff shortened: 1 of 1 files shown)


AmySteam:
If you are moving all Action APIs into the same section, would the popup and tooltip also go under [The extension action](https://pr-7398-static-dot-dcc-staging.uc.r.appspot.com/docs/extensions/mv3/user_interface/#action)?

AmySteam (changes requested):
Can we explore moving the other action API items into the same section? Other than that it LGTM 👍

jpmedley:
> Can we explore moving the other action API items into the same section?

Do you mean this: https://pr-7398-static-dot-dcc-staging.uc.r.appspot.com/docs/extensions/mv3/user_interface/#icons

Yeah, I failed to notice that one. I can move it now.

jpmedley:
@AmySteam Would you mind looking at [my latest change](https://github.com/GoogleChrome/developer.chrome.com/pull/7398/commits/6d5ef77cbe14abe76d36724c7ba8524bff9e0550). Make sure I didn't introduce any new problems.

AmySteam:
> > Can we explore moving the other action API items into the same section?
> 
> Do you mean this: https://pr-7398-static-dot-dcc-staging.uc.r.appspot.com/docs/extensions/mv3/user_interface/#icons
> 
> Yeah, I failed to notice that one. I can move it now.

I meant the tooltip https://pr-7398-static-dot-dcc-staging.uc.r.appspot.com/docs/extensions/mv3/user_interface/#tooltip and the popup, since they are both part of the Action API. 

Sorry I should have been mre explicit

jpmedley:
I moved the popup section too. You should probably look at this again.

jpmedley:
These are good catches. I don't know why I didn't think of them.

AmySteam (approved):
LGTM 👍

PR #7440, opened 2023-10-04
Add JSON to the list of web technologies

jpmedley:


jpmedley:
@AmySteam, the link to MDN was already in the file.

AmySteam (commented):
I think this list if getting too big already. My inclination would be no, but if any of the DRE's thinks it's valuable... sure...

@oliverdunk @sebastianbenz @patrickkettner  Thoughts?

patrickkettner:
agree the llist is long. But I thnk JSON is less liekly to be known then CSS, so 🤷‍♀️

jpmedley:
I added some context that I think will make this list less scary for potential extension developers.

I think it's important to give complete information to potential extension developers. This also saves us some reworking in another document.

jpmedley:
Pinging on this again. Thoughts?

oliverdunk:
> Pinging on this again. Thoughts?

To be honest, I'm not sure that this is particularly helpful (JSON is implied by JS for the most part, as is knowledge of web APIs). I think this change may actually make things more verbose but I'm not strongly opposed if we decide we really want it.

AmySteam:
> > Pinging on this again. Thoughts?
> 
> To be honest, I'm not sure that this is particularly helpful (JSON is implied by JS for the most part, as is knowledge of web APIs). I think this change may actually make things more verbose but I'm not strongly opposed if we decide we really want it.

I liked the simplicity of the original approach: CSS, HTML, and JS.

sebastianbenz:
+1 to the other's feedback, I think this update is well intended but not really helpful / actionable for readers.

oliverdunk:
I still feel somewhat indifferent on this. I'm not against it, but I'm not sure it adds much. Will defer to others on if we make the  change :)

PR #7496, opened 2023-10-10
Remove aside from Commands API

### LINKED ISSUE #4311: Update chrome.commands documentation on _execute_action
In the [chrome.commands](https://developer.chrome.com/docs/extensions/reference/commands/#verify-commands-registered) documentation, a comment mentions that `_execute_action` is not returned from `commands.getAll()`:

Screenshot 2022-11-15 at 18 07 51

I was able to fix this as part of https://crbug.com/1375790 in https://chromium.googlesource.com/chromium/src/+/4464f977da2eb9e8e1e0eb9775d71d20e845eb52. This means that the comment is no longer accurate from Chrome 110 (currently in Canary).

There were two suggestions made in the [comments](https://chromium-review.googlesource.com/c/chromium/src/+/4020384/comments/359f20ac_cb460ec4) of that review:

- The callout should be updated and eventually removed.
- It may make sense to include this on https://developer.chrome.com/docs/extensions/whatsnew/.

I could also see an argument for adding it to https://developer.chrome.com/docs/extensions/mv3/known-issues/.

AmySteam:
Fixes #4311

- Removes aside warning of a bug that has been resolved
- Fixes code sample

### DIFF
--- a/site/en/docs/extensions/reference/commands/index.md
+++ b/site/en/docs/extensions/reference/commands/index.md
@@ -280,7 +280,7 @@ experience by anticipating this possibility and checking for collisions at insta
 
 ```js
 chrome.runtime.onInstalled.addListener(({reason}) => {
-  if (reason === 'install') {
+  if (reason === chrome.runtime.OnInstalledReason.INSTALL) {
     checkCommandShortcuts();
   }
 });
--- a/site/en/docs/extensions/reference/commands/index.md
+++ b/site/en/docs/extensions/reference/commands/index.md
@@ -276,18 +276,11 @@ If an extension attempts to register a shortcut that is already used by another 
 second extension's shortcut will not register as expected. You can provide a more robust end user
 experience by anticipating this possibility and checking for collisions at install time.
 
-{% Aside %}
-
-`_execute_action` will not appear in the list
-of commands returned by `commands.getAll()`.
-
-{% endAside %}
-
 {% Label %}service-worker.js:{% endLabel %}
 
 ```js
-chrome.runtime.onInstalled.addListener((reason) => {
-  if (reason === chrome.runtime.OnInstalledReason.INSTALL) {
+chrome.runtime.onInstalled.addListener(({reason}) => {
+  if (reason === 'install') {
     checkCommandShortcuts();
   }
 });


site/en/docs/extensions/reference/commands/index.md
```diff
@@ -276,18 +276,11 @@ If an extension attempts to register a shortcut that is already used by another
...
-
-{% endAside %}
-
 {% Label %}service-worker.js:{% endLabel %}
 
 ```js
-chrome.runtime.onInstalled.addListener((reason) => {
-  if (reason === chrome.runtime.OnInstalledReason.INSTALL) {
+chrome.runtime.onInstalled.addListener(({reason}) => {
+  if (reason === 'install') {
```
oliverdunk, line 283: Did we intentionally move away from the enum here? Generally I think the enum is better whenever possible.

AmySteam: I can never get this to work on my code `chrome.runtime.OnInstalledReason.INSTALL)`, why is it superior?

Mohammed71o:
Je suis dans githud là,je veux pas savoir ses quoi,je juste vous dire que
si vous avez pas ses chose là,je suis c0mpreensife.

Le mar. 10 oct. 2023, 12 h 03, amysteamdev ***@***.***> a
écrit :

site/en/docs/extensions/reference/commands/index.md
oliverdunk: Hmm, it should. Let me know if you want help debugging. In reality it's unlikely to change but it's generally considered better practice because if the string itself changes your code would still work, and it also says a bit more explicitly where the value came from.

AmySteam, line 283:
```suggestion
chrome.runtime.onInstalled.addListener(({reason}) => {
  if (reason === chrome.runtime.OnInstalledReason.INSTALL) {
```

AmySteam: fair. But I'm leaving the brackets around reason, if not it won't work

oliverdunk: Of course! Thanks for spotting and fixing that.

oliverdunk (approved):
LGTM!

AmySteam:
@jpmedley Can you approve this? It doesn't need a copy edit. Thanks!

jpmedley (approved):
Nice catch on the callback's data type.

PR #7501, opened 2023-10-10
Updated Action reference

wizardofdocs:
Regrouped headers and updated phrasing

### DIFF
--- a/site/en/docs/extensions/reference/action/index.md
+++ b/site/en/docs/extensions/reference/action/index.md
@@ -6,7 +6,7 @@ To control your extension's toolbar button in Chrome's UI, use the `chrome.actio
 The action icons are displayed in the browser toolbar next to the [omnibox](https://en.wiktionary.org/wiki/omnibox). After installation, these appear in the extensions menu (the puzzle piece icon).
 Users can pin your extension icon to the toolbar.
 
-Every extension has an icon in the Chrome toolbar, even if the `action` key isn't specified.
+Every extension has an icon in the Chrome toolbar, even if the `action` key isn't added to the manifest.
 
 ## Manifest
 
--- a/site/en/docs/extensions/reference/action/index.md
+++ b/site/en/docs/extensions/reference/action/index.md
@@ -30,7 +30,7 @@ and include the `"action"` key in your [manifest file][doc-manifest]:
 }
 ```
 
-<!--I want to say something like "while the dictionary values are all optional, we recommend that you include at least $these." What are the these?-->
+The `"action"` key (along with its children) is optional. When it isn't included, your extension is still shown in the toolbar to provide access to the extension's menu. For this reason, we recommend that you always include at least the `"action"` and `"default_icon"` keys.
 
 ## Concepts and Usage
 
--- a/site/en/docs/extensions/reference/action/index.md
+++ b/site/en/docs/extensions/reference/action/index.md
@@ -41,7 +41,7 @@ and include the `action` key in your [manifest file][doc-manifest]:
 The icon is the main image on the toolbar for your extension, and is set by the `"default_icon"` key in
 your manifest's `action` entry. Icons must be 16 device-independent pixels (DIPs) wide and tall. 
 
-The `default_icon` key is a dictionary of sizes to image paths. Chrome uses these icons
+The `"default_icon"` key is a dictionary of sizes to image paths. Chrome uses these icons
 to choose which image scale to use. If an exact match is not found, Chrome selects the closest
 available and scales it to fit the image, which might affect image quality.
 
@@ -80,7 +80,7 @@ The tooltip, or title, appears when the user holds their cursor over the extensi
 toolbar. It's also included in the accessible text spoken by screenreaders when the button gets
 focus.
 
-The default tooltip is set using the `default_title` field of the `action` object in `manifest.json`.
+The default tooltip is set using the `"default_title"` field of the `action` object in `manifest.json`.
 You can also set it programmatically using the `action.setTitle()` method.
 
 #### Badge
@@ -118,7 +118,7 @@ An action's popup is shown when the user clicks on the extension's action button
 toolbar. The popup can contain any HTML contents you like, and will be automatically sized to fit
 its contents. The popup's size must be between 25x25 and 800x600 pixels. <!--Are these the DIPs from earlier?-->
 
-The popup is initially set by the `default_popup` property in the `action` key in the
+The popup is initially set by the `"default_popup"` property in the `action` key in the
 `manifest.json` file. This property should point to a relative path within the extension
 directory. It can also be updated dynamically to point to a different relative path using the
 `action.setPopup()` method.
--- a/site/en/docs/extensions/reference/action/index.md
+++ b/site/en/docs/extensions/reference/action/index.md
@@ -80,7 +80,7 @@ The tooltip, or title, appears when the user holds their mouse pointer over the 
 toolbar. It's also included in the accessible text spoken by screen readers when the button gets
 focus.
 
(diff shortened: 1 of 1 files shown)


site/en/docs/extensions/reference/action/index.md
```diff
@@ -118,30 +112,33 @@ chrome.action.setBadgeBackgroundColor(
...
-The `action.onClicked` event will not be dispatched if the extension action has specified a popup to
+The `action.onClicked` event won't be sent if the extension action has specified a popup to
 show on click on the current tab.
 
 {% endAside %}
 
-## Per-tab state
+## Use cases
+
+<!--Is this a reasonable place to put Use Cases?-->
```
jpmedley, line 135:
```suggestion
```

jpmedley: Yes. This is exactly where we want them.

site/en/docs/extensions/reference/action/index.md
```diff
@@ -156,14 +153,14 @@ chrome.action.setBadgeText(
...
-If the `tabId` property is omitted, the setting is treated as a global setting. Tab-specific
-settings take priority over any global settings.
+If the `tabId` property is left out, the setting is treated as a global setting. Tab-specific
+settings take priority over global settings.
 
-## Enabled state
+### Enabled state
 
-By default, toolbar actions are enabled (clickable) on every tab. You can control this using the
+By default, toolbar actions are clickable on every tab. You can control this using the
```
jpmedley, line 161:
```suggestion
By default, toolbar actions are enabled (clickable) on every tab. You can control this using the
```

jpmedley: I think you should change this back. 'Clickable' is a clarifying remark for a word that reflects that methods that are used.

site/en/docs/extensions/reference/action/index.md
```diff
@@ -248,19 +245,19 @@ chrome.action.onClicked.addListener((tab) => {
...
 
-### Emulating pageActions with declarativeContent
+### Emulate pageActions with declarativeContent
 
-The chrome.action API replaced the [browserAction][api-browser-action] and
-[pageAction][api-page-action] APIs in Manifest V3. By default, actions are similar to browser
-actions, but it is possible to emulate the behavior of a page action using the action API.
+The `chrome.action` API replaced the [`browserAction`][api-browser-action] and
+[`pageAction`][api-page-action] APIs in Manifest V3. By default, actions are similar to Manifest V2's browser
+actions, but you can use the action API to emulate the behavior of page actions. <!--Do we want this to reference the old version? I can clean that up if needed.-->
```
jpmedley, line 252: Please do.

jpmedley (changes requested):

site/en/docs/extensions/reference/action/index.md
```diff
@@ -2,18 +2,16 @@
...
-The action icons are displayed in the browser toolbar, to the right of the omnibox (on left-to-right
-devices). After installation, by default, these appear in the extensions menu (the puzzle piece).
-Users can choose to pin your extension icon to the toolbar.
+To control your extension's toolbar button in Chrome's UI, use the `chrome.action` API.
+The action icons are displayed in the browser toolbar next to the omnibox. <!--can I just say "address bar," or is there a reason to use the technical term?--> After installation, by default, these appear in the extensions menu (the puzzle piece icon).
+Users can pin your extension icon to the toolbar.
 
-Note that every extension will have an icon in the toolbar in Chrome, even if the `action` key is
-not specified.
+Every extension has an icon in the Chrome toolbar, even if the `action` key isn't specified.
```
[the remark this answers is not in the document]
wizardofdocs: The style guide discourages "above" and "below" referring to other parts of the same document, because they cause problems with screen readers. They're also not very specific: what "below" are we telling the reader to look at? I don't think you mean the following section, because that doesn't have an example without an action key.

site/en/docs/extensions/reference/action/index.md
```diff
@@ -32,27 +30,27 @@ and include the `action` key in your [manifest file][doc-manifest].
 }
 ```
 
-Each of these values is optional; an empty dictionary is technically allowed.
+<!--I want to say something like "while the dictionary values are all optional, we recommend that you include at least $these." What are the these?-->
```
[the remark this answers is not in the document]
wizardofdocs: is this the below?

site/en/docs/extensions/reference/action/index.md
```diff
@@ -68,36 +66,32 @@ chrome.action.setIcon({imageData: imageData}, () => { /* ... */ });
...
+rendering engine can display, including PNG, JPEG, BMP, ICO, and others. SVG isn't supported.
+Unpacked extensions must use PNG images.
 
-### Tooltip (title)
+#### Tooltip (title)
 
-The tooltip, or title, appears when the user hovers the mouse on the extension's icon in the
-toolbar. It is also included in the accessible text spoken by screenreaders when the button gets
+The tooltip, or title, appears when the user holds their cursor over the extension's icon in the
+toolbar. It's also included in the accessible text spoken by screenreaders when the button gets
```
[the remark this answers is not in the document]
wizardofdocs: "It's," meaning "it is," needs the apostrophe and is the correct usage here. I'll fix "screen reader."

site/en/docs/extensions/reference/action/index.md
wizardofdocs: To clarify: what I want to do is remove the reference to how Manifest V2 works. Is that correct?

site/en/docs/extensions/reference/action/index.md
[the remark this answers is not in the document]
jpmedley: Yes.

site/en/docs/extensions/reference/action/index.md
```diff
@@ -118,30 +112,31 @@ chrome.action.setBadgeBackgroundColor(
...
 ```
 
-### Popup
+#### Popup
 
-An action's popup will be shown when the user clicks on the extension's action button in the
+An action's popup is shown when the user clicks on the extension's action button in the
 toolbar. The popup can contain any HTML contents you like, and will be automatically sized to fit
-its contents. The popup cannot be smaller than 25x25 and cannot be larger than 800x600.
+its contents. The popup's size must be between 25x25 and 800x600 pixels. <!--Are these the DIPs from earlier?-->
```
jpmedley, line 119:
```suggestion
its contents. The popup's size must be between 25x25 and 800x600 pixels. 
```

jpmedley: I've added this to the list of things to fix later.

site/en/docs/extensions/reference/action/index.md
```diff
@@ -248,19 +243,19 @@ chrome.action.onClicked.addListener((tab) => {
...
 
-### Emulating pageActions with declarativeContent
+### Emulate pageActions with declarativeContent
 
-The chrome.action API replaced the [browserAction][api-browser-action] and
-[pageAction][api-page-action] APIs in Manifest V3. By default, actions are similar to browser
-actions, but it is possible to emulate the behavior of a page action using the action API.
+The `chrome.action` API replaced the [`browserAction`][api-browser-action] and
+[`pageAction`][api-page-action] APIs in Manifest V3. By default, actions are similar to Manifest V2's browser
+actions, but you can use the action API to emulate the behavior of page actions. <!--Do we want this to reference the old version? I can clean that up if needed.-->
```
jpmedley, line 250:
```suggestion
```

jpmedley (changes requested):

jpmedley (approved):

PR #7508, opened 2023-10-11
Add callout about percentage rollout and new users

oliverdunk:
Just catching up on some of the docs we've put out recently - this one is great!

Spotted one thing that we didn't callout but might be useful. I've seen this surprise developers quite often, especially when looking at their analytics and seeing the version split.


sebastianbenz (approved):

AmySteam (approved):
LGTM 👍

PR #7513, opened 2023-10-11
Update web.dev links to new format

tunetheweb:

- Bulk updates all web.dev links to new formats (mostly in /articles with a few exceptions, no trailing slashes, anchor links use underscores instead of dashes).
- I've also tried to replace some redirects with the correct link as the above change will break the redirects.

This was a bulk update so I'm not claiming I've done everything correctly, and there will likely be some broken links (especially for redirects) but think it solves most of the real links.


rachelandrew (approved):
Didn't check each one by one, but a look through a representative selection looks sound. Thanks for this!

tunetheweb:
Anything to get a little better CWVs (by avoiding a redirect) 😁

Will give @devnook and/or @matthiasrohmer a chance to look over it, in the morning and then merge as don't wanna leave a big PR like this open for long to avoid merge conflicts as much as possible.

matthiasrohmer (approved):
Thank you, Barry! Haven't thought about the web.dev links we own.

PR #7532, opened 2023-10-13
Fix typo `function` is `func`

aiktb:

- Fix typo `function` is `func`

It can be found from the [chrome extension documentation](https://developer.chrome.com/docs/extensions/reference/scripting/#:~:text=to%20ISOLATED.-,func,-void%C2%A0optional) that there is no parameter named `function`, and the correct parameter is `func`.

Snipaste_2023-10-13_14-22-18
Snipaste_2023-10-13_14-26-07


AmySteam (approved):
Thanks @aiktb! Good catch 😄

PR #7537, opened 2023-10-13
Updated History reference and cleaned up navigation type table

wizardofdocs:
There's not a lot here. Are there more concepts or something that need to be on this page?

### DIFF
--- a/site/en/docs/extensions/reference/history/index.md
+++ b/site/en/docs/extensions/reference/history/index.md
@@ -3,9 +3,11 @@ api: history
 has_warning: This permission <a href="/docs/extensions/mv3/permission_warnings/#permissions_with_warnings">triggers a warning</a>.
 ---
 
+To interact with the user's browser history, use the history API.
+
 ## Manifest
 
-You must declare the "history" permission in the [extension manifest][1] to use the history API. For
+To use the history API, declare the `"history"` permission in the [extension manifest][1]. For
 example:
 
 ```json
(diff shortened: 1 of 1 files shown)


site/en/docs/extensions/reference/history/index.md
```diff
@@ -19,28 +21,52 @@ example:
...
 transition type is "link".
 
-The following table describes each transition type.
+The following are all available transition types:
+
+`"link"`
+: The user got to this page by clicking a link on another page.
+
+`"typed"`
+: The user got this page by typing the URL in the address bar. Also used for other explicit navigation actions. <!--okay, what else is there?-->
```
jpmedley, line 38:
```suggestion
: The user got this page by typing the URL in the address bar. This is also used for other explicit navigation actions. 
```

site/en/docs/extensions/reference/history/index.md
```diff
@@ -19,28 +21,52 @@ example:
...
+`"keyword_generated"`
+: Corresponds to a visit generated for a keyword. <!--How are keyword, keyword_generated, and generated different?-->
```
jpmedley, line 65:
```suggestion
: Corresponds to a visit generated for a keyword.
```

jpmedley (changes requested):

site/en/docs/extensions/reference/history/index.md
wizardofdocs: I'd like to make sure the user understands the difference, and I don't think what's here is sufficient. What else might readers need to know?

jpmedley: I've added keywords to the list.

PR #7541, opened 2023-10-13
Change the way pinning is presented.

jpmedley:
This lets me remove the same idea in a context where it doesn't even fit.


AmySteam (approved):
👍

PR #7548, opened 2023-10-16
Add requires policy tag

oliverdunk:
Adds a new tag which shows when an API is only available to policy installed extensions.

Matching chrome-types PR: https://github.com/GoogleChrome/chrome-types/pull/59


matthiasrohmer (approved):
We need to keep in mind to migrate this logic, in context of the proposal I sent around earlier.

Technically, this change looks good to me. Without knowing context: is this around https://chromeenterprise.google/policies/ or something else?

PR #7580, opened 2023-10-18
Remove colon from JSON keys

jpmedley:
We don't do this anywhere else. Frankly, I think we shouldn't since the colon only seems relevant in discussions of syntax, which this is not.

Fixes #SOME_ISSUE_NUMBER

-
-
-


AmySteam:
Joe, I have another PR open for this file. You can close this issue and I will fix it on mine end

jpmedley:
Sounds good.

PR #7584, opened 2023-10-18
Incorporate text that was in the FAQ

jpmedley:
A definition of extensions should be the first thing folks see.


AmySteam (approved):
LGTM 👍

PR #7636, opened 2023-10-24
Update documentation on static ruleset limits

oliverdunk:
Updates our documentation with the new raised limits for static rulesets.

Context: https://chromium-review.googlesource.com/c/chromium/src/+/4939511

### DIFF
--- a/site/en/docs/extensions/reference/declarativeNetRequest/index.md
+++ b/site/en/docs/extensions/reference/declarativeNetRequest/index.md
@@ -211,7 +211,12 @@ rule you're using.
 
 #### Static rules {: #static-rules }
 
-Static rules are those specified in rule files declared in the manifest file. An extension can specify up to 50 static [rulesets](#type-Ruleset) as part of the `"rule_resources"` manifest key, but only 10 of these rulesets can be enabled at a time. The latter is called the [`MAX_NUMBER_OF_ENABLED_STATIC_RULESETS`](#property-MAX_NUMBER_OF_ENABLED_STATIC_RULESETS). Collectively, those rulesets are guaranteed at least 30,000 rules. This is called the [`GUARANTEED_MINIMUM_STATIC_RULES`](#property-GUARANTEED_MINIMUM_STATIC_RULES).
+{% Aside 'note' %}
+Prior to Chrome 120, extensions were limited to a total of 50 static rulesets, and only 10 of these
+could be enabled at the same time.
+{% endAside %}
+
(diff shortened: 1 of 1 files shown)


site/en/docs/extensions/reference/declarativeNetRequest/index.md
```diff
@@ -211,7 +211,12 @@ rule you're using.
...
+{% Aside 'note' %}
+Prior to Chrome 120, extensions were limited to a total of 50 static rulesets, and only 10 of these
+could be enabled at the same time.
+{% endAside %}
+
+Static rules are those specified in rule files declared in the manifest file. An extension can specify up to 100 static [rulesets](#type-Ruleset) as part of the `"rule_resources"` manifest key, but only 50 of these rulesets can be enabled at a time. The latter is called the [`MAX_NUMBER_OF_ENABLED_STATIC_RULESETS`](#property-MAX_NUMBER_OF_ENABLED_STATIC_RULESETS). Collectively, those rulesets are guaranteed at least 30,000 rules. This is called the [`GUARANTEED_MINIMUM_STATIC_RULES`](#property-GUARANTEED_MINIMUM_STATIC_RULES).
```
jpmedley, line 219:
```suggestion
Static rules are those specified in rule files declared in the manifest file. An extension can specify up to 100 static [rulesets](#type-Ruleset) as part of the `"rule_resources"` manifest key, but only 50 of these rulesets can be enabled at a time. The latter is called the [`MAX_NUMBER_OF_ENABLED_STATIC_RULESETS`](#property-MAX_NUMBER_OF_ENABLED_STATIC_RULESETS). Collectively, those rulesets are guaranteed at least 30,000 rules. This is called the [`GUARANTEED_MINIMUM_STATIC_RULES`](#property-GUARANTEED_MINIMUM_STATIC_RULES).

{% Aside 'note' %}
Prior to Chrome 120, extensions were limited to a total of 50 static rulesets, and only 10 of these
could be enabled at the same time.
{% endAside %}
```

jpmedley: This is how I'd expect this to go even if it weren't a note. Supporting information goes after the information it supports. (That's called 'inverted pyramid' style.) 

I'd also suggest adding a link and text talking about [minimum chrome version](https://developer.chrome.com/docs/extensions/mv3/manifest/minimum_chrome_version/).

jpmedley (changes requested):

site/en/docs/extensions/reference/declarativeNetRequest/index.md
oliverdunk: Also added the link and text you suggested!

jpmedley (approved):

PR #7655, opened 2023-10-25
Mention Ephemeral mode in minimum_chrome_version docs

oliverdunk:
Based on some feedback, mention that minimum_chrome_version can behave in a surprising way for users of [ephemeral mode](https://support.google.com/chrome/a/answer/3538894). This is important because it impacts how safe rollouts can be in certain situations.


AmySteam (approved):
👍

oliverdunk:
Thanks @AmySteam for the quick review!

PR #7664, opened 2023-10-26
Replace Chrome OS with enterprise in description of ephemeral mode

oliverdunk:
I added this aside yesterday but just learned this applies to all enterprise users, not just Chrome OS.


AmySteam (approved):
👍

PR #7670, opened 2023-10-26
Add missing information

jpmedley:
This adds two pieces of information that are in the content scripts document and will allow depublication of content later.


AmySteam (approved):
👍 LGTM

PR #7675, opened 2023-10-27
More content deduplication for "content_scripts"

jpmedley:


AmySteam (changes requested):
Thanks for the additional information. However, I really don't like the way it breaks the rest of the document's formatting.

I would recommend adding the additional information but leaving the formatting as it was.

Issue #7601, opened 2023-10-19
Content feedback - tutorial incomplete/ incorrect


Wuzla:
https://developer.chrome.com/docs/extensions/mv3/getstarted/tut-tabs-manager/

Just above Test that it works section ..........
-- CURRENT -- 
const button = document.querySelector("button");
button.addEventListener("click", async () => {
  const tabIds = tabs.map(({ id }) => id);
  const group = await chrome.tabs.group({ tabIds });
  await chrome.tabGroups.update(group, { title: "DOCS" });
});

-- SHOULD BE -- 
const button = document.querySelector('button');
button.addEventListener('click', async () => {
  const tabIds = tabs.map(({ id }) => id);
  if (tabIds.length) {
    const group = await chrome.tabs.group({ tabIds });
    await chrome.tabGroups.update(group, { title: 'DOCS' });
  }
});

dstmarthe:
Thanks for raising this. We’re moving to a new platform and so we’ll fix this over there. I’ve moved the issue to our new tracker and you can see it at [https://issuetracker.google.com/issues/309981761](https://issuetracker.google.com/issues/309981761)

PR #7688, opened 2023-10-30
Make sure there are tabs before manipulating them.

jpmedley:
Addresses #7601 

@AmySteam This seems like a reasonable addition to this code sample. You probably wouldn't be trying to run it unless you had tabs open to documentation, but it's also probably easy to overlook. How many hours have we all wasted because silly minor thing we need to do.

This is already in the related sample.


jpmedley:
@AmySteam, Is this something you want to do?

PR #5862, opened 2023-03-30
 Update to FLEDGE settings content [DON'T MERGE BEFORE REVIEW]

samdutton:
Updated branch name from samdutton-patch-4.

PR #7699, opened 2023-10-31
Update Protected Audience article

samdutton:
Changes from #5862, which refers to the old /fledge/ URL for this article.


kevinkiklee:
Hmm, no page loads when I enter `chrome://settings/adPrivacy/sites` to the address bar.  Do you know if that page is still active?

samdutton:
> Hmm, no page loads when I enter `chrome://settings/adPrivacy/sites` to the address bar. Do you know if that page is still active?

@kevinkiklee I see chrome://settings/adPrivacy/sites (with sites) in Stable 119.0.6045.59, and in Canary. What happens when you load that chrome: URL?

nancymic2 (approved):

PR #7719, opened 2023-11-02
Fix incorrect API use in index.md

yi-gu:
`IdentityProvider.logout` isn't the right API to call in this case

- Replace `IdentityProvider.logout` with `navigator.login.setStatus("logged-out")`


yi-gu:
@heyawhite PTAL!
cc @agektmr

agektmr (approved):
LGTM

PR #7738, opened 2023-11-03
fix: change fir to fire

rustyzone:
Noticed this small typo.

@sebastianbenz Also had a question in general around the need to use `setTimeout` for below 30 seconds & then `chrome.alarms` for more than 30 seconds - this feels like an area that could cause confusion for new developers who might start with the need for a short time (using setTimeout) and then fall into an issue when trying to use this next for a delay of 60 seconds.


rachelandrew (approved):

sebastianbenz:
Thanks for the fix Russell! Yeah, that's a good point. This probably warrants it's own short how to guide. I'll create an issue for this.

PR #7754, opened 2023-11-06
what's new update: 120 crx udpate

sebastianbenz:


AmySteam (approved):
Approving with a small suggestion to improve flow

PR #7761, opened 2023-11-07
add storage.session in "What's new in extensions"

hanguokai:
`storage.session` was available in Chrome 96 without doc. It was official available in Chrome 102 with doc.


AmySteam (approved):
Thanks @hanguokai! This is super helpful 💯

PR #7771, opened 2023-11-08
What's new in DevTools 120

sofiayem:
This PR contains changes to exiting content. Both of them of high importance. One is solved with a replace command across several files, the other is a single line. I'll follow up with an identical CL myself.


rachelandrew (approved):
The content isn't available to make a CL yet, so can you add an issue to Buganizer as in the email I sent on Monday just so we know to keep track? You can close it once you d the CL.

PR #7772, opened 2023-11-08
Updated Topics bucket assignment link

samdutton:
For migration: will add issue for this, and add this to hotlist.


nancymic2 (approved):
is just "places" the best text for that link? It's so small :). anyway, just a thought

PR #7776, opened 2023-11-09
Document device sleep and persistence behaviour of alarms API

oliverdunk:
Adds some documentation on more nuanced behaviour in the `chrome.alarms` API.

The wording here is quite vague but unfortunately I think it's the best we can do at the moment. There are several open issues around the behaviour we are close to resolving but which are still open:

- https://github.com/w3c/webextensions/issues/433
- https://github.com/w3c/webextensions/issues/406
- https://bugs.chromium.org/p/chromium/issues/detail?id=1285798

While we could wait for those to be resolved, I think it's important we at least document the current behaviour since this is a heavily used API and the fixes may take time.

Fixes https://github.com/GoogleChromeLabs/Extension-Docs/issues/167

### DIFF
--- a/site/en/docs/extensions/reference/alarms/index.md
+++ b/site/en/docs/extensions/reference/alarms/index.md
@@ -27,9 +27,9 @@ scheduled.
 
 ### Persistence
 
-Alarms generally persist until an extension is updated. However, this is not guarenteed and alarms
+Alarms generally persist until an extension is updated. However, this is not guaranteed, and alarms
 may be cleared when the browser is restarted. Consequently, consider setting a value in storage
-when an alarm is created and then making sure it exists each time your service worker starts up:
+when an alarm is created, and then ensure it exists each time your service worker starts up. For example:
 
 ```js
 const STORAGE_KEY = "user-preference-alarm-enabled";
--- a/site/en/docs/extensions/reference/alarms/index.md
+++ b/site/en/docs/extensions/reference/alarms/index.md
@@ -65,6 +65,15 @@ chrome.alarms.onAlarm.addListener((alarm) => {
 });
 ```
 
+### Device sleep
+
+The behavior of alarms when a device goes to sleep is currently undefined. An alarm will never fire
+early but may fire significantly later than expected if a device went to sleep after an alarm was scheduled.
+
+### Persistence
+
+Alarms generally persist until an extension is updated. However, this is not guarenteed and alarms may be cleared when the browser is restarted. Consequently, consider storing any important alarms in storage and making sure they exist each time your service worker starts up.
+
 [repo-samples]: https://github.com/GoogleChrome/chrome-extensions-samples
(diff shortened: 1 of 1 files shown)


site/en/docs/extensions/reference/alarms/index.md
```diff
@@ -65,6 +65,15 @@ chrome.alarms.onAlarm.addListener((alarm) => {
 });
 ```
```
AmySteam, line 67: @oliverdunk NIT: Could we add these items under a " ## Concepts and usage" section above the examples version? We are trying to follow this pattern. Example [ReadingList concepts](https://developer.chrome.com/docs/extensions/reference/readingList/#concepts),  [DNR concepts](https://developer.chrome.com/docs/extensions/reference/declarativeNetRequest/#concepts-and-usage)

oliverdunk: Done 😄

site/en/docs/extensions/reference/alarms/index.md
```diff
@@ -17,6 +17,20 @@ To use the `chrome.alarms` API, declare the `"alarms"` permission in the [manife
...
+
+The behavior of alarms when a device goes to sleep is currently undefined. An alarm will never fire
+early but may fire significantly later than expected if a device went to sleep after an alarm was
+scheduled.
+
+### Persistence
+
+Alarms generally persist until an extension is updated. However, this is not guarenteed and alarms
+may be cleared when the browser is restarted. Consequently, consider storing any important alarms in
+storage and making sure they exist each time your service worker starts up.
```
sebastianbenz, line 32: Code snippet would be nice here.

sebastianbenz (approved):

site/en/docs/extensions/reference/alarms/index.md
oliverdunk: Good call, added some code :)

site/en/docs/extensions/reference/alarms/index.md
```diff
@@ -17,6 +17,39 @@ To use the `chrome.alarms` API, declare the `"alarms"` permission in the [manife
...
+
+The behavior of alarms when a device goes to sleep is currently undefined. An alarm will never fire
+early but may fire significantly later than expected if a device went to sleep after an alarm was
+scheduled.
+
+### Persistence
+
+Alarms generally persist until an extension is updated. However, this is not guarenteed and alarms
+may be cleared when the browser is restarted. Consequently, consider setting a value in storage
+when an alarm is created and then making sure it exists each time your service worker starts up:
```
AmySteam, line 32:
```suggestion
Alarms generally persist until an extension is updated. However, this is not guaranteed, and alarms
may be cleared when the browser is restarted. Consequently, consider setting a value in storage
when an alarm is created, and then ensure it exists each time your service worker starts up. For example:
```

AmySteam (approved):
Great additions! I made a suggestion that would be good to apply before merging. I'm approving nevertheless to not be a blocker

PR #7879, opened 2023-11-28
Update 3PCD DT registration start date

mihajlija:
Article updated with the start date the week of Dec 4 (Monday). 

cc @rowan-m


rowan-m (changes requested):
Also update dates in https://developer.chrome.com/docs/privacy-sandbox/third-party-cookie-phase-out/ but I don't think it needs to the aside callout for that page. I'm happy with the update otherwise.

PR #7908, opened 2023-12-05
Archive README

alexandrascript:
Announce archive in README

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-## developer.chrome.com Archived
+# developer.chrome.com Archived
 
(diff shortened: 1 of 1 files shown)


rachelandrew (approved):

paulirish:
Do you want to mark the repo as archived? I could do that if we're ready.

rachelandrew:
@paulirish we can do, I still have some issues to move over to the new issuetracker, but presumably it's all still available read-only once archived.

paulirish:
yeah the whole repo/issues will remain read only. 
but if there's any places you still wanna leave a comment, we should do that first.

i'm not in a rush, just offering to help. :)

rachelandrew:
thanks :) I'll have a look through as if there are any public issues to move, I'd like to comment and let them know where it went. Then we should be good to go.

